PR #42, opened 2023-06-12
update poetry pydantic dependency to 1.10.8

jinlin-flex:
[Jira Ticket:](https://flow360.atlassian.net/browse/FLPY-24)
Update poetry pydantic dependency to 1.10.8


maciej-flexcompute (approved):
squash before merging

PR #47, opened 2023-06-14
zip VolumeMesh before upload

jinlin-flex:
[Jira Ticket](https://flow360.atlassian.net/browse/FLPY-18?atlOrigin=eyJpIjoiZTRjNWVhNTQzYjY4NGYwNmIwZjE1Y2FiNzJhOGI5MmUiLCJwIjoiaiJ9)
zip VolumeMesh before upload


flow360/component/volume_mesh.py
```diff
@@ -452,7 +453,14 @@ def _submit_upload_mesh(self, progress_callback=None):
...
         mesh = VolumeMesh(self.id)
-        mesh.upload_file(remote_file_name, self.file_name, progress_callback=progress_callback)
+
+        # upload mesh
+        with zipfile.ZipFile(self.file_name + ".zip", "w", zipfile.ZIP_DEFLATED) as zipf:
+            zipf.write(self.file_name, arcname=self.file_name)
+
+        mesh.upload_file(
+            remote_file_name, self.file_name + ".zip", progress_callback=progress_callback
+        )
```
maciej-flexcompute, line 463: 1. where do we save this file? I think we should save it to temp location and delete after upload. We don't want to leave user with any extra file.
2. Zip is not supported by our backend, use *.gz
3. You need to first zip and, if successful, change NewVolumeMeshRequest() line 439
4. Can we do some performance test? How quick is gz in MB/s?

PR #62, opened 2023-10-03
Fix issues with long testing times on Windows caused by log rotation from multiple threads

awkrupka:
For the past two months testing times on Windows were excessively long - this was caused by OS-specific restrictions on file operations from multiple threads. Since this issue is test-specific a separate log file for test output has been created. Additionally the maximum file size before log rotation was small (10kB) and has been increased to 1MB. This brings Windows test performance in line with other systems.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,32 @@
+import os
+import tempfile
+
+import pytest
+
+from flow360.file_path import flow360_dir
+from flow360.log import log, set_logging_file
+
+"""
+Before running all tests redirect all test logging to a temporary log file, turn off log rotation 
+due to multi-threaded rotation being unsupported at this time
+"""
+
+
+def pytest_configure():
+    fo = tempfile.NamedTemporaryFile()
+    pytest.tmp_log_file = fo.name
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+from flow360.file_path import flow360_dir
+from flow360.log import log, set_logging_file
+
+"""
+Before running all tests redirect all test logging to a temporary log file, turn off log rotation 
+due to multi-threaded rotation being unsupported at this time
+"""
+
+log_dir = flow360_dir + "logs"
+log_path = os.path.join(log_dir, "flow360_temp.log")
```
maciej-flexcompute, line 14: `log_dir = os.path.join(flow360_dir, "logs")`
or combine with next line.

We should clean up after tests. Would suggest to log to a temp file which will be deleted at the end of the test

awkrupka: Fixed in dbe1d20, only logs from test_log.py are not temporary and remain in a separate small file that is cleaned for every test run.

maciej-flexcompute (approved):

PR #64, opened 2023-10-04
Added shared account selection module based on the old python API

awkrupka:
User can now change his account to one of the accounts shared with him (company and clients) while interacting with the backend. This can be done in one of two ways, by calling:

```
import flow360 as fl

fl.Accounts.choose_shared_account(<email_address>)
```

where the account will be switched if the provided email address exists and is shared with the current user or

```
import flow360 as fl

fl.Accounts.choose_shared_account()
```

where the user will be prompted to select an account from the list interactively.

```
Choosing account in interactive mode, please select email from the organization list: 
0: lei@flexcompute.com
1: thomas@flexcompute.com
2: qiqi@flexcompute.com
3: yuwei@flexcompute.com
4: chaitanya@flexcompute.com
5: ben@flexcompute.com
6: runda@flexcompute.com
7: jim@flexcompute.com
8: lin@flexcompute.com
9: james@flexcompute.com
10: yongwei@flexcompute.com
11: maciej@flexcompute.com
12: cj@flexcompute.com
13: feilin@flexcompute.com
Enter address of the account to switch to [0 - 13] or 'q' to abort:
```

When submitting resources to the server while using a shared account, the user is prompted to confirm the action:

```
WARNING: You are submitting a resource to a shared account maciej@flexcompute.com.                                             
         This message will not be shown again for this session if you confirm.                                                            
                                                                               
Are you sure you want to proceed? (y/n): 
```

This message shows once per session until the user logs out if the user confirms interactively or by calling
```
fl.Accounts.confirm_submit()
```
NOTE: Due to backend issues we currently do not support client accounts, as apikey authentication does not seem to work when requesting their data from the webapi. This will be fixed separately.

### DIFF
--- a/flow360/accounts_utils.py
+++ b/flow360/accounts_utils.py
@@ -1,0 +1,166 @@
+"""
+This module provides utility functions for managing access between interconnected accounts.
+
+Functions:
+- choose_shared_account(None) -> None - select account from the list of client and organization accounts interactively
+- choose_shared_account(email: str, optional) -> None - select account matching the provided email (if exists)
+- shared_account_info(None) -> str - return current shared account email address (if exists, None otherwise)
+- leave_shared_account(None) -> None - leave current shared account
+"""
+
+from requests import HTTPError
+
+from flow360.cloud.http_util import http
+from flow360.environment import Env
+from flow360.log import log
+
+from .exceptions import WebError
+
+
+class AccountsUtils:
+    """
+    Current account info and utility functions.
+    """
+
+    def __init__(self):
+        self._current_email = None
+        self._current_user_identity = None
+        self._confirmed_submit = False
+
+    @staticmethod
+    def _interactive_selection(users):
+        print(
+            "Choosing account in interactive mode, please select email from the organization list: "
+        )
+
+        user_count = len(users)
+
+        for i in range(0, user_count):
+            print(f"{i}: {users[i]['userEmail']}")
+
+        while True:
+            try:
+                value = input(
+                    f"Enter address of the account to switch to [0 - {user_count - 1}] or 'q' to abort: "
+                )
+                if value == "q":
+                    return None
+                if int(value) in range(0, user_count):
+                    return int(value)
+                print(f"Value out of range [0 - {user_count - 1}]")
+                continue
+            except ValueError:
+                print("Invalid input type, please input an integer value:")
+                continue
+
+    # Requires fixing from the backend side - support for portal webapi calls with apikey authentication
+    @staticmethod
+    def _get_supported_users():
+        try:
+            response = http.portal_api_get("auth")
+            access = response.json()["data"]
+            keys = access["user"]
+            supported_users = keys["guestUsers"]
+            if supported_users:
+                return supported_users
+            return []
+        except HTTPError as error:
+            raise WebError("Failed to retrieve supported user data from server") from error
+
+    @staticmethod
+    def _get_company_users():
+        try:
+            response = http.get("flow360/account")
+            company_users = response["tenantMembers"]
+            if company_users:
+                return company_users
+            return []
+        except HTTPError as error:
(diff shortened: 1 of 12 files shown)


flow360/error_messages.py
```diff
@@ -10,6 +10,12 @@
 """
 
 
+shared_submit_warning = """\
+You are submitting a resource from a shared account.
+To suppress this message run: flow360 configure --suppress-shared-account-prompt
+"""
```
maciej-flexcompute, line 16: I think the message should read:
`You are submitting a resource to a shared account (email address).`

awkrupka: Fixed in 5035ff0 - now confirmation works per-session and needs to be performed only once until the user logs out, either interactively or by the user calling 

```
fl.Accounts.confirm_submit()
```

flow360/accounts_utils.py
```diff
@@ -0,0 +1,166 @@
...
+        """
+        company_users = self._get_company_users()
+
+        if len(company_users) == 0:
+            log.info("There are no accounts shared with the current user")
+            return
+
+        selected = None
+
+        addresses = [d["userEmail"] for d in company_users]
```
maciej-flexcompute, line 119: we try to avoid single letter variables, even in such instances. Use `user` instead

flow360/accounts_utils.py
```diff
@@ -0,0 +1,166 @@
...
+    def choose_shared_account(self, email=None):
+        """choose a shared account to impersonate
+
+        Parameters
+        ----------
+        email : str, optional
+           user email to impersonate (if email exists among shared accounts),
+           if email is not provided user can select the account interactively
+        """
+        company_users = self._get_company_users()
```
maciej-flexcompute, line 111: lets use a list that will have both company users and supported users. Supported in future. Suggest a name `shared_accounts`

flow360/component/utils.py
```diff
@@ -48,6 +50,31 @@ def wrapper_func(*args, **kwargs):
     return wrapper
 
 
+def confirm_proceed():
```
maciej-flexcompute, line 53: rename function:
`shared_account_confirm_proceed()` or something similar to indicate that this confirmation is only in case of shared account

flow360/log.py
```diff
@@ -289,7 +289,7 @@ def set_logging_file(
     filemode: str = "a",
     level: LogValue = DEFAULT_LEVEL,
     back_up_count: int = 10,
-    max_bytes: int = 10000,
+    max_bytes: int = 100000000,
```
maciej-flexcompute, line 292: why changing this again? Is it a conflict?

awkrupka: Accidentally changed when running tests to speed up the process before merging. Reverting in next commit to prevent conflict.

awkrupka: Reverted in 644f8ee

flow360/component/utils.py
awkrupka: Renamed to shared_account_confirm_submit() in 644f8ee

flow360/accounts_utils.py
awkrupka: Renamed to shared_accounts in 644f8ee

flow360/accounts_utils.py
awkrupka: Renamed to user in 644f8ee

maciej-flexcompute (approved):

flow360/accounts_utils.py
```diff
@@ -0,0 +1,166 @@
...
+        self._check_state_consistency()
+
+        if self._current_email is not None:
+            log.info(f"Currently operating as {self._current_email}")
+        else:
+            log.info("Currently not logged into a shared account")
+
+        return self._current_email
+
+    def leave_shared_account(self):
```
maciej-flexcompute, line 151: on leave, we should reset the submit confirm question

PR #65, opened 2023-10-04
Added environment-dependent apikeys

awkrupka:
The user can now declare separate apikeys for the dev environment by calling configure with an additional dev flag
```
flow360 configure --profile <profile_name> --dev --apikey aAAa0000(...)
```
which will cause the provided apikey to be used for this profile when dev environment is activated. This is backwards-compatible with the current solution as the default apikey can be configured by calling
```
flow360 configure --profile <profile_name> --apikey aAAa0000(...)
```
same as before. Since dev apikey does not use a different profile anymore, we do not set the profile automatically when activating an environment.

### DIFF
--- a/flow360/cloud/http_util.py
+++ b/flow360/cloud/http_util.py
@@ -22,6 +22,10 @@ def api_key_auth(request):
     """
     key = api_key()
     if not key:
+        if Env.current.name == "dev":
+            raise AuthorisationError(
+                "API key not found for env=dev, please set it by commandline: flow360 configure --dev."
+            )
         raise AuthorisationError(
             f"API key not found for profile={UserConfig.profile}, please set it by commandline: flow360 configure."
         )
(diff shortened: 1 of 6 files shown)


flow360/cloud/http_util.py
```diff
@@ -23,7 +23,8 @@ def api_key_auth(request):
     key = api_key()
     if not key:
         raise AuthorisationError(
-            f"API key not found for profile={UserConfig.profile}, please set it by commandline: flow360 configure."
+            f"API key not found for profile={UserConfig.profile} in "
+            f"environment={Env.current.name}, please set it by commandline: flow360 configure."
```
maciej-flexcompute, line 27: env is only for developers, so we don't want to show it usual auth error. There should be a message when used in Env.dev.active() and apikey not found:
`apikey not found for env=dev, set it by running flow360 configure with --dev flag`

awkrupka: Fixed in 2509716

maciej-flexcompute (approved):

PR #73, opened 2023-10-26
Introduced unit system, added support for [] for boundary/zones

maciej-flexcompute:


awkrupka (approved):
Approved, most changes were already reviewed in PR #75

PR #74, opened 2023-10-31
added support for case interpolation on new mesh

maciej-flexcompute:


examples/run_case_with_fork.py
```diff
@@ -1,31 +1,39 @@
...
 
 # submit case using json file
 params = fl.Flow360Params(OM6wing.case_json)
-case = fl.Case.create("OM6wing", params, volume_mesh.id)
+case = fl.Case.create("OM6wing", params, volume_mesh.id, solver_version='release-23.3.2.0')
 case = case.submit()
 
 # fork a case
-case_fork_1 = case.fork()
+case_fork_1 = case.fork(interpolate_on_mesh_id=volume_mesh2.id)
```
chai-flex, line 25: There are two ways of designing this feature:
1. fork to a new mesh id (which is what is done here)
2. create from a parent case id

Should we support both?

PR #75, opened 2023-11-02
Added support for dimensioned variables

awkrupka:


flow360/component/flow360_params/unit_system.py
```diff
@@ -0,0 +1,589 @@
...
+
+
+class MassType(DimensionedType):
+    """:class: MassType"""
+
+    dim = u.dimensions.mass
+    dim_name = "mass"
+
+
+class TimeType(DimensionedType):
```
chai-flex, line 199: It would be good to split the unit system into two parts: Base and Derived
Base unit system is Mass, Length, Time, Temperature
Derived unit system is Force, Velocity, Area, Pressure, etc.

chai-flex: Derived system can be created directly from Base, but maybe also have an option to override Derived

flow360/component/flow360_params/unit_system.py
```diff
@@ -0,0 +1,589 @@
...
+            if system == "Imperial":
+                return _imperial_system[dim_name]
+            if system == "Flow360":
+                return _flow360_system[dim_name]
+        return None
+
+    def __init__(self, **kwargs):
+        base_system = kwargs.get("base_system")
+
+        dim_names = [
```
chai-flex, line 457: What is the purpose of strings for dimension names here?

It is preferable to use enums or something similar to define properties

Python has reflection so you can convert enums to strings easily.

awkrupka: This is true, changed the argument type to string in 48ebb7c

awkrupka:
After consulting with Maciej - some checks fail for 3.7 but PR #73 from the base branch of this PR deprecates python 3.7 - this should not be an issue, we will possibly need to update the checks.

flow360/component/flow360_params/unit_system.py
chai-flex: Okay, as long as the user can specify a subset of the unit system (MLTK) and infer the others from the base using those units it is fine. The separate classes was more of a convenience as MLTK is a standard in scientific texts and our end users might be more familiar with it.

chai-flex (approved):

PR #83, opened 2023-11-13
Added schema generation and serialization/deserialization of dimensioned types

awkrupka:


examples/dev/schema.json
maciej-flexcompute, line 1: lets remove this file. Otherwise we would need keep it consistent is any change

awkrupka: I was sure this was not included in the commit, my bad - will remove.

awkrupka: Removed in c0fe2d3

maciej-flexcompute (approved):

PR #84, opened 2023-11-14
Rename linear solver config

chai-flex:
Config does not belong in a configuration file


flow360/component/flow360_params/solvers.py
```diff
@@ -289,6 +289,8 @@ class HeatEquationSolver(Flow360BaseModel):
 
     model_type: Literal["HeatEquation"] = pd.Field("HeatEquation", alias="modelType", const=True)
     equation_eval_frequency: Optional[PositiveInt] = pd.Field(alias="equationEvalFrequency")
-    linear_solver_config: Optional[LinearSolver] = pd.Field(
-        alias="linearSolverConfig", default=LinearSolver()
+    linear_solver: Optional[LinearSolver] = pd.Field(
+        validation_alias=AliasChoices("linearSolver", "linearSolverConfig"),
```
maciej-flexcompute, line 293: I think validation_alias and AliasChoices are pydantic v2. We are using pydantic v1 so far and have not started migration yet. For multiple aliases we have such solution:
https://github.com/flexcompute/Flow360/blob/143f28766126e286e0cb633541c990b991067b55/flow360/component/flow360_params/solvers.py#L33C2-L33C2

maciej-flexcompute (approved):

PR #85, opened 2023-11-16
Added the support for turbulence values specification

benflexcompute:


benflexcompute:
Test pending. The `poetry install` hangs on dev4 for no reason...

flow360/component/flow360_params/flow360_params.py
```diff
@@ -73,6 +73,98 @@ class MeshBoundary(Flow360BaseModel):
...
+    turbulentIntensity: Optional[NonNegativeFloat] = pd.Field(alias="turbulentIntensity")
+    specificDissipationRate: Optional[NonNegativeFloat] = pd.Field(alias="specificDissipationRate")
+    turbulentViscosityRatio: Optional[NonNegativeFloat] = pd.Field(alias="turbulentViscosityRatio")
+    turbulentLengthScale: Optional[PositiveFloat] = pd.Field(alias="turbulentLengthScale")
+    modifiedTurbulentViscosityRatio: Optional[PositiveFloat] = pd.Field(
+        alias="modifiedTurbulentViscosityRatio"
+    )
+    modifiedTurbulentViscosity: Optional[PositiveFloat] = pd.Field(
+        alias="modifiedTurbulentViscosity"
+    )
```
maciej-flexcompute, line 122: use snake_case for model parameters and camelCase for aliases

flow360/component/flow360_params/flow360_params.py
```diff
@@ -73,6 +73,98 @@ class MeshBoundary(Flow360BaseModel):
...
+
+    # pylint: disable=no-self-argument
+    @pd.root_validator(pre=True)
+    def validate_validate_combination(cls, value: dict):
+        """
```
maciej-flexcompute, line 127: lets try to achieve the same based on data structure without using custom validations. Custom validations are problematic to export to webUI via JSON schema whereas data structure can easily be exported.

for example:
```python
     set(["specificDissipationRate", "turbulentLengthScale"]),
```
can be done using:
```python
SpecificDissipationRateAndTurbulentLengthScale(Flow360BaseModel):
    specificdissipationrate: Optional[NonNegativeFloat] = pd.Field(alias="specificDissipationRate")
    turbulentlengthscale: Optional[PositiveFloat] = pd.Field(alias="turbulentLengthScale")
```

and then:
```python
class FreestreamBoundary(Boundary):
     """Freestream boundary"""

     type = pd.Field("Freestream", const=True)
     velocity: Optional[BoundaryVelocityType] = pd.Field(alias="Velocity")
     turbulence_value_specification: Optional[Union[SpecificDissipationRateAndTurbulentLengthScale, ...] = pd.Field(alias="turbulenceValueSpecification")
```
so `Union` will be responsible for all possible combinations and validations. When you provide 3, then pydantic will not be able to find valid a type for it.

the solver side would need to change:
from `boundaries -> name -> specificDissipationRate`
to `boundaries -> name -> turbulenceValueSpecification -> specificDissipationRate`

flow360/component/flow360_params/flow360_params.py
benflexcompute: Done

flow360/component/flow360_params/flow360_params.py
benflexcompute: I am concerned that the user will need to write something like:

`bc = FreestreamBoundary(turbulenceQuantities=SpecificDissipationRateAndTurbulentLengthScale(specificdissipationrate=0.2, turbulentlengthscale = 0.5))`

 which is super long. We can move this (check valid combinations) to the root validator. We will need a root validator for this PR anyway for checking SST and SA compatible combinations.

flow360/component/flow360_params/flow360_params.py
```diff
@@ -1529,7 +1699,7 @@ def to_flow360_json(self) -> dict:
         """
 
         solver_params = self.to_solver()
-        solver_params_json = solver_params.json(encoder=flow360_json_encoder)
+        solver_params_json = solver_params.json(exclude_unset=True, encoder=flow360_json_encoder)
```
maciej-flexcompute, line 1702: exclude_unset should not be changed, we need it for default settings used on webUI. What was the reason to add it here?

flow360/component/flow360_params/flow360_params.py
```diff
@@ -119,6 +119,152 @@ class MeshBoundary(Flow360BaseModel):
...
+class SpecificDissipationRateAndTurbulentLengthScale(SpecificDissipationRate, TurbulentLengthScale):
+    pass
+
+
+# pylint: disable=missing-class-docstring
+class TurbulentViscosityRatioAndTurbulentLengthScale(TurbulentViscosityRatio, TurbulentLengthScale):
+    pass
+
+
+AllowedTurbulenceSpecifications = Union[
```
maciej-flexcompute, line 252: Lets call this type the same as the parameter: `TurbulenceQuantitiesType`

flow360/component/flow360_params/flow360_params.py
maciej-flexcompute: Agree, it is now very long. I think we can create a function for the user for easier input:
```python
FreestreamBoundary(turbulence_quantities=turbulence_quantities(dissipation_rate=0.2, length_scale=0.5))
```
the function can then try to match type using pydantic and return object of a correct type. This way we can keep the data structure validation and have reasonable input way for the user.

maciej-flexcompute: Few other ideas:
1. We can remove `turbulent` from parameter names if they are provided to a function/class with turbulent in its name. 
2. The proposed function turbulence_quantities() can also named using class naming convention `TurbulenceQuantities()` if that will be more natural for the user in this specific instance

flow360/component/flow360_params/flow360_params.py
benflexcompute: Deleted. This was a residue from previous tries.

flow360/component/flow360_params/flow360_params.py
benflexcompute: Done

flow360/component/flow360_params/flow360_params.py
benflexcompute: Removed some "turbulent" as the others sounds strange without turbulent.

flow360/component/flow360_params/flow360_params.py
```diff
@@ -150,6 +371,9 @@ class FreestreamBoundary(Boundary):
 
     type = pd.Field("Freestream", const=True)
     velocity: Optional[BoundaryVelocityType] = pd.Field(alias="Velocity")
+    turbulence_quantities: Optional[TurbulenceQuantitiesType] = pd.Field(
+        alias="turbulenceQuantities"
+    )
```
maciej-flexcompute, line 376: move to an abstract class:
```python
class BoundaryWithTurbulenceQuantities(Boundary, metaclass=ABCMeta)
```
and inherit by it

flow360/component/flow360_params/flow360_params.py
```diff
@@ -119,6 +119,227 @@ class MeshBoundary(Flow360BaseModel):
...
+
+# pylint: disable=too-many-arguments, too-many-return-statements, too-many-branches
+def TurbulenceQuantities(
+    viscosityRatio=None,
+    modifiedViscosityRatio=None,
+    modifiedViscosity=None,
+    specificDissipationRate=None,
+    turbulentKineticEnergy=None,
+    turbulentLengthScale=None,
+    turbulentIntensity=None,
```
maciej-flexcompute, line 270: use snake_case for variables. I don't understand why linter does not complain here...
In general, in this python client we follow standard python naming convention:
```python
CONST
def function_name()

class ClassName()

variable_name

```
camelCase is only used as aliases to match solver convention on JSON export.

flow360/component/flow360_params/flow360_params.py
```diff
@@ -1399,7 +1644,7 @@ class UserDefinedDynamic(Flow360BaseModel):
     name: str = pd.Field(alias="dynamicsName")
     input_vars: List[str] = pd.Field(alias="inputVars")
     constants: Optional[Dict] = pd.Field()
-    output_vars: Union[Dict] = pd.Field(alias="outputVars")
+    output_vars: Union[Dict, list] = pd.Field(alias="outputVars")
```
maciej-flexcompute, line 1647: why this change in this PR?

tests/params/test_params_boundary.py
```diff
@@ -213,3 +216,12 @@ def test_boundary_types():
...
         MassOutflow(massFlowRate=-1)
+
+    bc = SubsonicOutflowMach(
+        name="SomeBC",
+        Mach=0.2,
+        turbulenceQuantities=TurbulenceQuantities(turbulentIntensity=0.2, viscosityRatio=10),
+    )
+
+    assert bc.turbulence_quantities.turbulent_intensity == 0.2
+    assert bc.turbulence_quantities.turbulent_viscosity_ratio == 10
```
maciej-flexcompute, line 227: lets extend these tests to cover different scenarios including raised exceptions. You can run `pytest tests --cov-report=html --cov=flow360 && open htmlcov/index.html`
to see test coverage. 
Screenshot 2023-12-14 at 13 14 31

flow360/component/flow360_params/flow360_params.py
benflexcompute: Changed

flow360/component/flow360_params/flow360_params.py
benflexcompute: Removed.

tests/params/test_params_boundary.py
benflexcompute: Added more tests

flow360/component/flow360_params/flow360_params.py
```diff
@@ -119,6 +119,227 @@ class MeshBoundary(Flow360BaseModel):
...
+    pass
+
+
+class TurbulentViscosityRatioAndTurbulentLengthScale(TurbulentViscosityRatio, TurbulentLengthScale):
+    pass
+
+
+# pylint: enable=missing-class-docstring
+
+TurbulenceQuantitiesType = Union[
```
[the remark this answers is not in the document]
benflexcompute: Good point. I made classes that are not supposed top be instantiated to be abstract and private.

benflexcompute:
Just talked with CJ. Additional changes pending for this PR. Please do not merge yet.

tests/params/test_params_boundary.py
```diff
@@ -1,4 +1,6 @@
 import unittest
+from os import write
+from re import T
```
maciej-flexcompute, line 3: it seems these are not used

flow360/component/flow360_params/flow360_params.py
```diff
@@ -161,6 +398,14 @@ class Boundary(Flow360BaseModel, metaclass=ABCMeta):
...
 
 
+class BoundaryWithTurbulenceQuantities(Flow360BaseModel, metaclass=ABCMeta):
+    """Turbulence Quantities on Boundaries"""
+
+    turbulence_quantities: Optional[TurbulenceQuantitiesType] = pd.Field(
+        alias="turbulenceQuantities"
+    )
+
+
```
maciej-flexcompute, line 408: `BoundaryWithTurbulenceQuantities` - better would be to inherit from `Boundary` and in the child classes inherit only from `BoundaryWithTurbulenceQuantities`, not both. When inheriting by 2 we need to remember to not put into conflict these 2 classes. 

Alternatively, if we also want to use `BoundaryWithTurbulenceQuantities` elsewhere (not noly boundaries), the class should be named `TurbulenceQuantities`

tests/params/test_params_boundary.py
benflexcompute: Removed

flow360/component/flow360_params/flow360_params.py
benflexcompute: I thought about it. The reason I did not do that is because then the `BoundaryWithTurbulenceQuantities` would have `type` and `name` property inherited from `Boundary`. 

These new properties will be ignored on the solver side. Just not sure if it will cause confusion to the users.

Anyway I have changed it to inherit from `Boundary`.

maciej-flexcompute (approved):

PR #88, opened 2023-11-20
Added unit schema generation for dimensioned types

awkrupka:


flow360/component/flow360_params/flow360_params.py
```diff
@@ -1264,6 +1268,7 @@ class Flow360Params(Flow360BaseModel):
     monitor_output: Optional[MonitorOutput] = pd.Field(alias="monitorOutput")
     volume_zones: Optional[VolumeZones] = pd.Field(alias="volumeZones")
     aeroacoustic_output: Optional[AeroacousticOutput] = pd.Field(alias="aeroacousticOutput")
+    unit_defaults: Optional[UnitDefaults] = pd.Field(alias="unitDefaults")
```
maciej-flexcompute, line 1271: create `ROOT/tools/integrations/unit_systems_defaults.py` script that will export these

examples/dev/dev_use_unit_system.py
```diff
@@ -195,3 +196,13 @@ def thread2():
...
 # Without locks thread2 executes before thread 1
 print(f"After running both threads: {threaded_data.m}")
+
+schema = fl.SI_unit_system.schema()
+
+print(schema)
+
+defaults = UnitDefaults()
+
+schema = defaults.schema()
```
maciej-flexcompute, line 206: we don't need schema for it, right? Just `defaults.dict()`

flow360/component/flow360_params/flow360_params.py
```diff
@@ -1323,3 +1328,8 @@ def __init__(self, filename: str = None, **kwargs):
     # pylint: disable=missing-class-docstring,too-few-public-methods
     class Config(Flow360BaseModel.Config):
         extra = "allow"
+
+class UnitDefaults(Flow360BaseModel):
+    defaults_SI: Dict = pd.Field(alias="defaultsSI", default=SI_unit_system.defaults())
+    defaults_CGS: Dict = pd.Field(alias="defaultsCGS", default=CGS_unit_system.defaults())
+    defaults_imperial: Dict = pd.Field(alias="defaultsImperial", default=imperial_unit_system.defaults())
```
maciej-flexcompute, line 1335: put in `ROOT/tools/integrations/unit_systems_defaults.py`

examples/dev/dev_use_unit_system.py
awkrupka: Correct, currently only a dict is exported in the script.

flow360/component/flow360_params/flow360_params.py
awkrupka: Fixed in bfd5375

flow360/component/flow360_params/unit_system.py
```diff
@@ -190,6 +190,13 @@ def __modify_schema__(cls, field_schema, field):
         field_schema["units"] = {}
         field_schema["value"]["type"] = "number"
         field_schema["units"]["type"] = "string"
+        if cls.dim_name is not None:
+            field_schema["units"]["dimension"] = cls.dim_name
+            field_schema["units"]["enum"] = [
+                str(_SI_system[cls.dim_name]),
+                str(_CGS_system[cls.dim_name]),
+                str(_imperial_system[cls.dim_name])
```
maciej-flexcompute, line 198: lets move the available units per dimension into a separate file

awkrupka: This has been partially solved, the schema generator now looks like this:

```
@classmethod
def __modify_schema__(cls, field_schema, field):
    field_schema["value"] = {}
    field_schema["units"] = {}
    field_schema["value"]["type"] = "number"
    field_schema["units"]["type"] = "string"
    if cls.dim_name is not None:
        field_schema["units"]["dimension"] = cls.dim_name
        # Local import to prevent exposing mappings to the user
        from flow360.component.flow360_params.exposed_units import extra_units
        units = [
            str(_SI_system[cls.dim_name]),
            str(_CGS_system[cls.dim_name]),
            str(_imperial_system[cls.dim_name]),
        ]
        units += extra_units[cls.dim_name]
        field_schema["units"]["enum"] = list(dict.fromkeys(units))
```

We use a local import of the extra_units dict - this way we don't expose any UI schema-specific data to the user. Default units for SI/CGS/Imperial are added within unit_system because

1. Adding them outside prevents us from using unit system's internal _SI_system, _CGS_system, _imperial_system due to circular import
2. Using unyt's unit_system.mks_unit_system / cgs_unit_system / imperial_unit_system is risky, because we modify the systems to accomodate viscosity and angular velocity from within the unit_system - this would potentially mean we would be missing those definitions if unit_system was not imported first

extra_units now contain minimal necessary extra data for the UI schema:

```
import unyt as u

""" 
Extra units to be included in the dimensioned type schema 
(default SI, CGS, imperial units are included by default)
"""
extra_units = {
    "mass": [],
    "length": [u.mm, u.inch],
    "time": [],
    "temperature": [],
    "velocity": [],
    "area": [],
    "force": [],
    "pressure": [],
    "density": [],
    "viscosity": [],
    "angular_velocity": [],
}

```

maciej-flexcompute (approved):

PR #91, opened 2023-11-22
Schema fixes - better enum format, not exporting full output field names to schema.

awkrupka:


flow360/component/flow360_params/flow360_output.py
```diff
@@ -85,24 +137,25 @@ class SurfaceOutput(Flow360BaseModel):
...
-        alias="animationFrequency"
-    )
-    animation_frequency_offset: Optional[int] = pd.Field(alias="animationFrequencyOffset")
-    animation_frequency_time_average: Optional[Union[PositiveInt, Literal[-1]]] = pd.Field(
-        alias="animationFrequencyTimeAverage"
-    )
-    animation_frequency_time_average_offset: Optional[int] = pd.Field(
-        alias="animationFrequencyTimeAverageOffset"
-    )
+    animation: Optional[AnimationSettings] = pd.Field(alias="animation")
```
maciej-flexcompute, line 140: 1. we also need to have legacy way of defining animations. We can put it in classes SurfaceLegacy (not sure what is the best naming convention here). 

2. We need `to_solver()` function:
```python
class Surface:
   def to_solver(params: Flow360Params) -> SurfaceLegacy:
      pass
```

awkrupka: Readded legacy animations along with to_solver methods. Once we switch to the animation_settings structure on the solver side this duplication can be removed and resolved in the updater for back-compatibility

tests/params/test_outputs.py
```diff
@@ -41,57 +43,7 @@ def test_aeroacoustic_output():
...
-
-    to_file_from_file_test(output)
-
-    with pytest.raises(pd.ValidationError):
-        output = AeroacousticOutput(
-            observers=[(0, 0, 0), (0, 1, 1)],
-            animation_frequency=1,
-            animation_frequency_offset=-2,
-            patch_type="other",
-        )
```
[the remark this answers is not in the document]
awkrupka: Readded - this was removed back when Aeroacoustic output was stripped down according to documentation. Now that the fields have been readded the tests work again.

maciej-flexcompute (approved):

PR #92, opened 2023-11-28
Add aeroacoustic write per patch feature

chai-flex:


flow360/component/flow360_params/flow360_params.py
```diff
@@ -914,6 +914,7 @@ class AeroacousticOutput(Flow360BaseModel):
     animation_frequency_offset: Optional[int] = pd.Field(alias="animationFrequencyOffset")
     patch_type: Optional[str] = pd.Field("solid", const=True, alias="patchType")
     observers: List[Coordinate] = pd.Field()
+    write_per_surface_output: Optional[bool] = pd.Field(alias="writePerSurfaceOutput")
```
maciej-flexcompute, line 917: does it have default value? You can pass default as first argument to `pd.Field()`

maciej-flexcompute (approved):

PR #100, opened 2023-12-07
migrate validation logic (1st PR)

feilin-flexcompute:
add validator to do checking across multiple attributes in Flow360


flow360/component/flow360_params/flow360_params.py
```diff
@@ -4,7 +4,6 @@
 # This is a temporary measure until flow360_temp models are merged
 # pylint: disable=too-many-lines
 # pylint: disable=unused-import
-from __future__ import annotations
```
maciej-flexcompute, line 7: This is needed when you refer to type before definition

flow360/component/flow360_params/solvers.py
```diff
@@ -186,17 +186,25 @@ def _get_optional_objects(cls) -> List[str]:
...
+class TurbulenceModelConstantsSST(Flow360BaseModel):
+    """:class:`TurbulenceModelConstantsSST` class"""
+
     C_DES1: Optional[float]
     C_DES2: Optional[float]
     C_d1: Optional[float]
     C_d2: Optional[float]
 
 
+TurbulenceModelConstants = Union[TurbulenceModelConstantsSA, TurbulenceModelConstantsSST]
```
maciej-flexcompute, line 205: `TurbulenceModelConstants` seems not be used, right?

flow360/component/flow360_params/solvers.py
```diff
@@ -289,6 +297,7 @@ class TurbulenceModelSolverSST(TurbulenceModelSolver):
     """:class:`TurbulenceModelSolverSST` class"""
 
     model_type: Literal["kOmegaSST"] = pd.Field("kOmegaSST", alias="modelType", const=True)
+    model_constants: Optional[TurbulenceModelConstantsSST] = pd.Field(alias="modelConstants")
```
maciej-flexcompute, line 300: every field present as an allowed type in Flow360Params structure, will be present in webUI available option. Therefore I think the best would be to add defaults to the `TurbulenceModelConstantsSA` and `TurbulenceModelConstantsSST`

flow360/component/flow360_params/solvers.py
```diff
@@ -380,7 +390,7 @@ class TransitionModelSolver(GenericFlowSolverSettings):
         alias="maxForceJacUpdatePhysicalSteps"
     )
     order_of_accuracy: Optional[Literal[1, 2]] = pd.Field(alias="orderOfAccuracy")
-    turbulence_intensity_percent: Optional[PositiveFloat] = pd.Field(
+    turbulence_intensity_percent: Optional[pd.confloat(ge=0.03, le=2.5)] = pd.Field(
```
maciej-flexcompute, line 393: do we have default here?

flow360/component/flow360_params/solvers.py
```diff
@@ -380,7 +390,7 @@ class TransitionModelSolver(GenericFlowSolverSettings):
...
         alias="turbulenceIntensityPercent"
     )
-    N_crit: Optional[PositiveFloat] = pd.Field(alias="Ncrit")
+    N_crit: Optional[pd.confloat(ge=1, le=11)] = pd.Field(alias="Ncrit")
```
maciej-flexcompute, line 396: do we have default here?

flow360/component/flow360_params/solvers.py
feilin-flexcompute: No.

flow360/component/flow360_params/solvers.py
feilin-flexcompute: yes

flow360/component/flow360_params/solvers.py
feilin-flexcompute: yes, added.

flow360/component/flow360_params/solvers.py
feilin-flexcompute: it is used in the base class of turbulence model solver, but it is useless, so removed.

flow360/component/flow360_params/solvers.py
feilin-flexcompute: added

flow360/component/flow360_params/solvers.py
feilin-flexcompute: restored due to model_constants is required for legacy

maciej-flexcompute (approved):

PR #103, opened 2023-12-11
Added custom optional field generation

awkrupka:


flow360/component/flow360_params/solvers.py
```diff
@@ -179,6 +179,10 @@ class NavierStokesSolver(GenericFlowSolverSettings):
     def _get_field_order(cls) -> List[str]:
         return ["*", "linearSolver"]
 
+    @classmethod
+    def _get_optional_objects(cls) -> List[str]:
+        return ["linearSolver"]
```
maciej-flexcompute, line 184: we can keep it as this in this PR. But I think we can create a generic method for it, right? Detect if something is optional and detect if it is base model

flow360/component/flow360_params/params_base.py
```diff
@@ -366,6 +372,39 @@ def _swap_key_in_nested_dict(cls, dictionary, key_to_replace, replacement_key):
 
         return dictionary
 
+    @classmethod
+    def _transform_optional_field(cls, model: dict, key: str):
```
maciej-flexcompute, line 376: rename function to be more descriptive, eg:
`_generate_schema_for_optional_objects()`

Or maybe we can create a member schema class, where we can put such functions.

flow360/component/flow360_params/params_base.py
```diff
@@ -366,6 +372,39 @@ def _swap_key_in_nested_dict(cls, dictionary, key_to_replace, replacement_key):
...
+        field = model["properties"].pop(key)
+        if field is not None:
+            ref = field.get("$ref")
+            if ref is None:
+                log.warning("Trying to apply optional field transform to a non-ref field")
```
maciej-flexcompute, line 381: if this is a problem, shouldn't we raise?

flow360/component/flow360_params/params_base.py
```diff
@@ -378,6 +417,9 @@ def flow360_schema(cls):
         schema = cls.schema()
         cls._clean_schema(schema)
         cls._fix_single_allof(schema)
+        optional = cls._get_optional_objects()
```
maciej-flexcompute, line 420: `optionals`

tools/integrations/data/NavierStokesSolver.json
```diff
@@ -72,6 +63,11 @@
       "minimum": 0.01,
       "maximum": 1,
       "type": "number"
+    },
+    "_addLinearSolver": {
+      "title": "Add linear solver",
+      "type": "boolean",
+      "default": false
```
maciej-flexcompute, line 70: will we have an option to use this 'displayed_name' for this internal flag?
Specifically here, I would prefer to have `Add linear solver config` as `adding linear solver` has different meaning

maciej-flexcompute (approved):

PR #112, opened 2024-01-02
[FLPY-76] Loc path class name removal

awkrupka:


examples/case_params_with_units.py
```diff
@@ -53,13 +53,14 @@
 
 del params_as_dict["fluid_properties"]
 
-errors, warnings = validate_flow360_params_model(params_as_dict)
-pprint(errors)
+with fl.SI_unit_system:
+    errors, warnings = validate_flow360_params_model(params_as_dict)
+    pprint(errors)
```
maciej-flexcompute, line 58: validation should not be called inside context

examples/case_params_with_units.py
```diff
@@ -69,3 +70,9 @@
 
 params_as_json = params_solver.to_flow360_json()
 print(params_as_json)
+
+# Class name removal from loc
+with fl.SI_unit_system:
```
maciej-flexcompute, line 75: validation should not be called inside context

flow360/services.py
```diff
@@ -109,6 +109,18 @@ def validate_flow360_params_model(params_as_dict):
     else:
         validation_errors = validation_errors.errors()
 
+    # Check if all validation loc paths are valid params dict paths that can be traversed
+    for error in validation_errors:
+        current = params_as_dict
+        for field in error["loc"]:
```
maciej-flexcompute, line 115: why not:
```python
for field in error["loc"][:-1]:
```

awkrupka: Fixed

examples/case_params_with_units.py
awkrupka: This fails without the context - in case-editor branch we have the unit system context passed as an argument, but on develop this is the only way to prevent the "Params must be created within a unit system context (...)" error message. Once changes from case-editor are applied this can be discarded.

examples/case_params_with_units.py
awkrupka: This fails without the context - in case-editor branch we have the unit system context passed as an argument, but on develop this is the only way to prevent the "Params must be created within a unit system context (...)" error message. Once changes from case-editor are applied this can be discarded.

maciej-flexcompute (approved):

PR #114, opened 2024-01-05
[FLPY-79, 80, 84] Schema config refactor, nested optionals, vector dimensioned fields, minor visual fixes.

awkrupka:


tools/integrations/schema_generation.py
```diff
@@ -61,17 +58,29 @@ class _Freestreams(Flow360BaseModel):
...
-    """
 
+class _TurbulenceModelSolver(Flow360BaseModel):
     solver: Union[fl.SpalartAllmaras, fl.KOmegaSST, fl.NoneSolver]
 
+    class SchemaConfig(Flow360BaseModel.SchemaConfig):
+        field_order = ["*", "linearSolver"]
+        optional_objects = ["anyOf/properties/linearSolver"]
+        exclude_fields = ["anyOf/properties/linearSolver/default"]
+        root_property = "properties/solver/anyOf"
```
maciej-flexcompute, line 76: why solver, not turbulenceModelSolver?

tools/integrations/schema_generation.py
```diff
@@ -80,51 +89,212 @@ class _FluidProperties(Flow360BaseModel):
...
+    """FluidDynamicsVolumeZone type"""
+
+    model_type = pd.Field("FluidDynamics", alias="modelType", const=True)
+    reference_frame: Optional[
+        Union[
+            ReferenceFrame,
+            ReferenceFrameExpression,
+            ReferenceFrameDynamic,
+        ]
+    ] = pd.Field(alias="referenceFrame")
```
maciej-flexcompute, line 120: are we able to achieve this by using `SchemaConfig.exclude...` or something similar?

awkrupka: Issue with exclude is that we would need to delete the parent based on, let's say - the title, as members of an Union are treated as array elements - this would have to be handled via some other config field and the above just seemed like the most convenient way to limit schema-specific members.

maciej-flexcompute (approved):

PR #118, opened 2024-01-10
Added case editor service functionalities

maciej-flexcompute:


flow360/services.py
```diff
@@ -61,45 +201,27 @@ def get_default_fork(params_as_dict):
...
-def validate_flow360_params_model(params_as_dict):
+def validate_flow360_params_model(params_as_dict, unit_system_name):
     """
     Validate a params dict against the pydantic model
     """
+
+    unit_system = init_unit_system(unit_system_name)
+
+    # removing _add properties as these are only used in WebUI
+    params_as_dict = remove_properties_with_prefix(params_as_dict, "_add")
```
awkrupka, line 212: Removing the prefixed values here is okay, but during fork/retry won't we sometimes also have to handle those possibly? If we did, we probably don't want to delete them.

tests/test_flow360_params.py
```diff
@@ -176,7 +176,9 @@ def test_flow360param():
 
 def test_flow360param1():
     with flow360.SI_unit_system:
-        params = Flow360Params(freestream=FreestreamFromVelocity(velocity=10 * u.m / u.s))
+        params = Flow360Params(
+            freestream=FreestreamFromVelocity(velocity=10 * u.m / u.s), boundaries={}
```
awkrupka, line 180: Does the user have to add the empty boundaries every time now? If so, shouldn't we handle this via some default?

maciej-flexcompute: the boundaries are obligatory for every case and we need explicitly set them. Implicit settings can lead to unexpected behaviour

awkrupka (approved):

PR #119, opened 2024-01-10
added exclude handling on export to flow360 json

maciej-flexcompute:


awkrupka (approved):
Looks good - something like that could possibly be used to rewrite schema generation in a similar fashion where we don't need to define paths to excluded fields in the root schema config

PR #121, opened 2024-01-11
integrate OM6 wing case

feilin-flexcompute:
- [x] fix default value of reconstruction_gradient_limiter
- [x] add to_flow360_dict()

### DIFF
--- a/flow360/__init__.py
+++ b/flow360/__init__.py
@@ -29,6 +29,7 @@ from .component.flow360_params.boundaries import (
     SupersonicInflow,
     WallFunction,
 )
+from .component.flow360_params.flow360_output import Slice, Slices
 from .component.flow360_params.flow360_params import (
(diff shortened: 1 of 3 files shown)


flow360/__init__.py
```diff
@@ -55,7 +56,6 @@
     PorousMedium,
     ProbeMonitor,
     SliceOutput,
-    Slices,
```
maciej-flexcompute, line 58: why remove slices here?

feilin-flexcompute: because Slices is imported from flow360_output.py: https://github.com/flexcompute/Flow360/pull/121/files#diff-6f2dbf2174cca53694115555325631797a3b0f630a967bf069593bb7468177f8R32

maciej-flexcompute (approved):

PR #123, opened 2024-01-15
Release 23.3.2+ features hidden

awkrupka:


flow360/component/flow360_params/conversions.py
```diff
@@ -184,12 +186,14 @@ def get_base_viscosity():
...
-        base_heat_flux = base_density * base_velocity**3
+    if Flags.beta_features():
+
+        def get_base_heat_flux():
+            base_density = get_base_density()
+            base_velocity = get_base_velocity()
+            base_heat_flux = base_density * base_velocity**3
 
-        return base_heat_flux
+            return base_heat_flux
```
maciej-flexcompute, line 196: no need to mark this as beta feature

flow360/component/flow360_params/turbulence_quantities.py
```diff
@@ -1,245 +1,240 @@
 """
 Turbulence quantities parameters
 """
-from abc import ABCMeta
-from typing import Optional, Union
+from globals.flags import Flags
```
maciej-flexcompute, line 4: can we just import this file under beta_features() but its definition leave unchanged?

globals/flags.py
maciej-flexcompute, line 1: maybe we can just use ENV variable to set flag? Then we can run tests in both scenarios

flow360/component/flow360_params/turbulence_quantities.py
awkrupka: We could, but this way the user will be able to import it directly (not from flow360 base module but from flow360.component.flow360_params.turbulence_quantities) and still use it which may not be what we want

flow360/component/flow360_params/unit_system.py
```diff
@@ -771,7 +781,8 @@ class UnitSystem(pd.BaseModel):
     density: DensityType = pd.Field()
     viscosity: ViscosityType = pd.Field()
     angular_velocity: AngularVelocityType = pd.Field()
-    heat_flux: HeatFluxType = pd.Field()
+    if Flags.beta_features():
+        heat_flux: HeatFluxType = pd.Field()
```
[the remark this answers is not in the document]
awkrupka: Reverted, was listed in PRs to remove from 23.3.2 so i figured it should be blocked

flow360/component/flow360_params/unit_system.py
```diff
@@ -789,9 +800,11 @@ class UnitSystem(pd.BaseModel):
         "density",
         "viscosity",
         "angular_velocity",
-        "heat_flux",
     ]
 
+    if Flags.beta_features():
+        _dim_names.append("heat_flux")
```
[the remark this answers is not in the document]
awkrupka: Reverted

flow360/component/flow360_params/unit_system.py
```diff
@@ -908,26 +921,26 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
-        flow360_force_unit,
-        flow360_pressure_unit,
-        flow360_density_unit,
-        flow360_viscosity_unit,
-        flow360_angular_velocity_unit,
-        flow360_heat_flux_unit,
-    ]
-}
+if Flags.beta_features():
+    flow360_heat_flux_unit = Flow360HeatFluxUnit()
```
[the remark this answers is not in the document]
awkrupka: Reverted

flow360/component/flow360_params/unit_system.py
```diff
@@ -985,7 +999,8 @@ def __init__(self):
         conversion_system["pressure"] = "flow360_pressure_unit"
         conversion_system["viscosity"] = "flow360_viscosity_unit"
         conversion_system["angular_velocity"] = "flow360_angular_velocity_unit"
-        conversion_system["heat_flux"] = "flow360_heat_flux_unit"
+        if Flags.beta_features():
+            conversion_system["heat_flux"] = "flow360_heat_flux_unit"
```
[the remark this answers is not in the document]
awkrupka: Reverted

flow360/component/flow360_params/unit_system.py
```diff
@@ -1017,7 +1032,8 @@ class _PredefinedUnitSystem(UnitSystem):
     density: DensityType = pd.Field(exclude=True)
     viscosity: ViscosityType = pd.Field(exclude=True)
     angular_velocity: AngularVelocityType = pd.Field(exclude=True)
-    heat_flux: HeatFluxType = pd.Field(exclude=True)
+    if Flags.beta_features():
+        heat_flux: HeatFluxType = pd.Field(exclude=True)
```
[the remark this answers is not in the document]
awkrupka: Reverted

flow360/component/flow360_params/conversions.py
awkrupka: Reverted

globals/flags.py
awkrupka: We could explore this, but I think what we want would be

1. Set beta_features to True
2. Import flow360, run tests
3. Set beta_features to False
4. Import flow360 again, run tests

and its step 4 that's the main issue, as we need to reimport flow360 at some point when running tests. ENV variables would have the advantage of not requiring the globals.flags module, at the cost of setting the state permanently (right now feature flags are disabled by default unless we explicitly set them before import)

it would work if we ran a different command like

<set variable> && pytest -rA && <set variable> && pytest -rA

awkrupka: Changed to environment variable: to run tests with the beta_features on we run

FLOW360_BETA_FEATURES=True pytest -rA

flow360/component/flow360_params/boundaries.py
```diff
@@ -9,9 +10,11 @@
 
 from ..types import Axis, PositiveFloat, PositiveInt, Vector
 from .params_base import Flow360BaseModel
-from .turbulence_quantities import TurbulenceQuantitiesType
 from .unit_system import VelocityType
 
+if os.environ.get("FLOW360_BETA_FEATURES", False):
```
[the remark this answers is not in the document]
awkrupka: Fixed, added a wrapper class so we still call Flags.beta_feature() in checks

config/flags.py
maciej-flexcompute, line 1: this file is outside flow360. When publishing to PyPI this path will be excluded so I think the package will not be able to init

maciej-flexcompute (approved):

PR #130, opened 2024-01-18
tagged linearSolver as bete feature and reverted linearSolverConfig

maciej-flexcompute:


tools/integrations/data/turbulence-model/json-schema.json
```diff
@@ -4,7 +4,7 @@
   "additionalProperties": false,
   "anyOf": [
     {
-      "title": "Spalart- allmaras",
+      "title": "Spalart  allmaras",
```
maciej-flexcompute, line 7: options are not respected here

awkrupka (approved):

PR #134, opened 2024-01-22
Create automatic schema update workflow

awkrupka:


.github/workflows/schema-update.yml
```diff
@@ -0,0 +1,41 @@
...
+      env:
+        API_TOKEN_GITHUB: ${{ secrets.API_TOKEN_GITHUB }}
+      with:
+        source_folder: 'tools/integrations/data/'
+        destination_repo: 'flexcompute/Flow360WebClient'
+        destination_folder: 'src/components/FormGenerator/schemas/case/'
+        destination_base_branch: 'develop'
+        destination_head_branch: 'schema-auto-update'
+        user_email: 'andrzej@flexcompute.com'
+        user_name: 'andrzej-krupka'
```
maciej-flexcompute, line 41: maybe we can create a separate github account for auto updates?

awkrupka: That's a good idea - perhaps automation@flexcompute.com / deployment@flexcompute.com with the corresponding GH handle?

.github/workflows/schema-update.yml
```diff
@@ -0,0 +1,41 @@
+# Used to automatically create a pull request with updated schemas
+name: Automatic schema update
+
+on:
+  push:
+    branches:
+      - develop
```
psmolaga, line 7: You can automate triggering based on path changes. By using a build trigger, you can remove the step of checking, then.

```
on:
  push:
    paths:
      - folder1/**
```

PR #139, opened 2024-01-25
Fix wrong velocity type for subsonic inflow

awkrupka:


flow360/component/flow360_params/boundaries.py
```diff
@@ -153,7 +153,7 @@ class SubsonicInflow(BoundaryWithTurbulenceQuantities):
         total_pressure_ratio: PositiveFloat = pd.Field(alias="totalPressureRatio")
         total_temperature_ratio: PositiveFloat = pd.Field(alias="totalTemperatureRatio")
         ramp_steps: Optional[PositiveInt] = pd.Field(alias="rampSteps")
-        velocity_direction: Optional[BoundaryVelocityType] = pd.Field(alias="velocityDirection")
+        velocity_direction: Optional[Axis] = pd.Field(alias="velocityDirection")
```
maciej-flexcompute, line 156: I think it can also be expression, so it should be `BoundaryAxisType`

maciej-flexcompute (approved):

PR #145, opened 2024-01-26
Added symmetryPlane

NasserFlexCompute:


flow360/component/flow360_params/boundaries.py
```diff
@@ -52,6 +52,10 @@ class SlipWall(Boundary):
 
     type: Literal["SlipWall"] = pd.Field("SlipWall", const=True)
 
+class SymmetryPlane(Boundary):
```
maciej-flexcompute, line 55: If this is next release feature it should be closed in Flags.beta_feature() if statement. So it is easy to switch between current version and next version. On release we will remove all beta_features() statements.

PR #147, opened 2024-01-29
Set boundary temperature to TemperatureType

yifan-flex:
Need to create variable types for volumeZones: volumetricHeatSource, heatCapacity, thermalConductivity


maciej-flexcompute:
@yifan-flex can you check if this is resolved in SimulationParams?

PR #149, opened 2024-01-29
Removed the legacy version of UDD

benflexcompute:
The outputLaw is a legacy input.

### DIFF
--- a/flow360/component/flow360_params/volume_zones.py
+++ b/flow360/component/flow360_params/volume_zones.py
@@ -29,6 +29,8 @@ class ReferenceFrameBase(Flow360BaseModel):
     """Base reference frame class"""
 
     model_type: str = pd.Field(alias="modelType")
+    center: LengthType.Point = pd.Field(alias="centerOfRotation")
+    axis: Axis = pd.Field(alias="axisOfRotation")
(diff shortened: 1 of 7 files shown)


flow360/component/flow360_params/volume_zones.py
```diff
@@ -51,7 +51,6 @@ class ReferenceFrameDynamic(Flow360BaseModel):
 
     center: LengthType.Point = pd.Field(alias="centerOfRotation")
     axis: Axis = pd.Field(alias="axisOfRotation")
-    is_dynamic: bool = pd.Field(True, alias="isDynamic", const=True)
```
maciej-flexcompute, line 54: are you sure we don't need this? How the solver knows that no omega is required here?

benflexcompute: According to the logic in the solver side preprocessing, we add 0 for theta, omega and omegaDot if they are not specified.

benflexcompute: the use of "isDynamic" has been deprecated for quite some time.

maciej-flexcompute (approved):

PR #157, opened 2024-01-31
added discriminator to reference frame

maciej-flexcompute:
The PR doesn't pass unittests yet

### DIFF
--- a/flow360/component/flow360_params/volume_zones.py
+++ b/flow360/component/flow360_params/volume_zones.py
@@ -25,7 +25,17 @@ from .params_base import Flow360BaseModel
 from .unit_system import AngularVelocityType, LengthType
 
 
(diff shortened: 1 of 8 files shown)


flow360/component/flow360_params/volume_zones.py
```diff
@@ -25,7 +25,17 @@
 from .unit_system import AngularVelocityType, LengthType
 
 
-class ReferenceFrameDynamic(Flow360BaseModel):
+class ReferenceFrameBase(Flow360BaseModel):
```
benflexcompute, line 28: Same name is also used here in my PR:
https://github.com/flexcompute/Flow360/pull/149/files#diff-10ebb6a178a1163f7ca99a742787d103081243666f6dc4523faeecf02fbde505R28:~:text=class%20ReferenceFrameBase(,class%20for%20setting

PR #158, opened 2024-01-31
Partially fixed the element count check for vector

benflexcompute:


flow360/component/flow360_params/unit_system.py
```diff
@@ -338,7 +338,7 @@ def validate(vec_cls, value):
                 value = _unit_object_parser(value, [u.unyt_array, _Flow360BaseUnit.factory])
                 value = _is_unit_validator(value)
 
-                if not isinstance(value, Collection) and len(value) != 3:
```
awkrupka, line 341: This was a logic bug, but since we still want to check whether the incoming value is a collection shouldn't we be checking 

`not isinstance(value, Collection) or len(value) != 3`?

Perhaps the type check can be moved into a separate if statement but without it for erroneous initialization values like "True" we fail on trying to call len(bool) with the message

`object of type 'bool' has no len() (type=type_error)`

instead of the more verbose

`arg 'True' needs to be a collection of 3 values`

awkrupka (changes requested):

maciej-flexcompute (approved):

awkrupka (approved):

PR #159, opened 2024-01-31
Modified Flow360Client based on the moved testCases

NasserFlexCompute:
Need to fix the unit tests

### DIFF
--- a/flow360/component/flow360_params/boundaries.py
+++ b/flow360/component/flow360_params/boundaries.py
@@ -8,6 +8,7 @@ from typing import Literal, Optional, Tuple, Union
 import pydantic as pd
 from pydantic import StrictStr
 
+from flow360.component.flow360_params.unit_system import PressureType
 from flow360.flags import Flags
 
 from ..types import Axis, NonNegativeFloat, PositiveFloat, PositiveInt, Vector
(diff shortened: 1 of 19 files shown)


flow360/component/flow360_params/boundaries.py
```diff
@@ -298,6 +313,25 @@ class RotationallyPeriodic(Boundary):
...
+        velocity: Optional[BoundaryVelocityType] = pd.Field(alias="Velocity")
+
+
+if Flags.beta_features():
+
+    class PressureOutflow(BoundaryWithTurbulenceQuantities):
+        """Outflow pressure for incompressible solver"""
+
+        type: Literal["PressureOutflow"] = pd.Field("PressureOutflow", const=True)
+        static_pressure: Optional[NonNegativeFloat] = pd.Field(alias="staticPressure")
```
feilin-flexcompute, line 331: I think pressure should be dimensional

flow360/component/flow360_params/solvers.py
```diff
@@ -510,6 +515,8 @@ class TransitionModelSolver(GenericFlowSolverSettings):
         )
 
 
+TransitionModelSolverType = Union[NoneSolver, TransitionModelSolver]
+
```
maciej-flexcompute, line 519: do we really need NoneSolver for Transition? I think preference here is just to omit this parameter.

NasserFlexCompute: Sure. will remove it.

maciej-flexcompute (approved):

PR #171, opened 2024-02-09
Default CFL

benflexcompute:


flow360/component/flow360_params/time_stepping.py
```diff
@@ -105,6 +120,17 @@ class UnsteadyTimeStepping(BaseTimeStepping):
...
+        displayed="CFL", options=["Ramp CFL", "Adaptive CFL"]
+    )
+
+    # pylint: disable=C0103
+    def __init__(self, **data):
+        super().__init__(**data)
+        if self.CFL is None or (isinstance(self.CFL, AdaptiveCFL) and self.CFL.asked_for_default()):
+            self.CFL = AdaptiveCFL.default_unsteady()
+        elif isinstance(self.CFL, RampCFL) and self.CFL.asked_for_default():
+            self.CFL = RampCFL.default_unsteady()
```
maciej-flexcompute, line 133: a more pydantic way would be to use validators instead of constructor, for examples:
```python
    @pd.root_validator(pre=True)
    def set_default_cfl(cls, values):
        print(values)
        if "CFL" in values:
            cfl = values['CFL']
            if isinstance(cfl, RampCFL):
                cfl_as_dict = cfl.dict(exclude_unset=True)
                print(cfl_as_dict)
        return values

```

here you can access what user provided, for example, user might specify only ramp->final but the rest will be default. `cfl_as_dict` is exactly what user provided. You want to set  values['CFL'] to default and overwrite it with `cfl_as_dict`. Note that type(values['CFL']) should be `RampCFL` when you return as no more validation will happen on this object.

For adaptive default, I would put them directly in CFL property definition:
```python
CFL: Optional[Union[RampCFL, AdaptiveCFL]] = pd.Field(AdaptiveCFL(min=..., max=..., ), displayed="CFL", options=["Ramp CFL", "Adaptive CFL"])
```

tools/integrations/data/freestream/json-schema.json
```diff
@@ -221,8 +237,16 @@
...
-          "type": "number"
+          "anyOf": [
+            {
+              "type": "integer",
+              "const": -1
+            },
+            {
+              "type": "number",
+              "exclusiveMinimum": 0
+            }
```
maciej-flexcompute, line 248: this can break our UI or can display in a weird way. Let's revert in this PR

flow360/component/flow360_params/time_stepping.py
benflexcompute: Sure I changed with root_validator way. But we still go through pydantic validation since we have pre=True right?

flow360/component/flow360_params/time_stepping.py
```diff
@@ -35,18 +36,18 @@ class RampCFL(Flow360BaseModel):
...
 
     @classmethod
-    def default_steady(cls):
+    def default_unsteady(cls):
         """
-        returns default steady CFL settings
+        returns default unsteady Ramp CFL settings
         """
-        return cls(initial=5, final=200, ramp_steps=40)
+        return cls(initial=1, final=1e6, ramp_steps=30)  ## Unknown value source
```
maciej-flexcompute, line 43: can we remove this comment?

flow360/component/flow360_params/time_stepping.py
```diff
@@ -95,6 +110,25 @@ class SteadyTimeStepping(BaseTimeStepping):
     model_type: Literal["Steady"] = pd.Field("Steady", alias="modelType", const=True)
     physical_steps: Literal[1] = pd.Field(1, alias="physicalSteps", const=True)
     time_step_size: Literal["inf"] = pd.Field("inf", alias="timeStepSize", const=True)
+    CFL: Optional[Union[RampCFL, AdaptiveCFL]] = pd.Field(
+        displayed="CFL",
+        options=["Ramp CFL", "Adaptive CFL"],
+        default=AdaptiveCFL().default_steady(),
```
maciej-flexcompute, line 116: it is class method, so should be `AdaptiveCFL.default_steady()`

tests/params/test_time_stepping.py
```diff
@@ -24,61 +24,97 @@ def change_test_dir(request, monkeypatch):
...
 
-        params = Flow360Params(
-            geometry={"meshUnit": "mm", "refArea": "m**2"},
-            boundaries={},
-            fluid_properties=fl.air,
-            freestream={"modelType": "FromMach", "temperature": 288.15, "Mach": 1, "mu_ref": 1},
-            time_stepping=ts,
-        )
+    # exported_json = json.loads(params.flow360_json())
+    # assert "meshUnit" not in exported_json["geometry"]
```
maciej-flexcompute, line 69: why this test is commented out?

maciej-flexcompute (approved):

PR #176, opened 2024-02-13
Move over solver default JSON

benflexcompute:


flow360/component/flow360_params/flow360_output.py
```diff
@@ -423,8 +426,8 @@ def validate_monitor(cls, values):
 class MonitorOutput(Flow360BaseModel):
     """:class:`MonitorOutput` class"""
 
-    monitors: Optional[Monitors] = pd.Field()
-    output_fields: Optional[CommonOutputFields] = pd.Field(alias="outputFields")
+    monitors: Optional[Monitors] = pd.Field(default={})
```
maciej-flexcompute, line 429: Why default to empty dict? This is optional parameter, so I think monitors=None make more sense, so no default at all.

flow360/component/flow360_params/flow360_params.py
```diff
@@ -491,10 +491,16 @@ class Geometry(Flow360BaseModel):
...
+    ref_area: Optional[AreaType.Positive] = pd.Field(
+        alias="refArea", displayed="Reference area", default_factory=lambda: 1
+    )
+    moment_center: Optional[LengthType.Point] = pd.Field(
+        alias="momentCenter", default_factory=lambda: (0.0, 0.0, 0.0)
+    )
+    moment_length: Optional[LengthType.Moment] = pd.Field(
+        alias="momentLength", default_factory=lambda: (1.0, 1.0, 1.0)
+    )
+    mesh_unit: Optional[LengthType] = pd.Field(alias="meshUnit", default_factory=lambda: 1)
```
andrzej-krupka, line 503: In pydantic V1 default_factory causes alot of issues, most importantly this one https://github.com/pydantic/pydantic/issues/1717 which breaks the updater which uses parse_obj extensively. There are other more subtle issues with default_factories on optional fields like confusing error messages.

@maciej-flexcompute maybe we should consider migrating to pydantic v2

flow360/component/flow360_params/flow360_output.py
```diff
@@ -530,10 +532,19 @@ class AeroacousticOutput(Flow360BaseModel, AnimatedOutput):
     """
 
     patch_type: Optional[str] = pd.Field("solid", const=True, alias="patchType")
-    observers: List[Coordinate] = pd.Field()
+    animation_frequency: Optional[PositiveInt] = pd.Field(
```
andrzej-krupka, line 535: Should this be animated @maciej-flexcompute? Documentation doesn't mention anything: https://docs.flexcompute.com/projects/flow360/en/latest/solverConfiguration/solverConfiguration.html#aeroacousticoutput

flow360/component/flow360_params/flow360_output.py
benflexcompute: monitors is actually not optional I will remove that.

Instead I will make Monitor/Slice/...Output optional with no defaults since they are not required.

flow360/component/flow360_params/flow360_output.py
benflexcompute: My bad. I will fix it to be consistent with documentation.

benflexcompute: Done

flow360/component/flow360_params/flow360_params.py
```diff
@@ -491,10 +492,17 @@ class Geometry(Flow360BaseModel):
...
+        alias="refArea", displayed="Reference area", default_factory=lambda: 1
+    )
+    moment_center: Optional[LengthType.Point] = pd.Field(
+        alias="momentCenter", default_factory=lambda: (0.0, 0.0, 0.0)
+    )
+    ##Note: moment_length does not allow negative components I failed to enforce that here after attempts
+    moment_length: Optional[LengthType.Moment] = pd.Field(
+        alias="momentLength", default_factory=lambda: (1.0, 1.0, 1.0)
+    )
+    mesh_unit: Optional[LengthType.Positive] = pd.Field(alias="meshUnit", default_factory=lambda: 1)
```
maciej-flexcompute, line 505: lets remove `default_factory` completely and implement defaults in `to_solver()`.

flow360/component/flow360_params/flow360_params.py
```diff
@@ -506,6 +514,7 @@ def to_solver(self, params: Flow360Params, **kwargs) -> Geometry:
     # pylint: disable=missing-class-docstring,too-few-public-methods
     class Config(Flow360BaseModel.Config):
         allow_but_remove = ["meshName", "endianness"]
+        exclude_on_flow360_export = ["mesh_unit"]
```
maciej-flexcompute, line 517: `mesh_unit` exclude is implemented in to_solver() method. It is a duplication of logic here.

flow360/component/flow360_params/params_base.py
```diff
@@ -289,6 +299,28 @@ def handle_deprecated_aliases(cls, values):
...
+        return values
+
+    @classmethod
+    def _handle_conflicting_fields(cls, values, conflicting_field: Conflicts = None):
+        conflicting_field1_value = values.get(conflicting_field.field1, None)
+        conflicting_field2_value = values.get(conflicting_field.field2, None)
+        if conflicting_field1_value is not None and conflicting_field2_value is not None:
+            raise ValueError(
+                f"{conflicting_field.field1} and {conflicting_field.field2} cannot be specified at the same time."
+            )
```
maciej-flexcompute, line 320: make sure that it works for both name and alias name

flow360/component/flow360_params/flow360_params.py
```diff
@@ -1076,6 +1096,8 @@ def _init_no_context(self, filename, legacy_fallback=False, **kwargs):
             with UnitSystem.from_dict(**unit_system):
                 super().__init__(**model_dict)
         else:
+            ## Legacy version always uses flow360 unit
+            unit_system_manager.set_current(flow360_unit_system)
```
maciej-flexcompute, line 1100: this is not right. The legacy updater should decide. The problem is with "comments" section produces by webUI which will use SI unit system. Also, `set_current` should not be used, because it goes around lock. It probably should be renamed to `_set_current`

flow360/component/flow360_params/params_base.py
```diff
@@ -1163,6 +1195,10 @@ def names(self) -> List[str]:
         """return names of all boundaries"""
         return [k for k, _ in self if k not in [COMMENTS, TYPE_TAG_STR]]
 
+    def items(self) -> List[str]:
+        """return names of all boundaries"""
+        return [obj for k, obj in self if k not in [COMMENTS, TYPE_TAG_STR]]
+
```
maciej-flexcompute, line 1201: with dicts, items should return `key, item` tuple, right, https://docs.python.org/3.12/library/stdtypes.html#dict.items

If you want to access item by name, there is `[]` accessor implemented.

flow360/component/flow360_params/solvers.py
```diff
@@ -75,32 +83,34 @@ class LinearSolver(Flow360BaseModel):
...
+    @pd.root_validator
+    def check_only_one_tolerance(cls, values):
+        """Check that we only allow one type of tolerance specified"""
+        absolute_tolerance, relative_tolerance = values.get("absolute_tolerance"), values.get(
+            "relative_tolerance"
+        )
+        if absolute_tolerance is not None and relative_tolerance is not None:
+            raise ValueError(
+                "Only one of absolute_tolerance or relative_tolerance can be specified"
+            )
```
maciej-flexcompute, line 100: why not use the new `Conflicts` here?

flow360/component/flow360_params/solvers.py
```diff
@@ -514,6 +516,24 @@ class TransitionModelSolver(GenericFlowSolverSettings):
             displayed="Linear solver config",
         )
 
+    # pylint: disable=missing-class-docstring,too-few-public-methods
+    class Config(Flow360BaseModel.Config):
+        conflicting_fields = [Conflicts(field1="N_crit", field2="turbulence_intensity_percent")]
```
maciej-flexcompute, line 521: Let's put `Config` as the last item in class definition.

flow360/component/flow360_params/solvers.py
```diff
@@ -514,6 +516,24 @@ class TransitionModelSolver(GenericFlowSolverSettings):
...
+        Convert turbulenceIntensityPercent to Ncrit
+        """
+
+        if self.turbulence_intensity_percent is not None:
+            Ncrit_converted = -8.43 - 2.4 * np.log(
+                0.025 * np.tanh(self.turbulence_intensity_percent / 2.5)
+            )
+            self.turbulence_intensity_percent = None
+            self.N_crit = Ncrit_converted
+        return super().to_solver(self, exclude=["turbulence_intensity_percent"], **kwargs)
```
maciej-flexcompute, line 535: this is the right way of handling to solver conversions and other defaults

tests/params/test_freestream.py
```diff
@@ -40,10 +40,12 @@ def test_freesteam():
         to_file_from_file_test(fs)
 
         p = fl.Flow360Params(
+            geometry=fl.Geometry(),
```
maciej-flexcompute, line 43: I don't like requesting user to specify empty structure. Let's handle it differently to let user omitting this parameter.

flow360/component/flow360_params/flow360_params.py
benflexcompute: Sure

flow360/component/flow360_params/flow360_params.py
benflexcompute: Good catch.

flow360/component/flow360_params/params_base.py
benflexcompute: Done. Just note that in the future we should not add default value for conflicting fields otherwise `to_file_from_file_test()` will fail.

flow360/component/flow360_params/flow360_params.py
benflexcompute: Sure. Removed.

flow360/component/flow360_params/params_base.py
benflexcompute: Good point. Removed .items() method.

flow360/component/flow360_params/solvers.py
benflexcompute: Good point.

flow360/component/flow360_params/solvers.py
benflexcompute: Done.

BTW I found that setting `N_crit` before setting `turbulence_intensity_percent` to None (in the `to_solver `function) reports validation error? Seems that it constantly check whether the `Config` is satisfied.

tests/params/test_freestream.py
benflexcompute: Removed and added the empty as default

maciej-flexcompute (approved):

PR #178, opened 2024-02-15
fixed updater for sliding interfaces and fluid properties from comments

maciej-flexcompute:


flow360/component/flow360_params/flow360_params.py
```diff
@@ -1713,8 +1713,8 @@ def update_model(self) -> Flow360BaseModel:
                     "iso_surface_output": try_update(self.iso_surface_output),
                     "monitor_output": self.monitor_output,
                     "aeroacoustic_output": self.aeroacoustic_output,
-                    "fluid_properties": None,
-                    "volume_zones": self.volume_zones,
+                    "fluid_properties": params.get("fluid_properties"),
```
andrzej-krupka, line 1716: I see the issue with params.update() with previously set fluid_properties, volume_zones and bet_disks fields set to None, but shouldn't we just remove them from this params.update() call and use the updated fields from code between lines 1669-1689?

andrzej-krupka: Fixed

andrzej-krupka (approved):

PR #179, opened 2024-02-19
Added results handling for all type of results

maciej-flexcompute:
Added results handling for all type of results:
- [x] added all CSV files
- [x] added all tar.gz files
- [x] added handling for list results (monitors, UDDs)
- [x] added dimensions handling for actuator disk and bet disk
- [x] added unit tests for `results.download()` and `results.set_downloader()` functionality
- [x] added examples for monitors
- [x] added examples for UDD
- [x] added examples for convergence
- [x] added example for forces
- [x] added example for volumetric and surface tar.gz 
- [x] replace multiple case with alpha-sweep example


andrzej-krupka:
Mostly finished, known issues are:

- results/heat_transfer.py and results/aeroacoustics.py are marked with TODO, no corresponding files exist on auto_test_1
- on Windows, python 3.11 (only this env) download test fails with permissions error when writing to a temp file
- alpha_sweep example seems to be missing from auto_test_1 cases

Other points are done, along the way some bugs were identified with the download and pydantic model logic which have been fixed since then. Since we're using pandas now, the minimum python version needed to be bumped to 3.9

flow360/component/results/case_results.py
```diff
@@ -0,0 +1,891 @@
...
+from ..flow360_params.flow360_params import Flow360Params
+from ..flow360_params.unit_system import (
+    Flow360UnitSystem,
+    ForceType,
+    MomentType,
+    PowerType,
+    is_flow360_unit,
+)
+
+TMP_DIR = tempfile.mkdtemp()
```
maciej-flexcompute, line 32: This is not ideal because we don't clean up temp files. Need to find a better way in the future

flow360/component/flow360_params/flow360_output.py
```diff
@@ -280,7 +280,8 @@ def schema_extra(schema, model):
     def to_solver(self, params, **kwargs) -> Slice:
         solver_values = self._convert_dimensions_to_solver(params, **kwargs)
         fields = solver_values.pop("output_fields")
-        fields = [to_short(field) for field in fields]
+        if fields is not None:
```
benflexcompute, line 285: I think now all `output_fields` has default of [].

flow360/component/flow360_params/unit_system.py
```diff
@@ -342,8 +351,16 @@ def validate(vec_cls, value):
                     isinstance(value, _Flow360BaseUnit) and isinstance(value.val, Collection)
                 )
 
-                if not is_collection or len(value) != 3:
-                    raise TypeError(f"arg '{value}' needs to be a collection of 3 values")
+                if length is None:
```
benflexcompute, line 354: I vaguely remember that this does not work if I give a scalar value/dict/set as argument. And I also vaguely remember that if I pass in a flow_360_unit_system value ((0,1,0)*flow_360_unit_system.length for example) this also breaks.

But I might be wrong and this has been fixed.

benflexcompute: I sent a commit showcasing the ones that does not work.

examples/results/actuator_disk.py
```diff
@@ -0,0 +1,62 @@
...
+# 2               0           20  1.605465e+09  1.066131e+08  1.605465e+09  kg*m**2/s**3  kg*m/s**2  kg*m**2/s**2
+# 3               0           30  1.655555e+09  1.066131e+08  1.655555e+09  kg*m**2/s**3  kg*m/s**2  kg*m**2/s**2
+# 4               0           40  1.678046e+09  1.066131e+08  1.678046e+09  kg*m**2/s**3  kg*m/s**2  kg*m**2/s**2
+# 5               0           50  1.708671e+09  1.066131e+08  1.708671e+09  kg*m**2/s**3  kg*m/s**2  kg*m**2/s**2
+# 6               0           60  1.787372e+09  1.066131e+08  1.787372e+09  kg*m**2/s**3  kg*m/s**2  kg*m**2/s**2
+# 7               0           70  1.932056e+09  1.066131e+08  1.932056e+09  kg*m**2/s**3  kg*m/s**2  kg*m**2/s**2
+
+
+# download resuts:
+results.set_destination(use_case_name=True)
```
chai-flex, line 57: if set_destination is not called, what is the behavior? Does it just use the local folder? We should probably have a default

examples/results/actuator_disk.py
```diff
@@ -0,0 +1,62 @@
...
+results.set_downloader(actuator_disks=True)
```
chai-flex, line 58: In the following case:
set_destination()
set_downloader(keys)
download()

can we just have?
set_destination
download(key1, key2, overwrite=True)

What is the point of set_downloader and download?

examples/results/actuator_disk.py
```diff
@@ -0,0 +1,62 @@
...
+case.wait()
+
+
+results = case.results
+
+# set values needed for units conversion:
+case.params.geometry.mesh_unit = 1 * u.m
+case.params.fluid_properties = fl.air
+
+print(results.actuator_disks.as_dataframe())
```
chai-flex, line 28: Is data automatically downloaded when as_dataframe is called?

examples/results/forces.py
```diff
@@ -0,0 +1,60 @@
...
+
+OM6wing.get_files()
+
+# submit mesh
+volume_mesh = fl.VolumeMesh.from_file(OM6wing.mesh_filename, name="OM6wing-mesh")
+volume_mesh = volume_mesh.submit()
+
+# submit case using json file
+params = fl.Flow360Params(OM6wing.case_json)
+case = volume_mesh.create_case("OM6wing-example", params)
```
chai-flex, line 12: can we have a submit_case method to bypass create_case + submit?

examples/results/forces.py
```diff
@@ -0,0 +1,60 @@
...
+case = case.submit()
+
+# wait until the case finishes execution
+case.wait()
```
chai-flex, line 16: Can we have automatic wait when results are requested?

flow360/examples/actuatorDisk/flow360.json
```diff
@@ -38,10 +38,6 @@
       "animationFrequency": -1,
       "nodeForcesPerUnitArea": true
     },
-    "sliceOutput": {
-      "outputFormat": "paraview",
-      "animationFrequency": -1
-    },
```
benflexcompute, line 44: Why is this deleted? To avoid user copying and using senseless JSON?

benflexcompute: I remember this actually serves as a good test for updater so it handles all situations. See the PR I merged yesterday for numericalDissipation.

examples/results/actuator_disk.py
maciej-flexcompute: yes. All CSV `results.<results_type>.<access_function>` uses lazy loading.

examples/results/forces.py
maciej-flexcompute: this is part of a larger design discussion. You can do: case.copy(), modify params, submit. **submit** is sending data to cloud. So I think making this behaviour patter for every operation will be more intuitive with users.

examples/results/forces.py
maciej-flexcompute: Actually, you can get partial results before case is finished

flow360/cloud/s3_utils.py
```diff
@@ -79,15 +72,27 @@ def create_base_folder(
 
     """
 
-    if to_folder != ".":
-        to_file = os.path.join(to_folder, os.path.basename(target_name))
+    if to_file is not None and to_folder != ".":
+        raise Flow360ValueError(
```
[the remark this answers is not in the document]
maciej-flexcompute: `to_file` can be `./path/to/filename` or `/path/to/filename`. Then it can be tricky to detect what user wanted

flow360/component/case.py
```diff
@@ -480,6 +498,30 @@ def has_bet_disks(self):
                 return True
         return False
 
+    def has_isosurfaces(self):
```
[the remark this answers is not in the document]
maciej-flexcompute: volume is true by default as well. Can we turn it off? has_slice is a good point

flow360/component/case.py
```diff
@@ -582,276 +624,248 @@ def create(
         )
         return new_case
 
+    def wait(self, refresh_rate_seconds=2, timeout_minutes=60):
```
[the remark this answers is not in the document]
maciej-flexcompute: good point, will remove

flow360/examples/actuatorDisk/flow360.json
maciej-flexcompute: ok

flow360/component/case.py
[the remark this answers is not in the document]
benflexcompute: Yes. I think now we have no default volume output anymore. I need to merge https://github.com/flexcompute/Flow360Core/pull/1197/files so the solver not longer assume there always is volumeOutput in the JSON though.

examples/results/actuator_disk.py
maciej-flexcompute: removed set_downloader function

flow360/component/results/case_results.py
andrzej-krupka: I suggest we move all globally-accessible temporary directories to a separate manager, call it TemporaryResourceManager and use the `__enter__` and `__exit__` methods to allocate and deallocate temporary resources as well as manage access to them for other systems, similar to how we use UserConfig / UnitSystem right now. Should we fix this now or later?

andrzej-krupka (approved):
Looks good apart from the temporary directory issue i commented on - we can implement it later if need be.

flow360/component/results/case_results.py
maciej-flexcompute: how about error, will it free resources when errors happen?

PR #195, opened 2024-03-08
Fix the reconstructionGradientLimiter updater

benflexcompute:


flow360/component/flow360_params/solvers.py
```diff
@@ -547,6 +547,12 @@ class TurbulenceModelSolverLegacy(TurbulenceModelSolver, LegacyModel):
     rotation_correction: Optional[bool] = pd.Field(alias="rotationCorrection")
 
     def update_model(self):
+        if self.reconstruction_gradient_limiter is None:
+            if self.model_type == "SpalartAllmaras":
```
chai-flex, line 551: Is there a model object type we can compare against? 

We should only define the mapping from model type string to object type in 1 place and then have the object type used everywhere

flow360/component/flow360_params/solvers.py
```diff
@@ -559,16 +563,21 @@ def update_model(self):
...
             "gridSizeForLES": self.grid_size_for_LES,
             "quadraticConstitutiveRelation": self.quadratic_constitutive_relation,
-            "reconstructionGradientLimiter": self.reconstruction_gradient_limiter,
             "modelConstants": self.model_constants,
         }
 
         try_set(model, "rotationCorrection", self.rotation_correction)
+        
+        if self.reconstruction_gradient_limiter is not None:
+            model["reconstructionGradientLimiter"] = self.reconstruction_gradient_limiter
```
benflexcompute, line 572: It is still not ideal in the sense that not all default values are set by the SA/KOmega.

chai-flex (approved):

PR #198, opened 2024-03-10
Added versioned schemas for 24.2

awkrupka:


tools/integrations/schema_generation.py
maciej-flexcompute, line 1: let's not generate schema per version because repo size will grow significantly. Let's use release branches instead

awkrupka: I'll need to adjust the automatic schema generation workflow in this case - currently it assumes pulling schemas from develop branch.

awkrupka: Closing this PR and making changes to release branches.

PR #201, opened 2024-03-11
Add constants Field in expression Initial condition

benflexcompute:


flow360/component/flow360_params/flow360_params.py
```diff
@@ -1401,7 +1401,9 @@ def update_model(self) -> Flow360BaseModel:
 class FreestreamLegacy(LegacyModel):
     """:class: `FreestreamLegacy` class"""
 
-    Reynolds: Optional[PositiveFloat] = pd.Field()
+    Reynolds: Optional[Union[pd.confloat(gt=0, allow_inf_nan=False), Literal["inf"]]] = pd.Field(
```
chai-flex, line 1404: I thought our local tests are written using python client, so where does legacy come from?

chai-flex (approved):

PR #208, opened 2024-03-19
Fix issue when submitting JSON without lowMachPreconditioner

benflexcompute:


flow360/component/flow360_params/solvers.py
```diff
@@ -193,7 +193,7 @@ def to_solver(self, params, **kwargs) -> NavierStokesSolver:
         Set preconditioner threshold to freestream Mach number
         """
 
-        if self.low_mach_preconditioner:
+        if self.low_mach_preconditioner is not None and self.low_mach_preconditioner:
```
maciej-flexcompute, line 196: Can we just do if self.low_mach_preconditioner is True:

maciej-flexcompute (approved):

flow360/component/flow360_params/solvers.py
benflexcompute: This probably is not a good fix. Let me dig deeper.

benflexcompute:
PR by mistake

PR #217, opened 2024-03-21
Move over all the user-facing solver preprocessing

benflexcompute:
Preprocessing (and corresponding tests) added:

- [x] addDefaultTurbulenceQuantitiesToBoundaries
- [x] legacy-config[outputType]["outputFormat"] == "both"
- [x] populateBETMetricsOutput.process(config)
- [x] addDefaultMonitorType(config)
- [x] putViscosityRatioInFreestreamToTurbulenceQuantities(config)
- [x] addTipGap.process(config)
- [x] averageBETCoeffs.process(config)
- [x] UDDLegacy/if "outputLaw" in udd
- [x] checkDuplicateOutput
- [x] normalizeVectors
- [x] setDefaultEquationEvalFrequency
- [x] activateHeatEquationSolver
- [x] processUserExpressions (except UDE)

Already added in other PRs:
- [x] adjustReconstructionGradientLimiterForSA
- [x] addTurbulenceModelConstants
- [x] addTransitionNcrit
- [x] copyUniversalOutputFields
- [x] updateAdaptiveCFLParams.process(config)


flow360/component/utils.py
```diff
@@ -145,3 +147,111 @@ def zstd_compress(file_path, output_file_path=None, compression_level=3):
...
+    magnitude = 0
+    for comp in vector:
+        magnitude += comp * comp
+    magnitude = numpy.sqrt(magnitude)
+    if numpy.abs(magnitude - 1.0) < UNIT_VECTOR_NORM_TOLERANCE:
+        return (True, magnitude)
+    return (False, magnitude)
+
+
+def normalize_vector(vector, name: str):
```
maciej-flexcompute, line 168: I think the vector normalisation could be pydantic field validator:

```python
    # pylint: disable=no-self-argument, protected-access
    @pd.validator("axis")
    def normalise_axis(cls, value):
      is_unit_vector, vector_norm = check_unit_vector(value)
       normalized_vector = [0, 0, 0]
       if not is_unit_vector:
           if vector_norm == 0:
               raise ValueError(f"Zero vector found")
           for dim in range(0, 3):
               normalized_vector[dim] = vector[dim] / vector_norm
           return tuple(normalized_vector)
       return vector
```
then there is no need to pass "name" as parameter and no need to call it in to_solver() method. The advantage is that error will be raised on assignment instead of on submit when we call to_solver()

see here for more details: 
https://docs.pydantic.dev/1.10/usage/validators/#pre-and-per-item-validators
https://docs.pydantic.dev/1.10/usage/validators/#reuse-validators

flow360/component/utils.py
```diff
@@ -145,3 +147,111 @@ def zstd_compress(file_path, output_file_path=None, compression_level=3):
...
+    """
+    All in one funciton to precess expressions
+    """
+    expression = str(expression)
+    expression = add_trailing_semicolon(expression)
+    expression = remove_state_var_square_bracket(expression)
+    expression = convert_if_else(expression)
+    expression = convert_caret_to_power(expression)
+    expression = convert_legacy_names(expression)
+    return expression
```
maciej-flexcompute, line 257: this can also be called in validator instead of to_solver()

flow360/component/utils.py
benflexcompute: Great instruction. I changed according to the example

flow360/component/utils.py
benflexcompute: Done

flow360/component/flow360_params/flow360_output.py
```diff
@@ -477,10 +523,85 @@ def schema_extra(schema, model):
     def to_solver(self, params, **kwargs) -> MonitorOutput:
         solver_model = super().to_solver(params, **kwargs)
         solver_values = solver_model.__dict__
+        _deduplicate_output_fields(solver_values, "monitors")
         _distribute_shared_output_fields(solver_values, "monitors")
         return MonitorOutput(**solver_values)
 
 
+class NoTypeMonitor(MonitorBase):
```
maciej-flexcompute, line 531: What is means "NoTypeMonitor" ?

benflexcompute: This it the legacy way of specifying monitor back in the days when we only have probe monitor. So there is no "type" needed.

benflexcompute: Renamed to `LegacyMonitor`

flow360/component/utils.py
```diff
@@ -145,3 +147,128 @@ def zstd_compress(file_path, output_file_path=None, compression_level=3):
...
+    """
+    is_unit_vector, vector_norm = check_unit_vector(vector)
+    normalized_vector = [0, 0, 0]
+    if not is_unit_vector:
+        if vector_norm == 0:
+            raise ValueError("Zero vector found")
+        for dim in range(0, 3):
+            normalized_vector[dim] = vector[dim] / vector_norm
+        return tuple(normalized_vector)
+    return vector
```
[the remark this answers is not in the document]
benflexcompute: Moved normalization into vector

maciej-flexcompute (approved):

PR #225, opened 2024-03-28
add fields for volume mesher

feilin-flexcompute:
added:
num_layers
surface_boundaries

### DIFF
--- a/flow360/component/meshing/params.py
+++ b/flow360/component/meshing/params.py
@@ -7,6 +7,8 @@ from typing import List, Optional, Union, get_args
 import pydantic as pd
 from typing_extensions import Literal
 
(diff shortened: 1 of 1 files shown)


flow360/component/meshing/params.py
```diff
@@ -265,3 +264,11 @@ def flow360_json(self) -> str:
         """
 
         return self.json(encoder=flow360_json_encoder)
+
+class VolumeBeta(Volume):
```
chai-flex, line 268: Can we use the flags.internal_features or advanced_featurs over here? similar to the beta_features we had before?

maciej-flexcompute (approved):

PR #231, opened 2024-04-05
Expose and add defaults to turb const

benflexcompute:


flow360/__init__.py
```diff
@@ -92,6 +92,8 @@
     LinearSolver,
     NavierStokesSolver,
     PressureCorrectionSolver,
+    TurbulenceModelConstantsSA,
```
chai-flex, line 95: Can we use a better name here? SpalartAllmarasModelConstants? similarly for kOmegaSST we can use KOmegaSSTModelConstants

benflexcompute: Done and renamed to `KOmegaSSTModelConstants` and `SpalartAllmarasModelConstants`

PR #234, opened 2024-04-10
[SCFD-1408] Fix turbulence quantities update for freestream

awkrupka:


maciej-flexcompute (approved):

flow360/component/flow360_params/flow360_params.py
```diff
@@ -546,9 +546,6 @@ class FreestreamBase(Flow360BaseModel, metaclass=ABCMeta):
     turbulent_viscosity_ratio: Optional[NonNegativeFloat] = pd.Field(
         alias="turbulentViscosityRatio"
     )
-    ## Legacy update pending.
-    ## The validation for turbulenceQuantities (make sure we have correct combinations, maybe in root validator)
-    ## is also pending. TODO
```
benflexcompute, line 551: Can we keep these as the current PR does not address this?

What I meant by this TODO comment is that currently we do not have validation check in the python client for example using SA turbulence model while specifying `specificDissipationRate` under `turbulence_quantities` which is numerically invalid.

awkrupka: Oh, I see! I will revert this in this case. I thought this was a leftover from before we had pydantic models for the quantities.

awkrupka: Fixed. Merging now.

PR #235, opened 2024-04-10
[SCFD-1408] Fix turbulence quantities update for freestream

awkrupka:


awkrupka:
Cherrypicked changes from 2a36e53 on develop.

PR #239, opened 2024-04-11
added strucuture and interface for Simulation object

maciej-flexcompute:


benflexcompute:
Rebased on to the pydantic version bump PR.

PR #242, opened 2024-04-15
Bump Pydantic version to V2

benflexcompute:


flow360/__init__.py
```diff
@@ -144,3 +145,130 @@
 from .flags import Flags
 from .user_config import UserConfig
 from .version import __version__
+
```
benflexcompute, line 148: ruff complains import but not used for all these __init__.py files.

examples/volume_mesher_v2_from_inputs.py
```diff
@@ -1,9 +1,10 @@
-import os
-
-os.environ["FLOW360_BETA_FEATURES"] = "1"
+# Comment out to pass linter. Uncomment below to run this example
+# import os
+# os.environ["FLOW360_BETA_FEATURES"] = "1"
```
benflexcompute, line 3: I really cannot come up with a good way of fixing this without introducing a new error... Ruff complains that the imports are not at root level for this one.

flow360/component/utils.py
```diff
@@ -78,14 +77,6 @@ def shared_account_confirm_proceed():
         return True
 
 
-# pylint: disable=bare-except
```
benflexcompute, line 81: does not seem to be used.

flow360/log.py
```diff
@@ -299,14 +296,13 @@ def set_logging_file(
...
-        except Exception as error:  # pylint: disable=broad-exception-caught
+        except Exception as error:
             del log.handlers["file"]
             log.warning(f"Log file could not be closed: {error}")
 
     try:
-        # pylint: disable=consider-using-with,unspecified-encoding
         file = open(fname, filemode)
-    except:  # pylint: disable=bare-except
+    except (IOError, OSError):
```
benflexcompute, line 305: What else exceptions you guys can think of?

tests/params/test_outputs.py
```diff
@@ -36,6 +34,11 @@ def change_test_dir(request, monkeypatch):
     monkeypatch.chdir(request.fspath.dirname)
 
 
+@pytest.fixture
```
benflexcompute, line 37: This fixes the ruff complaint about `array_equality_override`/`mock_response` not used but if I pass the `setup_array_equality_override` to the test function then the unit test fails so this probably is just a hack instead of a fix.

Not sure what is a proper way as I thought passing the `setup_array_equality_override` to the test function should have worked...

benflexcompute: There are many similar instances.

tests/test_unit_system.py
```diff
@@ -26,8 +23,13 @@
...
 
 
+@pytest.fixture
+def setup_array_equality_override():
+    return array_equality_override
+
+
 class DataWithUnits(pd.BaseModel):
-    l: LengthType = pd.Field()
+    L: LengthType = pd.Field()
```
benflexcompute, line 32: Ruff does not like single l character as variable name as this is said to be easily confused with other letters (like 1 or i) when using certain fonts.

Make sense to me so I changed it. Is their any backend data structures that also uses this. I do not want to break consistency.

benflexcompute (commented):
Done self-reviewing

tests/params/test_outputs.py
andrzej-krupka: https://github.com/astral-sh/ruff/issues/4046

Seems like we are not the only ones having this issue.

The suggested solution is moving the fixture definition to conftest.py it seems.

flow360/__init__.py
andrzej-krupka: I think we should have an __all__ variable listing all imported types: https://github.com/astral-sh/ruff/issues/484

It seems like ruff should be able to fix those if ran with the --fix flag.

examples/volume_mesher_v2_from_inputs.py
andrzej-krupka: Beta_features are somewhat hacky right now and I think they aren't used anywhere - in the refactor we will probably need to come up with a better solution for feature flagging - I think this could be removed for now? @maciej-flexcompute

flow360/__init__.py
maciej-flexcompute: see here: https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/__init__.py

flow360/component/flow360_params/unit_system.py
```diff
@@ -445,47 +435,41 @@ def validate(vec_cls, value):
             setattr(cls_obj, "__get_validators__", lambda: (yield getattr(cls_obj, "validate")))
             return cls_obj
 
-    # pylint: disable=invalid-name
```
maciej-flexcompute, line 448: why ruff doesn't complain here?

flow360/component/flow360_params/unit_system.py
```diff
@@ -33,61 +32,61 @@
 u.dimensions.inverse_area = 1 / u.dimensions.area
 u.dimensions.inverse_length = 1 / u.dimensions.length
 
-# pylint: disable=no-member
+
```
maciej-flexcompute, line 35: why ruff doesn't complain here?

flow360/component/flow360_params/unit_system.py
```diff
@@ -1174,7 +1156,6 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 _flow360_system = {u.dimension_type.dim_name: u for u in dimensions}
 
 
-# pylint: disable=too-many-instance-attributes
```
maciej-flexcompute, line 1177: why ruff doesn't complain here, what is the ruff limit?

flow360/component/flow360_params/validations.py
```diff
@@ -104,7 +104,6 @@ def _check_consistency_ddes_volume_output(values):
     return values
 
 
-# pylint: disable=line-too-long
```
maciej-flexcompute, line 107: why ruff doesn't complain here? What is the line limit?

flow360/component/flow360_params/validations.py
```diff
@@ -365,8 +362,6 @@ def _check_bet_disks_number_of_defined_polars(disk):
     return disk
 
 
-# pylint: disable=invalid-name
-# pylint: disable=too-many-arguments
```
maciej-flexcompute, line 369: why ruff doesn't complain here? What is the limit of arguments?

flow360/component/flow360_params/volume_zones.py
```diff
@@ -124,14 +110,12 @@ class ReferenceFrameExpression(ReferenceFrameBase):
     theta_radians: Optional[str] = pd.Field(alias="thetaRadians")
     theta_degrees: Optional[str] = pd.Field(alias="thetaDegrees")
 
-    # pylint: disable=missing-class-docstring,too-few-public-methods
```
maciej-flexcompute, line 127: why ruff doesn't complain here? Is class docstring not required?

tests/test_unit_system.py
maciej-flexcompute: this is ok

flow360/__init__.py
benflexcompute: I don't know if ruff auto fix them once we put __all__ in __init__.py. I manual fixed these.

examples/volume_mesher_v2_from_inputs.py
benflexcompute: I think @feilin-flexcompute added this BETA tag for new volume mesher pipeline so it is used for that only.

tests/params/test_outputs.py
benflexcompute: Changed to use pytest_plugins in conftest.py

benflexcompute:
Copied Tidy3D's ruff settings here.

benflexcompute:
Forked to https://github.com/flexcompute/Flow360/pull/247 and pylint is prefered.

PR #244, opened 2024-04-16
Change the submission of solver JSON to user JSON

benflexcompute:
Solver side PR: https://github.com/flexcompute/Flow360Core/pull/1197
testcases PR: https://github.com/flexcompute/testcases/pull/543

### DIFF
--- a/flow360/component/case.py
+++ b/flow360/component/case.py
@@ -312,7 +312,7 @@ class CaseDraft(CaseBase, ResourceDraft):
         data = {
             "name": self.name,
             "meshId": volume_mesh_id,
-            "runtimeParams": self.params.flow360_json(),
+            "runtimeParams": self.params.json(),
             "tags": self.tags,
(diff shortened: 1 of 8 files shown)


flow360/component/case.py
```diff
@@ -312,7 +312,7 @@ def submit(self, force_submit: bool = False) -> Case:
         data = {
             "name": self.name,
             "meshId": volume_mesh_id,
-            "runtimeParams": self.params.flow360_json(),
+            "runtimeParams": self.params.json(),
```
maciej-flexcompute, line 315: I think we need to tie this python (develop branch) to one version of solver so that we can still use develop branch to submit cases. Is solver ready?

benflexcompute: Good question. I have a PR for Core listed above. I am still working on testcases PR which changed quite a bit because of this.

maciej-flexcompute (approved):

PR #245, opened 2024-04-18
bumped pydantic version to 1.10.8 to remove copy() bug

maciej-flexcompute:


benflexcompute:
Can you add any unit test associated with the issue?

benflexcompute (approved):

PR #246, opened 2024-04-19
Ben/fixing issues found in solver core unit test

benflexcompute:
- [x] Added `ModifiedRestartSolution` for explicit restart manipulation [INTERNAL]
- [x] Decoupled `LinearSolverLegacy` from `LinearSolver`
- [x] Fix bug when LinearSolver is present but LinearSolverConfig is not. [Missing Legacy Pattern Found]
- [x] Deprecated and move `FreestreamInitialCondition` to legacy
- [x] Added surfaces to `surfaceOutputLegacy` [Missing Legacy Pattern Found]

### DIFF
--- a/flow360/component/flow360_params/solvers.py
+++ b/flow360/component/flow360_params/solvers.py
@@ -12,7 +12,13 @@ import pydantic as pd
 from typing_extensions import Literal
 
 from ..types import NonNegativeFloat, NonNegativeInt, PositiveFloat, PositiveInt
-from .flow360_legacy import LegacyModel, try_set, try_update
+from .flow360_legacy import (
+    LegacyModel,
+    LinearSolverLegacy,
+    set_linear_solver_config_if_none,
+    try_set,
+    try_update,
+)
 from .params_base import Conflicts, DeprecatedAlias, Flow360BaseModel
 from .time_stepping import UnsteadyTimeStepping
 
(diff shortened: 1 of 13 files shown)


flow360/component/flow360_params/solvers.py
```diff
@@ -561,11 +550,13 @@ class NavierStokesSolverLegacy(NavierStokesSolver, LegacyModel):
...
-    linear_solver_config: Optional[LinearSolverLegacy] = pd.Field(
-        alias="linearSolverConfig", default=LinearSolverLegacy()
-    )
+    linear_solver_config: Optional[LinearSolverLegacy] = pd.Field(alias="linearSolverConfig")
     linear_iterations: Optional[PositiveInt] = pd.Field(alias="linearIterations")
 
+    _processed_linear_solver_config = pd.validator(
+        "linear_solver_config", allow_reuse=True, pre=True
+    )(set_linear_solver_config_if_none)
+
```
maciej-flexcompute, line 559: if we just want to set defaults, why we don't do:
```python
linear_solver_config: Optional[LinearSolverLegacy] = pd.Field(default=LinearSolverLegacy(), alias="linearSolverConfig")
```
?

tests/data/cases/case_7.json
```diff
@@ -33,18 +33,25 @@
...
-      "Cp": true,
-      "Cf": true,
-      "CfTangent": true,
-      "CfNormal": true,
-      "CfVec": true,
-      "yPlus": true,
-      "wallDistance": true,
-      "Mach": true,
-      "nodeForcesPerUnitArea": true,
-      "animationFrequency": -1
```
maciej-flexcompute, line 47: is this covered by other test?

flow360/component/flow360_params/solvers.py
benflexcompute: Because defaults are applied before Pre=True validator.

Once the default LinearSolverLegacy() gets applied then the new validator will never take any action as the `linearSolverConfig` is no longer None.

tests/data/cases/case_7.json
benflexcompute: Yes, this boolean legacy pattern is seen in many json files. E.g. case_20.json.
Added unit test for this nonetheless.

maciej-flexcompute (approved):

PR #249, opened 2024-05-01
Configuration setting to use system SSL certificates

tim-goddard-flexcompute:
This is an in-progress patch to support using the OS certificate store (e.g. Windows certificate store) for Flow360 requests, when given a configuration option to change this behavior. Not expected to alter existing default behavior. Further testing is required before taking the "draft" mark off this.

We have a customer who is experiencing SSL-related errors, likely related to an SSL intercepting proxy. The certificates for these are usually distributed to the OS certificate store, but will not be copied to the certipy default bundle. This may resolve such issues, by allowing the OS store which the organisation is managing to be used.

Any advice on testing approach would be welcome. The current unit tests don't make any real HTTP requests, which would be needed to test this functionality. Making those requests would require Internet connectivity where tests are run. In the draft tests I ran, there were issues connecting to the Flow360 API (even public endpoints), because the client is not designed to connect without an access key.

### DIFF
--- a/flow360/user_config.py
+++ b/flow360/user_config.py
@@ -123,6 +123,11 @@ class BasicUserConfig:
     def enable_validation(self):
         """enable user side validation (pydantic)"""
         self._do_validation = True
-
+    
+    @property
+    def use_system_certs(self) -> bool:
+        map = self.config.get(self.profile, {})
+        setting = map.get("usesystemcerts", False)
+        return setting
 
 UserConfig = BasicUserConfig()
--- a/flow360/user_config.py
+++ b/flow360/user_config.py
@@ -23,6 +23,7 @@ class BasicUserConfig:
         self.set_profile(DEFAULT_PROFILE)
         self._check_env_profile()
         self._apikey = None
+        self._use_system_certs = None
         self._check_env_apikey()
         self._do_validation = True
         self._suppress_submit_warning = None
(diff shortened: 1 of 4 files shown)


tim-goddard-flexcompute:
Had a look at that failing matrix test, but it's complaining about a line not altered in this PR, last change dated to 6 months ago. No idea why that would pop out now.

flow360/user_config.py
```diff
@@ -124,5 +131,22 @@ def enable_validation(self):
...
+        -------
+        bool
+            whether to use the OS system certificate store for SSL validation"""
+        self._check_env_profile()
+        self._check_env_usessytemcerts()
+        if self._use_system_certs is not None:
+            return self._use_system_certs
+
+        config_map = self.config.get(self.profile, {})
+        setting = config_map.get("usesystemcerts", False)
```
maciej-flexcompute, line 148: this one to work requires usesystemcerts being available in config.toml file which is managed by flow360 CLI. You can add this option to the CLI so the user can set it only once: https://github.com/flexcompute/Flow360/blob/develop/flow360/cli/app.py

tests/_test_http.py
maciej-flexcompute, line 1: the `_test_...` are not run automatically in the pipeline. This is why I think this test will have no effect

maciej-flexcompute:
@dominik-flex can you look at this PR? Do we need this or can we close it?

dominik-flex:
@maciej-flexcompute I think we no longer need this.

The solution for requests made with the Python `requests` library is to prepare a custom CA bundle (.pem file) with the Zscaler root certificate in it and set the `REQUESTS_CA_BUNDLE` environment variable to point to that file. It worked for the customer in question.

PR #253, opened 2024-05-08
[Tracking] Simulation Interface

benflexcompute:


examples/simulation_examples/rotation.py
```diff
@@ -0,0 +1,105 @@
...
+                reference_geometry=ReferenceGeometry(
+                    area=1,
+                    moment_length=2,
+                    mesh_unit=3 * u.m,
+                ),
+            ),
+            Rotation(
+                entities=[rotation_zone_inner],
+                angular_velocity=5 * u.rpm,
+            ),
```
yifan-flex, line 87: Rotation shouldn't be a type of volume right? It's more like a property of a volume. Currently in json it's defined under FluidDynamics type of volume zones. Why didn't we go with that design?

benflexcompute: We renamed `volume` to `models` which suits the Rotation concept better. `volumes` created a lot of confusion.

"FluidDynamics" is not a good name for rotating zone. BETDisk is also "FluidDynamics" same for Actuator disk and so on.

Also I think nesting `rotation` inside `FludiDynamics` is not good design, it creates unnecessary burden to write FludiDynamics every time when the Rotation itself already implies that.

yifan-flex: I see. This makes sense. What's the consideration of putting navier_stokes_solver and turbulent_model_solver definitions under FluidDynamics? It's possible that the user does not define a FluidDynamics model but uses only Rotation in the case of SRF, or only PorousMedia in the case of a single porous zone. Should these classes also have solver definitions as their attributes?
It doesn't make sense to me to enable the possibility of individual navier stokes solver configurations for each FluidDynamics or Rotation model though. These zones are strongly coupled. I don't think It's likely that in the future we will need such functionality either

benflexcompute: I think by default we will have one `FluidDynamics` applied to all volumes if the user did not specify one. Or we throw error.

We had a previous design with NS and turb solver in all models and got comment from application team that each model should focus on attributes that are specific to that model.

I agree with you that we do not see the need for changing NS config in each zone at this moment but the main goal of this design is to have an interface that is _as future-proof as possible_. It will be very painful to change interface and consider compatibility issues when we realize some time in the future that some new parameter in NS/SA/SST etc DO need to change with zones.

yifan-flex: I think even if we need to have solver configuration change with zones in the future, the interface should be solver config owning zones, not the other way around. Let's discuss tomorrow

flow360/component/simulation/material/material.py
```diff
@@ -0,0 +1,24 @@
...
+
+class Material(Flow360BaseModel):
+    # contains models of getting properites, for example US standard atmosphere model
+    name: str = pd.Field()
+    dynamic_viscosity: float = pd.Field()
+
+
+class Air(Material):
+    name: Literal["air"] = pd.Field(frozen=True)
+    dynamic_viscosity: float = pd.Field(18.03e-6, frozen=True)
```
chai-flex, line 24: This has no unit?

flow360/component/simulation/mesh.py
```diff
@@ -0,0 +1,47 @@
+from typing import List, Optional, Union
+
+import pydantic as pd
+
+from flow360.component.simulation.framework.base_model import Flow360BaseModel
+
+from .zones import BoxZone, CylindricalZone
+
+
+class FaceRefinement(Flow360BaseModel):
```
chai-flex, line 10: Should we names consistent with what are used in the simulation interface?

SurfaceRefinement? VolumeRefinement?

flow360/component/simulation/meshing_param/edge_params.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class ByHeight(Flow360BaseModel):
+    """Surface edge refinement by specifying first layer height of the anisotropic layers"""
+
+    type: Literal["height"] = pd.Field("height", frozen=True)
+    value: LengthType.PositiveFloat = pd.Field()
+
+
+class ByAspectRatio(Flow360BaseModel):
```
chai-flex, line 24: This should be renamed as AspectRatioBasedRefinement.

ByAspectRatio is not good english and neither is a good class name. similar for the others.

If the above is too long, just AspectRatio is better

flow360/component/simulation/meshing_param/params.py
```diff
@@ -0,0 +1,57 @@
...
+    - `class` CylindricalRefinement
+
+    Affects surface meshing:
+    - surface_layer_growth_rate
+    - `class` FaceRefinement
+    - `class` SurfaceEdgeRefinement
+    """
+
+    # Volume **defaults**:
+    farfield: Literal["auto", "quasi-3d", "user-defined"] = pd.Field(
```
chai-flex, line 37: we should probably rename this as the concept of farfield is external aero specific. it is probably too late though

flow360/component/simulation/meshing_param/params.py
```diff
@@ -0,0 +1,57 @@
...
+        description="Type of farfield generation."
+    )
+    refinement_factor: pd.PositiveFloat = pd.Field(
```
chai-flex, line 40: rename to global_refinement_factor? user might get confused with another keyword called refinements. also why do we need this to be seprate? can it be unified with refinments

flow360/component/simulation/meshing_param/params.py
```diff
@@ -0,0 +1,57 @@
...
+        ge=0,
+        le=1,
+        description="Narrow gap treatment strength used when two surfaces are in close proximity. Use a value between 0 and 1, where 0 is no treatment and 1 is the most conservative treatment. This parameter has a global impact where the anisotropic transition into the isotropic mesh. However, the impact on regions without close proximity is negligible.",
+    )
+
+    surface_layer_growth_rate: float = pd.Field(
+        ge=1, description="Global growth rate of the anisotropic layers grown from the edges."
+    )
+
+    refinements: Optional[
```
chai-flex, line 53: I don't see the faces attribute over from the volume mesher json? is there a reason for it to not be there?

flow360/component/simulation/material/material.py
benflexcompute: @yifan-flex Please send fix

benflexcompute: This is an unused file

benflexcompute: deleted

flow360/component/simulation/mesh.py
benflexcompute: We do have this Face concept right now which is different from Surface.

`Surface` is defined as the group of patches (which is the smallest unit) that share the same **boundary condition**. Face is defined as the group of patches (which is the smallest unit) that share the same **meshing parameters**

chai-flex: I don't think there should be such a distinction. A group of patches should be one entity. We can either use it for meshing or solver or both. Why are you making the user confused that it is called different things in different places

Similar for ZoneRefinement. Why not just call it VolumeRefinement

benflexcompute: This is not something coming from Simulation design. The Face-Surface distinction comes from our mesher. I also in favor of unification. CC: @rundaji

flow360/component/simulation/meshing_param/params.py
benflexcompute: I think Face is a concept that only applies to the surface meshing.

flow360/component/simulation/mesh.py
rundaji: Just FYI, the patches (NURBS) can be categized in two different ways
(1) `faceName` v.s. (2) `groupName`/`boundaryName`

(1) `faceName` is closely related to meshing:
As shown in figure 3.3.2, the patches with same `faceName` will share the same max_edge_length
https://docs.flexcompute.com/projects/flow360/en/latest/preprocessing/automatedMeshing/automatedMeshing.html#surface-meshing
As shown in figure 7.7.2, the patchces with same `faceName` will share the same
https://docs.flexcompute.com/projects/flow360/en/latest/tutorials/AutoMeshingInternalFlow/AutoMeshingInternalFlow.html#volume-mesh

(2) `groupName`/`boundaryName` is more related to boundary conditions when dumping the ugrid/cgns files

rundaji: @benflexcompute Could you direct me to where `Surface` is define for group of patches for boundary conditions?

flow360/component/simulation/meshing_param/volume_params.py
```diff
@@ -0,0 +1,42 @@
...
+Meshing settings that applies to volumes.
+"""
+
+
+class UniformRefinement(Flow360BaseModel):
+    entities: EntityList[Box, Cylinder] = pd.Field()
+    spacing: LengthType.Positive = pd.Field()
+
+
+class AxisymmetricRefinement(Flow360BaseModel):
```
rundaji, line 20: (1) Refinement
(1.1) Box
(1.2) Cylinder
(a) By default, we only require minimal input, a cylinder with uniform refinement zone will be generated
(b) If the user prescribe additional parameters, then an anisotropic cylindrical refinement zone, could be an annular cylinder, will be generated
(2) Interface
Interface will be an alias of refinement (Chai's suggestion)
(2.1) Cylinder
This will be an alias of Refinement -> Cylinder, with enclosed objects
(2.2) Sphere
This will be an alias of Refinement -> Sphere, with enclosed objects
P.S. Currently we don't support spherical refinement/interface right now, will support this later on (in-house mesher)

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -0,0 +1,664 @@
...
+    # pylint: disable=no-member
+    mesh_unit: LengthType.Positive,
+):
+    """
+    Get the solver json from the simulation parameters.
+    """
+
+    translated = {}
+    ##:: Step 1: Get geometry:
+    geometry = remove_units_in_dict(dump_dict(input_params.reference_geometry))
```
maciej-flexcompute, line 520: @benflexcompute 
reference_geometry is optional and default is None, it will break here

flow360/component/simulation/translator/utils.py
```diff
@@ -0,0 +1,339 @@
...
+            return new_dict
+        for key, value in input_dict.items():
+            if isinstance(value, dict) and value.keys() == unit_keys:
+                new_dict[key] = value["value"]
+            else:
+                new_dict[key] = remove_units_in_dict(value)
+        return new_dict
+    if isinstance(input_dict, list):
+        return [remove_units_in_dict(item) for item in input_dict]
+    return input_dict
```
maciej-flexcompute, line 133: the old design (Flow360Params) has a check that every unit that is removed from dict is "flow360" unit. Here we just remove units which is dangerous.

yifan-flex: @benflexcompute

benflexcompute: @yifan-flex can you send a PR to make sure this is the case?

yifan-flex (approved):

.github/workflows/test.yml
```diff
@@ -3,9 +3,9 @@ name: unit testing
 on:
   workflow_dispatch:
   push:
-    branches: [ develop, main, release-candidate/* ]
+    branches: [ develop, main, release-candidate/*, UnderConstruction/SimulationInterface ]
```
benflexcompute, line 6: Will remove upon merge

.github/workflows/codestyle.yml
```diff
@@ -56,4 +58,4 @@ jobs:
       - name: Install dependencies
         run: poetry install
       - name: Run pylint
-        run: poetry run pylint flow360 --rcfile .pylintrc
+        run: poetry run pylint $(git ls-files "flow360/*.py") --rcfile .pylintrc
```
yifan-flex, line 61: We should update the pylint command in README

benflexcompute: Done

flow360/component/simulation/user_story_examples/debug_divergence_by_change_mesh_para.py
maciej-flexcompute, line 1: I think we need to delete entire `user_story_examples`

flow360/component/validator.py
```diff
@@ -26,6 +27,8 @@ def _get_url(self):
             return "validator/surfacemesh/validate"
         if self is Validator.CASE:
             return "validator/case/validate"
+        if self is Validator.GEOMETRY:
+            return "validator/geometry/validate"
```
maciej-flexcompute, line 31: do we have geometry validator? I think this should be removed

flow360/examples/base_test_case.py
```diff
@@ -111,6 +111,7 @@ def _get_version_prefix(cls):
 
     @classmethod
     def _real_path(cls, *args):
+        here = os.path.dirname(os.path.abspath(__file__))
```
maciej-flexcompute, line 114: `here` is already defined in line 15

tests/pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+filterwarnings =
+    error
```
maciej-flexcompute, line 3: what is this for?

tools/integrations/data_v2/UDD/UserDefinedDynamic-release-24.6.json
maciej-flexcompute, line 1: I think we should remove all data under /data_v2 folder and let python script to generate them when needed

flow360/component/flow360_params/updater.py
```diff
@@ -18,6 +18,7 @@ def _no_update(params_as_dict):
     ("23.3.0", "23.3.*", _no_update),
     ("23.3.*", "24.2.*", _no_update),
     ("24.2.*", "24.2.*", _no_update),
+    ("24.2.*", "24.3.*", _no_update),  # we should not allow to submit Flow360Params to version 24.3
```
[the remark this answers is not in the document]
benflexcompute: Right. Let's keep this as it for now. If 24.7 cannot wait for the simulation then we also add it here then.

flow360/component/flow360_params/params_base.py
```diff
@@ -21,12 +21,16 @@
...
+)
+
 from ...error_messages import do_not_modify_file_manually_msg
 from ...exceptions import Flow360FileError, Flow360ValidationError
 from ...log import log
 from ..types import COMMENTS, TYPE_TAG_STR
 from .conversions import need_conversion, require, unit_converter
-from .unit_system import DimensionedType, is_flow360_unit
 
 SUPPORTED_SOLVER_VERSION = "release-23.3.2.0"
```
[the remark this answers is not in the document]
benflexcompute: Removed

flow360/component/flow360_params/conversions.py
```diff
@@ -146,6 +151,10 @@ def get_base_length():
         base_length = params.geometry.mesh_unit.to("m").v.item()
         return base_length
 
+    def get_base_angle():
+        # pylint: disable=no-member
+        return 1 * u.rad
```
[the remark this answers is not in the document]
benflexcompute: I do not think we used to have any angle type in Flow360Params

flow360/component/flow360_params/params_base.py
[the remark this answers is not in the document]
benflexcompute: removed

flow360/component/geometry_v1.py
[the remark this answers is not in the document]
benflexcompute: It is not used anywhere except examples. Let's remove this.

flow360/component/simulation/cloud.py
```diff
@@ -0,0 +1,114 @@
...
+def generate_volume_mesh(
+    source_asset: AssetBase, params: SimulationParams, async_mode: bool = True
+):
+    """generate volume mesh from the geometry"""
+    return _run(source_asset, params, VolumeMesh, async_mode)
+
+
+def run_case(source_asset: AssetBase, params: SimulationParams, async_mode: bool = True):
+    """run case from the geometry"""
+    return _run(source_asset, params, Case, async_mode)
```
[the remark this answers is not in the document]
benflexcompute: I think `run_case` is fine. It is not a very long name.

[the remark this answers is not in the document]
benflexcompute: Now the case name appears to be Case_${draft name}. We need back end change for a complete support for user specified case name.

flow360/component/simulation/conversion.py
```diff
@@ -0,0 +1,331 @@
...
+
+    elif dimension == u.dimensions.frequency:
+        base_time = get_base_time()
+
+        flow360_conversion_unit_system.base_frequency = base_time ** (-1)
+
+    elif dimension == u.dimensions.angle:
+
+        # pylint: disable=no-member
+        flow360_conversion_unit_system.base_angle = 1 * u.rad
```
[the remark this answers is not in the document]
benflexcompute: With some testing it seems that only the value part (1) makes a difference.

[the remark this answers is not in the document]
benflexcompute: I removed the u.rad part.

flow360/component/simulation/framework/base_model.py
```diff
@@ -0,0 +1,651 @@
...
+    unit_converter,
+)
+from flow360.component.simulation.unit_system import LengthType
+from flow360.component.types import COMMENTS, TYPE_TAG_STR
+from flow360.error_messages import do_not_modify_file_manually_msg
+from flow360.exceptions import Flow360FileError
+from flow360.log import log
+
+
+def custom_to_camel(string: str) -> str:
```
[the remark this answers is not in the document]
benflexcompute: custom to signify that this has a custom rule of capitalization. Renamed to snake_to_camel

flow360/component/simulation/framework/entity_registry.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        index = 0
+        print("---- Content of the registry ----")
+        # pylint: disable=no-member
+        for entity_bucket, entities in self.internal_registry.items():
+            print(f"    Entities of type '{entity_bucket}':")
+            for entity in entities:
+                print(f"    - [{index:03d}]\n{entity}")
+                index += 1
+        print("---- End of content ----")
```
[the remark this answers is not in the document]
benflexcompute: Done. Changed to __str__

flow360/component/simulation/models/solver_numerics.py
```diff
@@ -0,0 +1,363 @@
...
+
+    Example
+    -------
+    >>> ns = NavierStokesSolver(absolute_tolerance=1e-10)
+    """
+
+    absolute_tolerance: PositiveFloat = pd.Field(1.0e-10)
+
+    CFL_multiplier: PositiveFloat = pd.Field(1.0)
+    kappa_MUSCL: pd.confloat(ge=-1, le=1) = pd.Field(-1)
```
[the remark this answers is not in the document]
benflexcompute: Let's collect feedback on renaming stuff during internal testing and do all at once later. Otherwise it is a front end nightmare if we constantly changes things here and there.

flow360/component/simulation/operating_condition.py
```diff
@@ -0,0 +1,238 @@
...
+        temperature_offset: TemperatureType = 0 * u.K,
+    ):
+        """Constructs a thermal state from the standard atmosphere model."""
+        # pylint: disable=fixme
+        # TODO: add standard atmosphere implementation
+        density = 1.225 * u.kg / u.m**3
+        temperature = 288.15 * u.K
+
+        state = cls(density=density, temperature=temperature, material=Air())
+
```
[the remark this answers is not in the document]
benflexcompute: Implemented with unit test

flow360/component/simulation/operating_condition.py
```diff
@@ -0,0 +1,238 @@
...
+    def dynamic_viscosity(self) -> ViscosityType.Positive:
+        """Computes dynamic viscosity."""
+        return self.material.get_dynamic_viscosity(self.temperature)
+
+    # TODO: should we make this private_attribute?
+    @pd.validate_call
+    def mu_ref(self, mesh_unit: LengthType.Positive) -> pd.PositiveFloat:
+        """Computes nondimensional dynamic viscosity."""
+        # TODO: use unit system for nondimensionalization
+        return (self.dynamic_viscosity / (self.speed_of_sound * self.density * mesh_unit)).v.item()
```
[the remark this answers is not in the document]
benflexcompute: Ok. Made it a private function for now.

flow360/component/simulation/outputs/output_entities.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __eq__(self, other):
+        if isinstance(other, _OutputItemBase):
+            return (self.name + "-" + self.__class__.__name__) == (
+                other.name + "-" + other.__class__.__name__
+            )
+        return False
+
+    def __str__(self):
+        return f"{self.__class__.__name__} with name: {self.name}"
```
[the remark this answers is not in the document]
benflexcompute: This is when there is error. The error message will have more readable message.

tests/pytest.ini
benflexcompute: To force developers handling warnings.

flow360/component/simulation/outputs/outputs.py
```diff
@@ -0,0 +1,192 @@
...
+    output_type: Literal["SurfaceIntegralOutput"] = pd.Field("SurfaceIntegralOutput", frozen=True)
+
+
+class ProbeOutput(_AnimationSettings):
+    """Probe monitor output settings."""
+
+    name: Optional[str] = pd.Field(None)
+    entities: Optional[EntityList[ProbeGroup]] = pd.Field(None, alias="probe_groups")
+    output_fields: UniqueAliasedStringList[CommonFieldNames] = pd.Field()
+    output_type: Literal["ProbeOutput"] = pd.Field("ProbeOutput", frozen=True)
```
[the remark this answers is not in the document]
benflexcompute: Multiple ProbeOutput will be inconsistent with other output types.

For example we do not force user to specify one SliceOutput for each slice right? Each slice gets dumped into one vtu/plt file and same for each probeGroup (one csv file).

The benefit is to share the same output fields across multiple probeGroup.

I created a ticket for your second point. We actually have multiple places which needs the same treatment.

flow360/component/simulation/outputs/outputs.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+class AeroAcousticOutput(Flow360BaseModel):
+    """AeroAcoustic output settings."""
+
+    name: Optional[str] = pd.Field(None)
+    patch_type: Literal["solid"] = pd.Field("solid", frozen=True)
+    # pylint: disable=no-member
+    observers: List[LengthType.Point] = pd.Field()
```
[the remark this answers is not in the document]
benflexcompute: Right. We can change this. And yes we plan to have import from CSV a class method of ProbeGroup.

flow360/component/simulation/services.py
```diff
@@ -0,0 +1,270 @@
...
+        raise ValueError(str(err)) from err
+    except Exception as err:  # tranlsation itself is not supposed to raise any other exception
+        raise ValueError(
+            f"Unexpected error translating to {target_name} json: " + str(err)
+        ) from err
+
+    if translated_dict == {}:
+        raise ValueError(f"No {target_name} parameters found in given SimulationParams.")
+    # pylint: disable=fixme
+    # TODO: Implement proper hashing. Currently floating point creates headache for reproducible hashing.
```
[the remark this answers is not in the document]
benflexcompute: I think not. I remember tried on the webUI and it already can skip the surface meshing when I already generated it before. I removed this todo.

flow360/component/simulation/simulation_params.py
```diff
@@ -0,0 +1,281 @@
...
+from flow360.error_messages import (
+    unit_system_inconsistent_msg,
+    use_unit_system_for_simulation_msg,
+)
+from flow360.exceptions import Flow360ConfigurationError, Flow360RuntimeError
+from flow360.version import __version__
+
+AllowedModelTypes = Annotated[
+    Union[VolumeModelTypes, SurfaceModelTypes], pd.Field(discriminator="type")
+]
```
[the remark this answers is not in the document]
benflexcompute: renamed to ModelTypes. Let's again do all discriminator change in a dedicated PR.

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -0,0 +1,866 @@
...
+    """
+    for freq_key in ["animationFrequencyTimeAverage", "animationFrequency"]:
+        if freq_key not in output_dict:
+            output_dict[freq_key] = -1
+    for offset_key in ["animationFrequencyTimeAverageOffset", "animationFrequencyOffset"]:
+        if offset_key not in output_dict:
+            output_dict[offset_key] = 0
+    if "startAverageIntegrationStep" not in output_dict:
+        output_dict["startAverageIntegrationStep"] = -1
+    return output_dict
```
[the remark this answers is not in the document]
benflexcompute: yes, we are still in testing and validation stage.

flow360/component/simulation/translator/utils.py
benflexcompute: It is now addressed

flow360/component/validator.py
benflexcompute: Removed

flow360/examples/base_test_case.py
benflexcompute: removed

tools/integrations/data_v2/UDD/UserDefinedDynamic-release-24.6.json
benflexcompute: Removed

flow360/component/simulation/operating_condition.py
```diff
@@ -72,14 +72,34 @@ def from_standard_atmosphere(
...
+        temperature += temperature_offset
 
+        # Calculate pressure at the given altitude
+        pressure = p0 * (1 - (lapse_rate * altitude / t0)) ** (g0 / (r * lapse_rate))
+
+        # Calculate density at the given altitude
+        density = pressure / (r * temperature)
+
+        # Construct and return the thermal state
+        state = cls(density=density, temperature=temperature, material=Air())
```
maciej-flexcompute: not sure where this model comes from, but I think implementation of US standard atmosphere model is more complex, here is one implementation: https://github.com/flexcompute/design360/blob/develop/design360/client/utils/atmosphere_model.py

maciej-flexcompute (approved):

flow360/component/simulation/operating_condition.py
benflexcompute: I copied your implementation

PR #259, opened 2024-05-15
add version to surface meshing params

feilin-flexcompute:


examples/surface_mesh_airplane_v2_from_inputs.py
```diff
@@ -0,0 +1,29 @@
+import flow360 as fl
+from flow360.examples import Airplane
+
+os.environ["FLOW360_BETA_FEATURES"] = "1"
+
+params = fl.SurfaceMeshingParams(
+    version="v2",
+    max_edge_length=0.16,
+    edges={
```
chai-flex, line 9: I don't think the v2 mesher supports edges or faces, let's just keep max_edge_length

feilin-flexcompute: fixed

chai-flex (approved):

PR #267, opened 2024-05-20
integrate geometry, surface and volume mesher

feilin-flexcompute:
- [x] add version for surface mesher
- [x] add/modify key-value in meshers for DB
- [x] add examples for integration tests from geometry to simulation


examples/simulation_examples/airplane_case.py
```diff
@@ -0,0 +1,57 @@
+import flow360 as fl
+from flow360 import units as u
+
+
+def createBaseParams_airplane():
```
maciej-flexcompute, line 5: `create_base_params_airplane`

also, why do we want this to be in the examples? We already have example on how to create Flow360Params. We can use this as second example. But other examples should not refer to this file. If they need to load example config, it needs to be loaded from file. Every example should be standalone runnable script

examples/simulation_examples/cylinder_case.py
maciej-flexcompute, line 1: same as ariplane_case.py comment

flow360/component/surface_mesh.py
```diff
@@ -35,13 +38,46 @@ class SurfaceMeshDownloadable(Enum):
     CONFIG_JSON = "config.json"
 
 
+class SurfaceMeshFileFormat(Enum):
+    """
+    Surface mesh file format
+    """
+
+    UGRID = "lb8.ugrid"
+    STL = "stl"
```
maciej-flexcompute, line 47: a test is needed for this class

examples/simulation_examples/airplane_case.py
feilin-flexcompute: fixed.

examples/simulation_examples/cylinder_case.py
feilin-flexcompute: fixed

maciej-flexcompute (approved):

PR #270, opened 2024-05-21
[ModelRef] Added Field Definitions for **Outputs**

benflexcompute:
Depends on https://github.com/flexcompute/Flow360/pull/258


flow360/component/simulation/outputs/output_entities.py
```diff
@@ -0,0 +1,31 @@
...
+from flow360.component.simulation.base_model import Flow360BaseModel
+
+
+class Slice(EntityBase):
+    _entity_type: Literal["SliceType"] = "SliceType"
+    slice_normal: Tuple[float, float, float] = pd.Field()
+    slice_origin: Tuple[float, float, float] = pd.Field()
+
+
+class Isosurface(EntityBase):
```
benflexcompute, line 14: IsoSurface or Isosurface?????

flow360/component/simulation/outputs/output_entities.py
```diff
@@ -0,0 +1,35 @@
...
+    _entity_type: Literal["SurfaceIntegralMonitorType"] = "SurfaceIntegralMonitorType"
+    type: Literal["surfaceIntegral"] = pd.Field("surfaceIntegral", frozen=True)
+    entities: EntityList[Surface] = pd.Field(alias="surfaces")
+
+
+@final
+class ProbeMonitor(EntityBase):
+    _entity_type: Literal["ProbeMonitorType"] = "ProbeMonitorType"
+    type: Literal["probe"] = pd.Field("probe", frozen=True)
+    locations: List[Tuple[float, float, float]] = pd.Field()
```
maciej-flexcompute, line 35: I think according to our designs, entity is a singular object and you can provide multiple entities to one entity list

flow360/component/simulation/outputs/outputs.py
```diff
@@ -0,0 +1,126 @@
...
+    Slice,
+    SurfaceIntegralMonitor,
+)
+from flow360.component.simulation.primitives import Surface
+from flow360.component.simulation.types import NonNegativeAndNegOneInt
+
+"""Mostly the same as Flow360Param counterparts.
+Caveats:
+1. Check if we support non-average and average output specified at the same time in solver. (Yes but they share the same output_fields)
+2. We do not support mulitple output frequencies for the same type of output.
```
maciej-flexcompute, line 19: how do we enforce this? By validators or by data level design?

flow360/component/simulation/outputs/outputs.py
```diff
@@ -0,0 +1,126 @@
...
+"""
+
+# TODO: We need a proper list
+SurfaceOutputFields = List[str]
+VolumeOutputFields = List[str]
+SliceOutputFields = List[str]
+MonitorOutputFields = List[str]
```
maciej-flexcompute, line 26: can we use this: https://github.com/flexcompute/Flow360/blob/develop/flow360/component/flow360_params/flow360_fields.py
I think this good, pydantic agnostic design

flow360/component/simulation/outputs/outputs.py
```diff
@@ -0,0 +1,126 @@
...
+class _TimeAverageAdditionalAnimationSettings(Flow360BaseModel):
+    """
+    Additional controls when using time-averaged output.
+
+    Notes
+    -----
+        Old `computeTimeAverages` can be infered when user is explicitly using for example `TimeAverageSurfaceOutput`.
+    """
+
+    start_average_integration_step: Optional[NonNegativeAndNegOneInt] = pd.Field(
```
maciej-flexcompute, line 54: why is this called "start_average_integration_step"? Maybe we can call it just `start_step`

flow360/component/simulation/outputs/outputs.py
```diff
@@ -0,0 +1,126 @@
...
+        default=False,
+        description="Enable writing all surface outputs into a single file instead of one file per surface. This option currently only supports Tecplot output format. Will choose the value of the last instance of this option of the same output type (SurfaceOutput or TimeAverageSurfaceOutput) in the `output` list.",
+    )
+    output_fields: List[str] = pd.Field()
+
+
+class TimeAverageSurfaceOutput(SurfaceOutput, _TimeAverageAdditionalAnimationSettings):
+    """
+    Caveats:
+    Solver side only accept exactly the same set of output_fields (is shared) between VolumeOutput and TimeAverageVolumeOutput.
```
maciej-flexcompute, line 71: how about surface?

flow360/component/simulation/outputs/output_entities.py
```diff
@@ -0,0 +1,35 @@
...
+    slice_normal: Tuple[float, float, float] = pd.Field()
+    slice_origin: Tuple[float, float, float] = pd.Field()
+
+
+@final
+class Isosurface(EntityBase):
+    _entity_type: Literal["IsosurfaceType"] = "IsosurfaceType"
+    surface_field: str = pd.Field()
+    # TODO: Maybe we need some unit helper function to help user figure out what is the value to use here?
+    surface_field_magnitude: float = pd.Field(description="Expect scaled value.")
```
maciej-flexcompute, line 21: Why the names are `surface field` and `surface field magnitude`? I think names of these parameters are incorrect. It is volume field of which we take isosurface, right? Shouldn't this be called just `field` and `field_magnitude` or `iso_value`? Isosurface of surface field is line, not surface.

I'd suggest to rename them.

flow360/component/simulation/outputs/outputs.py
```diff
@@ -0,0 +1,126 @@
...
+    pass
+
+
+class SliceOutput(_AnimationSettings):
+    slices: EntityList[Slice] = pd.Field()
+    output_fields: List[str] = pd.Field()
+
+
+class IsosurfaceOutput(_AnimationSettings):
+    isosurfaces: EntityList[Isosurface] = pd.Field()
```
maciej-flexcompute, line 98: I think isosurface should not have entities. Why is that not just a List[Isosurface] ?

flow360/component/simulation/outputs/outputs.py
```diff
@@ -0,0 +1,126 @@
...
+    output_fields: List[str] = pd.Field()
+
+
+class MonitorOutput(_AnimationSettings):
+    monitors: EntityList[SurfaceIntegralMonitor, ProbeMonitor] = pd.Field()
```
maciej-flexcompute, line 103: I think better will be to split these monitors to two separate classes:


```python
class PointMonitorOutput(_AnimationSettings):
     points: EntityList[Point] = pd.Field()
     output_fields: List[str] = pd.Field()

class SurfaceMonitorOutput(_AnimationSettings):
     monitors: EntityList[Surface] = pd.Field()
     output_fields: List[str] = pd.Field()
```

the only disadvantage is that you cannot specify the same output fields for point and surface monitors. But I am not sure if this is a big inconvenience and if this is the real problem. On surface you track forces, on points you track velocity/pressure.

flow360/component/simulation/types.py
maciej-flexcompute, line 1: maybe we can put them into existing types but we will split them into v1, v2 and common? 

example:
```python
from .types.v1 import PositiveFloat
```

flow360/component/simulation/outputs/output_entities.py
benflexcompute: Agreed

flow360/component/simulation/outputs/output_entities.py
```diff
@@ -0,0 +1,35 @@
...
+import pydantic as pd
+
+from flow360.component.simulation.framework.entity_base import EntityBase, EntityList
+from flow360.component.simulation.primitives import Surface
+
+
+@final
+class Slice(EntityBase):
+    _entity_type: Literal["SliceType"] = "SliceType"
+    slice_normal: Tuple[float, float, float] = pd.Field()
```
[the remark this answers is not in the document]
benflexcompute: Sure. Removed alias as it is not needed.

flow360/component/simulation/outputs/output_entities.py
```diff
@@ -0,0 +1,35 @@
...
+    slice_origin: Tuple[float, float, float] = pd.Field()
```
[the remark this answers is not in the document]
benflexcompute: Done

flow360/component/simulation/outputs/output_entities.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@final
+class Isosurface(EntityBase):
```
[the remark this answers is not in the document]
benflexcompute: `Isosurface` as well as `Slice` are indeed geometrical representation (they are composed of triangles with connectivity) but just not the ones that exist in the original mesh.

They do not have to inherit from `EntityBase` though since they will not get registered by anyone. I removed the inheritance. Not sure if this is what you mean.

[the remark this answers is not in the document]
benflexcompute: Removed inheritance.

flow360/component/simulation/outputs/output_entities.py
```diff
@@ -0,0 +1,35 @@
...
+    _entity_type: Literal["IsosurfaceType"] = "IsosurfaceType"
+    surface_field: str = pd.Field()
+    # TODO: Maybe we need some unit helper function to help user figure out what is the value to use here?
+    surface_field_magnitude: float = pd.Field(description="Expect scaled value.")
+
+
+@final
+class SurfaceIntegralMonitor(EntityBase):
```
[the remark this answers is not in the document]
benflexcompute: This may contains a collection of surfaces to perform integral on multiple surfaces. It is not restricted to just one surface.

flow360/component/simulation/outputs/output_entities.py
```diff
@@ -0,0 +1,35 @@
...
+    _entity_type: Literal["SurfaceIntegralMonitorType"] = "SurfaceIntegralMonitorType"
+    type: Literal["surfaceIntegral"] = pd.Field("surfaceIntegral", frozen=True)
+    entities: EntityList[Surface] = pd.Field(alias="surfaces")
+
+
+@final
+class ProbeMonitor(EntityBase):
```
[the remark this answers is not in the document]
benflexcompute: Point is probably not that intuitive. But I get your point that the Monitor part is redundent. I changed it to Probe.

flow360/component/simulation/outputs/output_entities.py
benflexcompute: Removed inheritance from EntityBase.

flow360/component/simulation/outputs/output_entities.py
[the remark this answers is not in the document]
benflexcompute: Also changed SurfaceIntegralMonitor to SurfaceIntegral

flow360/component/simulation/outputs/outputs.py
benflexcompute: It is hard to do that from data level design because we allow multiple instances of surfaceOutput inside the output list. They are all independent of each other. I think we will have to do it with validators on the Param level.

flow360/component/simulation/outputs/outputs.py
benflexcompute: This is a typo, same thing for surface and volume

flow360/component/simulation/outputs/outputs.py
benflexcompute: Sure

flow360/component/simulation/outputs/outputs.py
benflexcompute: Done

flow360/component/simulation/outputs/output_entities.py
benflexcompute: Done

flow360/component/simulation/outputs/outputs.py
benflexcompute: Done

flow360/component/simulation/outputs/outputs.py
benflexcompute: Seperated into two types of outputs.

flow360/component/simulation/types.py
benflexcompute: Actually I removed the new one. It is not that much simpler than adding pd. in front and other definitions are not that generic/universal.

flow360/component/simulation/outputs/outputs.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class IsosurfaceOutput(_AnimationSettings):
+    isosurfaces: UniqueItemList[Isosurface] = pd.Field()
+    output_fields: UniqueAliasedItemList[CommonFields] = pd.Field()
+
+
+class SurfaceIntegralOutput(_AnimationSettings):
+    monitors: UniqueItemList[SurfaceList] = pd.Field()
+    output_fields: UniqueAliasedItemList[CommonFields] = pd.Field()
```
maciej-flexcompute, line 106: I see the motivation here, but it seems it is too many levels now. Here is an example:

```python
sim = Simulation(
   ...
   outputs=[
      SurfaceIntegralOutput(
           monitor=[
                SurfaceList(entities=[geo['wing1'], geo['wing2']]),
                SurfaceList(entities=[geo['wing3'], geo['wing4']])
           ],
           output_fields=['Cp', 'Cf'],
       ),
  ]
  )

```
instead, I think this would be much easier:

```python
sim = Simulation(
   ...
   outputs=[
      SurfaceIntegralOutput(
            entities=[geo['wing1'], geo['wing2']],
            output_fields=['Cp', 'Cf'],
        ),
        SurfaceIntegralOutput(
            entities=[geo['wing3'], geo['wing4']],
            output_fields=['Cp', 'Cf'],
        ),
  ]
  )

```

flow360/component/simulation/outputs/outputs.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class ProbeOutput(_AnimationSettings):
+    probes: UniqueItemList[Probe] = pd.Field()
```
maciej-flexcompute, line 110: why 'Probe', not 'Point', and why probes, not entities?

flow360/component/simulation/outputs/outputs.py
```diff
@@ -0,0 +1,134 @@
...
+    output_fields: UniqueAliasedItemList[CommonFields] = pd.Field()
+
+
+class AeroAcousticOutput(Flow360BaseModel):
+    patch_type: Optional[str] = pd.Field("solid", frozen=True)
+    observers: List[Tuple[float, float, float]] = pd.Field()
```
maciej-flexcompute, line 116: observers are type o points as well, right?

flow360/component/simulation/framework/unique_list.py
```diff
@@ -0,0 +1,70 @@
...
+            f"Input item to this list must be unique but {[str(item) for item, count in Counter(v).items() if count > 1]} appears multiple times."
+        )
+    return v
+
+
+def _validate_unique_aliased_item(v: List[str]) -> List[str]:
+    for item in v:
+        if get_aliases(item)[1] in v:
+            raise ValueError(
+                f"Input item to this list must be unique but {item} and {get_aliases(item)[1]} are both present."
```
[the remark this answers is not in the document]
benflexcompute: we can for the output_field.

But for Isosurface instances, for example, we check isosurface names only and if the user have two isosurface with **same** name but different settings, which do we remove?

flow360/component/simulation/framework/unique_list.py
```diff
@@ -0,0 +1,70 @@
...
+            (cls,),
+            {"__annotations__": annotations},
+        )
+        return new_cls
+
+
+def _validate_unique_list(v: List) -> List:
+    if len(v) != len(set(v)):
+        raise ValueError(
+            f"Input item to this list must be unique but {[str(item) for item, count in Counter(v).items() if count > 1]} appears multiple times."
```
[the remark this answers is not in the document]
benflexcompute: Added deduplication for the outputField

flow360/component/simulation/outputs/outputs.py
benflexcompute: PointOutput is strange as Point is a geometrical concept. Probe is used in both Tecplot and Paraview which is more related to the monitor concept.

flow360/component/simulation/outputs/outputs.py
benflexcompute: I see your point but then we need to change all of the outputs. Otherwise there will be a bit of inconsistency in the sense that "output_fields" after modification is no longer a shared concept and name is now specified at the root level.
See for example:
```
outputs = [
    SurfaceIntegralOutput(
        name = "monitor-1"
        entities=[geo["wing1"], geo["wing2"]],
        output_fields=["Cp", "Cf"],
    ),
    SurfaceIntegralOutput(
        name = "monitor-2"
        entities=[geo["wing3"], geo["wing4"]],
        output_fields=["Cp", "Cf"],
    ),
    SliceOutput(
        entities = [
            Slice(
                name = "slice-1",
                slice_normal = [1.0, 0.0, 0.0],
                slice_origin = [0.0, 0.0, 1.0],
            ),
            Slice(
                name = "slice-2",
                slice_normal = [1.0, 0.0, 0.0],
                slice_origin = [0.0, 1.0, 1.0],
            ),
        ],
        output_fields=["primitiveVars"],
    )
]
```
The problem is that once we change **ALL** outputs then there is no way to share output_fields for multiple slices which is a huge step backward.

flow360/component/simulation/outputs/outputs.py
benflexcompute: I changed the probes to entities but keep probe as alias. entities is too generic I think for user to understand what they should put here.

flow360/component/simulation/outputs/outputs.py
benflexcompute: Good catch. Changed.

maciej-flexcompute (approved):

PR #272, opened 2024-05-22
[ModelRef] Added Field Definitions for **TimeStepping**

benflexcompute:


flow360/component/simulation/time_stepping/time_stepping.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class SteadyTimeStepping(Flow360BaseModel):
+    """
+    Steady time stepping component
+    """
+
+    model_type: Literal["Steady"] = pd.Field("Steady", frozen=True)
+    physical_steps: Literal[1] = pd.Field(1, frozen=True)
+    time_step_size: Literal["inf"] = pd.Field("inf", frozen=True)
```
maciej-flexcompute, line 93: do we need to keep these?

flow360/component/simulation/time_stepping.py
maciej-flexcompute, line 1: why do we use folders additionally?

flow360/component/simulation/time_stepping/time_stepping.py
```diff
@@ -0,0 +1,139 @@
...
+    #  TODO:
+    # # pylint: disable=arguments-differ
+    # def to_solver(self, params, **kwargs) -> BaseTimeStepping:
+    #     """
+    #     returns configuration object in flow360 units system
+    #     """
+    #     return super().to_solver(params, **kwargs)
+
+
+class SteadyTimeStepping(Flow360BaseModel):
```
maciej-flexcompute, line 86: I think we are missing `max_pseudo_steps` here. Maybe we can just call them `max_steps` ? Just an idea

flow360/component/simulation/time_stepping/time_stepping.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class UnsteadyTimeStepping(Flow360BaseModel):
+    """
+    Unsteady time stepping component
+    """
+
+    model_type: Literal["Unsteady"] = pd.Field("Unsteady", frozen=True)
+    physical_steps: pd.PositiveInt = pd.Field()
+    time_step_size: TimeType.Positive = pd.Field()
```
maciej-flexcompute, line 121: I think these 2 parameter names can be unified:
```
time_steps
time_step_size
```
or
```
steps
step_size
```

flow360/component/simulation/time_stepping/time_stepping.py
```diff
@@ -0,0 +1,139 @@
...
+    CFL: Optional[Union[RampCFL, AdaptiveCFL]] = pd.Field(
+        default=AdaptiveCFL.default_unsteady(),
+    )
+
```
maciej-flexcompute, line 125: we are missing number of pseudo steps

flow360/component/simulation/time_stepping/time_stepping.py
```diff
@@ -0,0 +1,139 @@
...
+            return values  # will be handeled by default value
+        cfl_input = values["CFL"]
+        if isinstance(cfl_input, AdaptiveCFL):
+            cfl_input = _apply_default_to_none(cfl_input, AdaptiveCFL.default_steady())
+        elif isinstance(cfl_input, RampCFL):
+            cfl_input = _apply_default_to_none(cfl_input, RampCFL.default_steady())
+        return values
+
+
+class UnsteadyTimeStepping(Flow360BaseModel):
```
maciej-flexcompute, line 114: should inherit from BaseTimeStepping, right

flow360/component/simulation/time_stepping.py
benflexcompute: The goal is to have all keys in simulationParam under a same name folder.
Like models, operation_conditions etc for better organization. I think we might be able to take advantage of this nice organization during documentation effort.

flow360/component/simulation/time_stepping/time_stepping.py
benflexcompute: This is a typo. I should have inherited from `BaseTimeStepping` which has the max_pseudo_steps.
`max_step` is not confusing in steady but is in unsteady.

flow360/component/simulation/time_stepping/time_stepping.py
benflexcompute: again it is collision with max_psuedo_step.

flow360/component/simulation/time_stepping/time_stepping.py
benflexcompute: Yes

flow360/component/simulation/time_stepping.py
maciej-flexcompute: I see. I think this extra layer is not necessary. It makes imports looking like this:
```python
time_stepping.time_stepping.TimeStepping
```

flow360/component/simulation/time_stepping/time_stepping.py
maciej-flexcompute: My proposition was to have different names for this parameter on steady and unsteady. If you think of using only one of these classes, parameter names are very confusing. They only have sense when you look at both of the classes at the same time.
```python
UnsteadyTimeStepping(physical_steps=10,  time_step_size=1, max_pseudo_steps=10)
```
These are lengthy parameter names. Instead I would make parameters that are more often used or required, to be easier available. For example:

```python
UnsteadyTimeStepping(steps=10, step_size=1, max_pseudo_steps=10)
```

Similarly, steady time stepping does not need to be considered in the context of unsteady:
```python
SteadyTimeStepping(max_pseudo_steps=10)
```
instead, I think better it is:
```python
SteadyTimeStepping(max_steps=10)
```

There is no single parameter that has two meanings in two contexts so I think there is no collision. It is just a proposition, if you think this will be very confusing we don't need to do it. Just don't think about current users, think about new users.

flow360/component/simulation/time_stepping.py
benflexcompute: but we do not expose this to user right? the user would just do 
```python
from flow360 import UnsteadyTimeStepping, SteadyTimeStepping
``` etc.

Comment (benflexcompute) [reply flow360/component/simulation/time_stepping.py:1]: I would strongly prefer having things grouped. The timestepping laying in the root level while the output for example has its own folder is confusing and inconsistent.

File: flow360/component/simulation/time_stepping/time_stepping.py (hunk 3 above)
Comment (benflexcompute) [reply flow360/component/simulation/time_stepping/time_stepping.py:119]: we can do this.

Comment (maciej-flexcompute) [reply flow360/component/simulation/time_stepping/time_stepping.py]: yes, I like the `Steady`

File: flow360/component/simulation/time_stepping/time_stepping.py [hunk 6]
```diff
@@ -0,0 +1,139 @@
...
+
+class Unsteady(BaseTimeStepping):
+    """
+    Unsteady time stepping component
+    """
+
+    model_type: Literal["Unsteady"] = pd.Field("Unsteady", frozen=True)
+    max_pseudo_steps: int = pd.Field(
+        2000, gt=0, le=100000, description="Maximum pseudo steps within one physical step."
+    )
```
maciej-flexcompute, line 119: the default should be different here. Not user if we want to work on defaults now?

maciej-flexcompute (approved):

flow360/component/simulation/time_stepping/time_stepping.py
benflexcompute: I think the default comes from BaseTimeStepping when looking at Flow360Params which is 2000. Also the doc says 2000 as well.

benflexcompute: Ok there was a threshold of 100 from old validaitons. Changed to that.

PR #273, opened 2024-05-23
[ModelRef] Added Field Definitions for **VolumeModels**

benflexcompute:


flow360/component/simulation/models/material.py
```diff
@@ -0,0 +1,67 @@
...
+class MaterialBase(Flow360BaseModel):
+    # Basic properties required to define a material.
+    # For example: young's modulus, viscosity as an expression of temperature, etc.
+    ...
+    name: str = pd.Field()
+
+
+class Gas(MaterialBase):
+    # contains models of getting properites, for example US standard atmosphere model
+    name: Literal["gas"] = pd.Field(frozen=True)
```
benflexcompute, line 17: name default?

flow360/component/simulation/models/material.py
```diff
@@ -0,0 +1,67 @@
...
+    viscosity_type: Literal["Sutherland", "constant"] = pd.field("Sutherland")
+    constant_dynamic_viscosity: Optional[ViscosityType] = pd.Field(None)
```
benflexcompute, line 19: Can you add 2 classes for handling `Sutherland` and `constant` respectively? Move for example constant_dynamic_viscosity inside that model.

This avoids writing validations.

```python
...
viscosity_type = SutherLand(SutherlandConstants = ...),
...
```

flow360/component/simulation/models/material.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+    specific_heat_ratio: float = pd.Field(1.4, frozen=True)
+    gas_constant: GasConstantType = pd.Field(287.0529, frozen=True)
+    Prandtl_number: float = pd.Field(0.72, frozen=True)
+
+
+class Solid(MaterialBase):
+    name: Literal["solid"] = pd.Field(frozen=True)
+    thermal_conductivity: ThermalConductivityType = pd.Field()
+    density: Optional[ThermalConductivityType] = pd.Field(None)
```
benflexcompute, line 56: DensityType?

flow360/component/simulation/models/material.py
```diff
@@ -0,0 +1,67 @@
...
+
+class Air(Gas):
+    name: Literal["air"] = pd.Field(frozen=True)
+    constant_dynamic_viscosity: ViscosityType = pd.Field(1.825e-5)
+    sutherland_constants: SutherlandConstants = pd.Field(
+        SutherlandConstants(
+            reference_viscosity=1.716e-5, reference_temperature=273, effective_temperature=111
+        )
+    )
+    specific_heat_ratio: float = pd.Field(1.4, frozen=True)
```
benflexcompute, line 48: just FYI that user can still override the value here even frozen.
for example
```python
my_air = Air(specific_heat_ratio = 1000)
```
passes validation. Why do we want these frozen?

flow360/component/simulation/models/solver_numerics.py
```diff
@@ -0,0 +1,411 @@
...
+
+    model_config = pd.ConfigDict(
+        conflicting_fields=[Conflicts("absolute_tolerance", "relative_tolerance")]
+    )
+
+
+class GenericSolverSettings(Flow360BaseModel, metaclass=ABCMeta):
+    """:class:`GenericSolverSettings` class"""
+
+    absolute_tolerance: Optional[PositiveFloat] = pd.Field(1.0e-10)
```
benflexcompute, line 75: No need for Optional here?

flow360/component/simulation/models/solver_numerics.py
```diff
@@ -0,0 +1,411 @@
...
+    """
+
+    CFL_multiplier: PositiveFloat = pd.Field(1.0)
+    kappa_MUSCL: pd.confloat(ge=-1, le=1) = pd.Field(-1)
+
+    numerical_dissipation_factor: pd.confloat(ge=0.01, le=1) = pd.Field(1)
+    limit_velocity: bool = pd.Field(False)
+    limit_pressure_density: bool = pd.Field(False)
+
+    model_type: Literal["Compressible"] = pd.Field("Compressible", alias="modelType", const=True)
```
benflexcompute, line 157: Let's remove these aliases, these exists because we wanted to export to Flow360 JSON which uses camelCase.

benflexcompute: Also the const is deprecated.

flow360/component/simulation/models/solver_numerics.py
```diff
@@ -0,0 +1,411 @@
...
+            if self.low_mach_preconditioner_threshold is None:
+                self.low_mach_preconditioner_threshold = params.freestream.Mach
+
+        return super().to_solver(self, **kwargs)
+
+
+class SpalartAllmarasModelConstants(Flow360BaseModel):
+    """:class:`SpalartAllmarasModelConstants` class"""
+
+    model_type: Literal["SpalartAllmarasConsts"] = pd.Field(
```
benflexcompute, line 178: Same here.

benflexcompute: Let's try fixing the pydantic V1 issues in the file.

flow360/component/simulation/models/volume_models.py
```diff
@@ -0,0 +1,128 @@
...
+from flow360.component.simulation.framework.base_model import Flow360BaseModel
+from flow360.component.simulation.models.material import Air, MaterialType
+from flow360.component.simulation.models.solver_numerics import (
+    NavierStokesSolver,
+    TransitionModelSolver,
+    TurbulenceModelSolverType,
+)
+
+
+class VolumeModelBase(Flow360BaseModel):
```
benflexcompute, line 14: Maybe instead have 
```
class VolumeModelBase --> volumes: EntityList[Union[GenericVolume, str]] = pd.Field()
class VolumePhysicsBase(VolumeModelBase) --> material: MaterialType = pd.Field(); nitial_conditions: Optional[dict] = pd.Field(None)
```
so you write preprocess once for expansion, also the naming for VolumeModelBase and FluidDynamicsAddOnBase is kindof awkward (FluidDynamicsAddOnBase and FluidDynamics).

flow360/component/simulation/models/material.py
yifan-flex: specific_heat_ratio shouldn't be allowed to change for Air. I need to overwrite with Literal then?

flow360/component/simulation/models/solver_numerics.py
yifan-flex: If the user wants to do absolute_tolerance = None it should be fine

flow360/component/simulation/models/volume_models.py
yifan-flex: What does separating VolumeModelBase and VolumePhysicsBase do?
What's wrong the naming? FluidDynamicsAddOnBase looks unnecessary though, I can just remove it

flow360/component/simulation/models/solver_numerics.py
benflexcompute: Then what if absolute and relative residual are both none?

flow360/component/simulation/models/volume_models.py
benflexcompute: Yes let's then remove FluidDynamicsAddOnBase.

benflexcompute: The expansion logic then only exist in VolumeModelBase and you write it once. All models that inherit from the VolumeModelBase will then expand automatically. Although this probably is already included/handeled in the "preprocess/to_solver" mechanism. Ignore this if true.

The naming of `FluidDynamicsAddOnBase` and `VolumeModelBase` make it seem/sounds like the `VolumeModelBase` is the universal and general one while `FluidDynamicsAddOnBase` is some specialization of `VolumeModelBase` by adding one some more field. The FluidDynamics also coincide with the FluidDynamics model name but the FluidDynamics does not use this addon class.

I name it `VolumeModelBase` and `VolumePhysicsBase` because the reason we have `material` and `initial_condition` is because we are now in a different physics and therefore non-tightly coupled zone in your word.

Yes you can remove the `FluidDynamicsAddOnBase`. We will then need to rename the `VolumeModelBase` as it is not intuitive why some of the volume models inherit from `VolumeModelBase` but some does not.

flow360/component/simulation/models/material.py
```diff
@@ -0,0 +1,67 @@
...
+
+class Gas(MaterialBase):
+    # contains models of getting properites, for example US standard atmosphere model
+    name: Literal["gas"] = pd.Field(frozen=True)
+    viscosity_type: Literal["Sutherland", "constant"] = pd.field("Sutherland")
+    constant_dynamic_viscosity: Optional[ViscosityType] = pd.Field(None)
+    sutherland_constants: Optional[SutherlandConstants] = pd.Field(None)
+    specific_heat_ratio: float = pd.Field()
+    gas_constant: GasConstantType = pd.Field()
+    Prandtl_number: float = pd.Field()
```
maciej-flexcompute, line 23: let's not capitalise after name

flow360/component/simulation/models/volume_models.py
```diff
@@ -0,0 +1,128 @@
...
+    volumes: EntityList[Union[GenericVolume, str]] = pd.Field()
+
+
+class FluidDynamics(VolumeModelBase):
+    """
+    General FluidDynamics volume model that contains all the common fields every fluid dynamics zone should have.
+    """
+
+    navier_stokes_solver: Optional[NavierStokesSolver] = pd.Field(None)
+    turbulence_model_solver: Optional[TurbulenceModelSolverType] = pd.Field(None)
```
maciej-flexcompute, line 35: why defaults are none? We probably want defaults here

flow360/component/simulation/models/volume_models.py
```diff
@@ -0,0 +1,128 @@
...
+    transition_model_solver: Optional[TransitionModelSolver] = pd.Field(None)
+
+    material: MaterialTypes = pd.Field(Air())
```
maciej-flexcompute, line 38: is user supposed to change it? Does our solver support it?

flow360/component/simulation/models/volume_models.py
```diff
@@ -0,0 +1,128 @@
...
+)
+
+
+class VolumeModelBase(Flow360BaseModel):
+    """
+    Base class for equation models
+
+    """
+
+    volumes: EntityList[Union[GenericVolume, str]] = pd.Field()
```
maciej-flexcompute, line 20: why change entities to volumes? @benflexcompute is this our new convention? Maybe we can use alias?

flow360/component/simulation/models/volume_models.py
```diff
@@ -0,0 +1,128 @@
...
+
+class ActuatorDisk(FluidDynamicsAddOnBase):
+    """Same as Flow360Param ActuatorDisks.
+    Note that `center`, `axis_thrust`, `thickness` can be acquired from `entity` so they are not required anymore.
+    """
+
+    volumes: Optional[EntityList[Cylinder]] = pd.Field(None)
+
+    center: Optional[Coordinate] = pd.Field(None)
+    thickness: Optional[PositiveFloat] = pd.Field(None)
```
maciej-flexcompute, line 58: center and thickness will be in Cylinder as per docstring. I think we can remove them

flow360/component/simulation/models/volume_models.py
```diff
@@ -0,0 +1,128 @@
...
+    Note that `center_of_rotation`, `axis_of_rotation`, `radius`, `thickness` can be acquired from `entity` so they are not required anymore.
+    """
+
+    volumes: Optional[EntityList[Cylinder]] = pd.Field(None)
+
+    rotation_direction_rule: Literal["leftHand", "rightHand"] = pd.Field("rightHand")
+    center_of_rotation: Optional[Coordinate] = pd.Field(None)
+    axis_of_rotation: Optional[Axis] = pd.Field(None)
+    thickness: Optional[LengthType.Positive] = pd.Field(None)
+    radius: Optional[LengthType.Positive] = pd.Field(None)
```
maciej-flexcompute, line 74: same as for AD, I think we can remove

flow360/component/simulation/models/volume_models.py
```diff
@@ -0,0 +1,128 @@
...
+    number_of_blades: pd.conint(strict=True, gt=0, le=10) = pd.Field()
+    omega: AngularVelocityType.NonNegative = pd.Field()
```
maciej-flexcompute, line 76: why is this non-negative? Is it only controlled by `rotation_direction_rule` ?

flow360/component/simulation/models/volume_models.py
```diff
@@ -0,0 +1,128 @@
...
+    Note: Should use the unit system to convert degree or degree per second to radian and radian per second
+    """
+
+    # (AKA omega) In conflict with `rotation_per_second`.
+    angular_velocity: Optional[Union[float, pd.StrictStr]] = pd.Field(None)
+    # (AKA theta) Must be expression otherwise zone become static.
+    rotation_angle_radians: Optional[pd.StrictStr] = pd.Field(None)
+    parent_volume_name: Optional[VolumeEntityTypes] = pd.Field(None)
+    center: Optional[LengthType.Point] = pd.Field(None)
+    axis: Optional[Axis] = pd.Field(None)
```
maciej-flexcompute, line 103: same here. The problem is that we need to "detect" center and axis from volume mesh, which I think we do not have implemented yet

flow360/component/simulation/models/volume_models.py
```diff
@@ -0,0 +1,128 @@
...
+class Rotation(FluidDynamicsAddOnBase):
+    """Similar to Flow360Param ReferenceFrame.
+    Note that `center`, `axis` can be acquired from `entity` so they are not required anymore.
+    Note: Should use the unit system to convert degree or degree per second to radian and radian per second
+    """
+
+    # (AKA omega) In conflict with `rotation_per_second`.
+    angular_velocity: Optional[Union[float, pd.StrictStr]] = pd.Field(None)
+    # (AKA theta) Must be expression otherwise zone become static.
+    rotation_angle_radians: Optional[pd.StrictStr] = pd.Field(None)
```
maciej-flexcompute, line 100: we probably need to unify the movement specification, it can be either const velocity or expression as a function of theta

flow360/component/simulation/models/volume_models.py
```diff
@@ -0,0 +1,128 @@
...
+
+    volumes: EntityList[Volume, Box, str] = pd.Field()
+
+    darcy_coefficient: InverseAreaType.Point = pd.Field()
+    forchheimer_coefficient: InverseLengthType.Point = pd.Field()
+    volumetric_heat_source: Optional[HeatSourceType] = pd.Field(None)
+    # box specific
+    axes: Optional[List[Axis]] = pd.Field(None)
+    center: Optional[LengthType.Point] = pd.Field(None)
+    lengths: Optional[LengthType.Moment] = pd.Field(None)
```
maciej-flexcompute, line 117: I think we should remove them and use only volumes, right?

flow360/component/simulation/models/material.py
yifan-flex: You mean I shouldn't capitalize Prandtl? It is conventionally capitalized though. We also have `Mach`

flow360/component/simulation/models/volume_models.py
yifan-flex: Changed defaults. Why did we have `NoneSolver()` in flow360_params? What's the reason for using it instead of `None`?

flow360/component/simulation/models/volume_models.py
yifan-flex: Renamed `VolumeModelBase` to `PDEModelBase`

flow360/component/simulation/models/volume_models.py
yifan-flex: This is here so that we still support the definition AD if no volume entities are defined by user

flow360/component/simulation/models/volume_models.py
yifan-flex: Same as above

flow360/component/simulation/models/volume_models.py
yifan-flex: Yes, this is supposed to be speed

flow360/component/simulation/models/volume_models.py
yifan-flex: @benflexcompute

flow360/component/simulation/models/volume_models.py
yifan-flex: `rotation: Union[AngularVelocityType, pd.StrictStr] = pd.Field()`
If str is specified it's interpreted as theta. Is this confusing?

flow360/component/simulation/models/volume_models.py
yifan-flex: Same as AD and BET

flow360/component/simulation/models/solver_numerics.py
yifan-flex: We need validation here anyways right? Absolute and relative residual cannot both be set
Actually it might be better to merge the two fields

flow360/component/simulation/models/volume_models.py
yifan-flex: For now Air is the only fluid material model exposed. Users can change reference_viscosity and reference_temperature in Sutherland law

flow360/component/simulation/models/material.py
maciej-flexcompute: yes, this is what I mean. Let's not capitalise after name. It should be `mach`, `reynolds`, `prandtl`

flow360/component/simulation/models/volume_models.py
maciej-flexcompute: I think this was legacy reason. Solver expected turbulence having a type "None", not being equal to None. It applies defaults when solver is not provided. Also, there was some difficulties with heat solver, I remember. So we ended up implementing `NoneSolver`. But I am up for removing `NoneSolver` if we can do it.

flow360/component/simulation/models/material.py
```diff
@@ -0,0 +1,55 @@
...
+    name: Literal["solid"] = pd.Field("solid", frozen=True)
+    thermal_conductivity: ThermalConductivityType = pd.Field()
+    density: Optional[DensityType] = pd.Field(None)
+    specific_heat_capacity: Optional[SpecificHeatCapacityType] = pd.Field(None)
+
+
+class Aluminum(Solid):
+    name: Literal["Aluminum"] = pd.Field("Aluminum", frozen=True)
+    thermal_conductivity: Literal["235"] = pd.Field("235", frozen=True)
+    density: Literal["2710"] = pd.Field("2710", frozen=True)
```
benflexcompute, line 51: similarly here.

flow360/component/simulation/models/material.py
```diff
@@ -0,0 +1,55 @@
...
+    # For example: young's modulus, viscosity as an expression of temperature, etc.
+    ...
+    name: str = pd.Field()
+
+
+class Gas(MaterialBase):
+    name: Literal["gas"] = pd.Field("gas", frozen=True)
+    dynamic_viscosity: Union[ViscosityType, Sutherland] = pd.Field()
+    specific_heat_ratio: float = pd.Field()
+    gas_constant: GasConstantType = pd.Field()
```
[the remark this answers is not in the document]
yifan-flex: We need to define it

flow360/component/simulation/models/volume_models.py
```diff
@@ -0,0 +1,152 @@
...
+    number_of_blades: pd.conint(strict=True, gt=0, le=10) = pd.Field()
+    omega: AngularVelocityType.NonNegative = pd.Field()
+    chord_ref: LengthType.Positive = pd.Field()
+    n_loading_nodes: pd.conint(strict=True, gt=0, le=1000) = pd.Field()
+    blade_line_chord: LengthType.NonNegative = pd.Field(0)
+    initial_blade_direction: Optional[Axis] = pd.Field(None)
+    tip_gap: Union[LengthType.NonNegative, Literal["inf"]] = pd.Field("inf")
+    mach_numbers: List[NonNegativeFloat] = pd.Field()
+    reynolds_numbers: List[PositiveFloat] = pd.Field()
+    alphas: List[float] = pd.Field()
```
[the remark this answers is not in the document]
yifan-flex: From Maciej: "For alpha, beta, can we just assume these are in degrees and take floats, not unit quantities? This is will be so easy to make mistake because default units of angle are radians. Alternatively, we can introduce another unit in the unit system but this seems to be an unnecessary complication"

maciej-flexcompute (approved):

PR #274, opened 2024-05-23
[ModelRef] Add Field Definitions for **SurfaceModels**

yifan-flex:


flow360/component/simulation/models/surface_models.py
```diff
@@ -0,0 +1,128 @@
...
+
+    model_config = ConfigDict(
+        conflicting_fields=[Conflicts("temperature", "heat_flux")],
+    )
+
+    type: Literal["Wall"] = pd.Field("Wall", frozen=True)
+    use_wall_function: bool = pd.Field(False)
+    velocity: Optional[BoundaryVelocityType] = pd.Field(None)
+    velocity_type: Literal["absolute", "relative"] = pd.Field("relative")
+    temperature: Optional[Union[pd.PositiveFloat, pd.StrictStr]] = pd.Field()
```
benflexcompute, line 44: Let's just copy the Flow360Param unit type here. Unit system should be ready soon.

flow360/component/simulation/models/surface_models.py
```diff
@@ -0,0 +1,128 @@
...
+    heat_flux: Optional[Union[float, pd.StrictStr]] = pd.Field(None)
+
+
+class VelocityFromMachAndAngles(Flow360BaseModel):
+    Mach: float = pd.Field()
+    beta: float = pd.Field(0)
+    alpha: float = pd.Field(0)
```
benflexcompute, line 51: I thought you think alpha beta should not have default 0?

flow360/component/simulation/models/surface_models.py
```diff
@@ -0,0 +1,128 @@
...
+class Outflow(BoundaryBaseWithTurbulenceQuantities):
+    """Replace Flow360Param:
+    - SubsonicOutflowPressure
+    - SubsonicOutflowMach
+    - MassOutflow
+    """
+
+    type: Literal["Outflow"] = pd.Field("Outflow", frozen=True)
+    pressure: Optional[PressureType] = pd.Field(None)
+    Mach: Optional[float] = pd.Field(None)
```
benflexcompute, line 78: what if user wants to use velocity?

flow360/component/simulation/models/surface_models.py
```diff
@@ -0,0 +1,128 @@
...
+    type: Literal["Inflow"] = pd.Field("Inflow", frozen=True)
+    total_temperature: TemperatureType = pd.Field()
+    velocity_direction: Optional[Axis] = pd.Field(None)
+    total_pressure: Optional[PressureType] = pd.Field(None)
+    mass_flow_rate: Optional[MassFluxType] = pd.Field(None)
+
+
+class TranslationallyPeriodic(BoundaryBase):
+    type: Literal["TranslationallyPeriodic"] = pd.Field("TranslationallyPeriodic", frozen=True)
+    paired_patch_name: Optional[str] = pd.Field(None)
```
benflexcompute, line 102: Use entity.

flow360/component/simulation/models/surface_models.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class RotationallyPeriodic(BoundaryBase):
+    type: Literal["RotationallyPeriodic"] = pd.Field("RotationallyPeriodic", frozen=True)
+    paired_patch_name: Optional[str] = pd.Field(None)
```
benflexcompute, line 107: Same here

flow360/component/simulation/models/surface_models.py
yifan-flex: That's not implemented in our solver

flow360/component/simulation/models/surface_models.py
yifan-flex: I was thinking of validation complications before `VelcoityFrom` models were created

flow360/component/simulation/models/surface_models.py
yifan-flex: Our solver currently only supports user supplied pressure or Mach number at the outflow boundary. The state here is not fully known, so specifying Mach number is not the same as specifying velocity

flow360/component/simulation/models/surface_models.py
```diff
@@ -0,0 +1,130 @@
...
+    type: str = pd.Field()
+    entities: EntityList[Surface] = pd.Field(alias="surfaces")
+
+
+class BoundaryBaseWithTurbulenceQuantities(BoundaryBase, metaclass=ABCMeta):
+    turbulence_quantities: Optional[TurbulenceQuantitiesType] = pd.Field(None)
+
+
+class HeatFlux(Flow360BaseModel):
+    heat_flux: Union[HeatFluxType, pd.StrictStr] = pd.Field()
```
benflexcompute, line 26: For these single field class that takes in simple objects I was thinking adding the before model validator to pass the input to the field so that user writes:
`HeatFlux(2*u.J/u.m/u.m)` instead of
`HeatFlux(heat_flux = 2*u.J/u.m/u.m)`

flow360/component/simulation/models/surface_models.py
```diff
@@ -0,0 +1,130 @@
...
+    - WallFunction
+    - SolidIsothermalWall
+    - SolidAdiabaticWall
+    """
+
+    type: Literal["Wall"] = pd.Field("Wall", frozen=True)
+    use_wall_function: bool = pd.Field(False)
+    velocity: Optional[VelocityVectorType] = pd.Field(None)
+    velocity_type: Literal["absolute", "relative"] = pd.Field("relative")
+    heat_spec: Optional[Union[HeatFluxType, TemperatureType, HeatFlux, Temperature]] = pd.Field(
```
benflexcompute, line 63: why is this a union of 4 types?

flow360/component/simulation/models/surface_models.py
```diff
@@ -0,0 +1,130 @@
...
+
+class Outflow(BoundaryBaseWithTurbulenceQuantities):
+    """Replace Flow360Param:
+    - SubsonicOutflowPressure
+    - SubsonicOutflowMach
+    - MassOutflow
+    """
+
+    type: Literal["Outflow"] = pd.Field("Outflow", frozen=True)
+    spec: Union[PressureType, MassFluxType.NonNegative, Pressure, MassFlowRate, Mach] = pd.Field()
```
benflexcompute, line 82: Same here. The `PressureType `is included in `Pressure `right?

flow360/component/simulation/models/surface_models.py
```diff
@@ -0,0 +1,116 @@
...
+class Inflow(BoundaryBaseWithTurbulenceQuantities):
+    """Replace Flow360Param:
+    - SubsonicInflow
+    - MassInflow
+    """
+
+    model_config = ConfigDict(
+        conflicting_fields=[Conflicts("total_pressure", "mass_flow_rate")],
+        require_one_of=["total_pressure", "mass_flow_rate"],
+    )
```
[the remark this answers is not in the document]
benflexcompute: Yes I agree. We should even go extra mile and make this what the "type" contains. Since the "type" of "Inflow" is a "TotalPressureInflow". The old type can be in other names which is not important as just a discriminator.

flow360/component/simulation/models/surface_models.py
benflexcompute: This can wait though as we have so much on the plate right now.

flow360/component/simulation/models/surface_models.py
maciej-flexcompute: you mean boundary condition is on enforcing local Mach number which depends on local temperature. How users come up with Mach number for the outflow?

flow360/component/simulation/models/surface_models.py
maciej-flexcompute: yes, I was thinking the same

flow360/component/simulation/models/surface_models.py
```diff
@@ -0,0 +1,130 @@
...
+class Pressure(Flow360BaseModel):
+    pressure: PressureType = pd.Field()
+
+
+class MassFlowRate(Flow360BaseModel):
+    mass_flow_rate: MassFluxType.NonNegative = pd.Field()
+
+
+class Mach(Flow360BaseModel):
+    Mach: NonNegativeFloat = pd.Field()
```
maciej-flexcompute, line 46: maybe we can have only one attribute named `value`

flow360/component/simulation/models/surface_models.py
```diff
@@ -0,0 +1,130 @@
...
+    velocity_type: Literal["absolute", "relative"] = pd.Field("relative")
+    heat_spec: Optional[Union[HeatFluxType, TemperatureType, HeatFlux, Temperature]] = pd.Field(
+        None
+    )
+
+
+class Freestream(BoundaryBaseWithTurbulenceQuantities):
+    type: Literal["Freestream"] = pd.Field("Freestream", frozen=True)
+    velocity: Optional[VelocityVectorType] = pd.Field(None)
+    velocity_type: Literal["absolute", "relative"] = pd.Field("relative")
```
maciej-flexcompute, line 71: what is relative on freestream?

flow360/component/simulation/models/surface_models.py
maciej-flexcompute: actually it is easy:

```python
class _SingleAttributeModel(Flow360BaseModel):
    def __init__(self, *args, **kwargs) -> None:    
        if len(args) == 1 and 'value' not in kwargs:
            kwargs['value'] = args[0]
        super().__init__(**kwargs)
```

flow360/component/simulation/models/surface_models.py
benflexcompute: Let's keep as is for now.

flow360/component/simulation/models/surface_models.py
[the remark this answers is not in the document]
benflexcompute: Agreed to use "spec"

benflexcompute (approved):

flow360/component/simulation/models/surface_models.py
benflexcompute: Note: We will implement auto type detection. and convert input to Class(value = input)

PR #275, opened 2024-05-23
[ModelRef] Add Field Definitions for **OperatingCondition**

yifan-flex:


flow360/component/simulation/operating_condition.py
```diff
@@ -4,47 +4,24 @@
...
-    beta: float = pd.Field(0)
-    temperature: float = pd.Field(288.15)
-    reference_velocity: Optional[float] = pd.Field()  # See U_{ref} definition in our documentation
 
-    initial_flow_condition: Optional[tuple[str, str, str, str, str]] = pd.Field(
-        ("NotImplemented", "NotImplemented", "NotImplemented")
-    )
-    turbulence_quantities = Optional[TurbulenceQuantities] = pd.Field()
+class StandardAtmosphere(Flow360BaseModel):
+    altitude: LengthType = pd.Field(0)
```
benflexcompute, line 14: maybe altitude and temperature can be another set of combination (if valid)? I thought standard atmosphere condition is a const condition for example the altitude is just 0?

flow360/component/simulation/operating_condition.py
```diff
@@ -4,47 +4,24 @@
...
-    inlet_velocity: float = pd.Field()
-
-
-class SolidOperatingConditions(Flow360BaseModel):
-    initial_temperature: float = pd.Field()
-
+class OperatingConditions(Flow360BaseModel):
+    """
+    Defines all the operating conditions for different physics.
+    The type of operating condition has to match its volume type.
```
benflexcompute, line 21: Do we still plan to allow operating condition to be overwritten per volume?

yifan-flex: I think we decided not to allow per volume operating condition right? If the user wants a different freestream they can just change the boundary condition, so there's no use of operation condition per zone

flow360/component/simulation/operating_condition.py
yifan-flex: standard atmosphere is a model that relates the altitude to pressure, density and temperature.

maciej-flexcompute: Right. It is such model: https://www.digitaldutch.com/atmoscalc/

we have it already implemented for panel

examples/simulation_examples/Simulation_interface_illustration.py
```diff
@@ -66,38 +66,33 @@
...
+            condition=TemperatureDensity(temperature=300, density=1.225),
+        ),
+        reference_geometry=ReferenceGeometry(
+            area=1.15315084119231,
+            moment_length=(0.801672958512342, 0.801672958512342, 0.801672958512342),
+        ),
+        models=[
             FluidDynamics(
-                entities=["*"],  # This means we apply settings to all the volume zones
+                volumes=["*"],  # This means we apply settings to all the volume zones
```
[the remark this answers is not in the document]
yifan-flex: I think this is part of merging `volumes` and `surfaces` into `models` @benflexcompute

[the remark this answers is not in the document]
benflexcompute: what I do is 
```python
entities : EntityList[SomeType] = pd.Field(alias = "volumes")
```
But I might not have enforced this even myself.. We will need a re-check before release for consistency.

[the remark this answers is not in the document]
benflexcompute: The idea is that it is an entity indeed but we allow alias for user to easily understand what they should put there.

flow360/component/simulation/operating_condition.py
```diff
@@ -4,47 +4,24 @@
...
+    Operating conditions defines the physical (non-geometrical) reference values for the problem.
+    """
 
-OperatingConditionTypes = Union[
-    ExternalFlowOperatingConditions, InternalFlowOperatingConditions, SolidOperatingConditions
-]
+    velocity: Optional[VelocityType] = pd.Field(None)
+    Mach: Optional[float] = pd.Field(None)
```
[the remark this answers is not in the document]
maciej-flexcompute: the primary motivation here is user experience with interface. This is why we introduced operating conditions. They should contain alpha beta angles as well, as this is intended to be the only required part of simulation. Everything else can have defaults. This is problem centric approach. Specifying angles in freestream is solver centric approach. We just wanted to simplify it. To address your questions/points
1. I don't think Mach will be often provided. All our customers think in dimensioned values. We can come up with different input method if that is awkward
2. It is not meant to store info for nondimensionalization because user doesn't care about that. This is solver technical detail. It is side effect of operating conditions, not it's purpose.
3. ok

flow360/component/simulation/operating_condition.py
```diff
@@ -1,50 +1,117 @@
...
+        return cls(
+            density=density,
+            temperature=temperature,
+            material=Air(),
+        )
+
+    @computed_field
+    @property
+    def altitude(self) -> LengthType:
+        return self._altitude
```
benflexcompute, line 37: Have you tried and see if the privateAttr can be serialized by pydantic?

flow360/component/simulation/operating_condition.py
```diff
@@ -1,50 +1,117 @@
...
+        # TODO: implement
+        return 1.825e-5
+
+
+class GenericOperatingCondition(Flow360BaseModel):
+    """
+    Operating condition defines the physical (non-geometrical) reference values for the problem.
+    """
+
+    speed: VelocityType.Positive
```
benflexcompute, line 66: may be call it reference_speed? User might get confused what speed this is?

flow360/component/simulation/operating_condition.py
```diff
@@ -1,50 +1,117 @@
...
+    thermal_condition: ThermalCondition = ThermalCondition()
```
benflexcompute, line 67: Is this a proper way of specifying the default? This looks like you are making `thermal_condition` a class attribute instead.

flow360/component/simulation/operating_condition.py
```diff
@@ -1,50 +1,117 @@
...
+
+    @validate_arguments
+    @classmethod
+    def from_Mach(
+        cls,
+        Mach: PositiveFloat,
+        thermal_condition: ThermalCondition = ThermalCondition(),
+    ):
+        velocity = Mach * self.thermal_condition.speed_of_sound
+        return cls(velocity=velocity, thermal_condition=thermal_condition)
```
benflexcompute, line 77: velocity --> speed?

flow360/component/simulation/operating_condition.py
```diff
@@ -1,50 +1,117 @@
...
+
+    @validate_arguments
+    @classmethod
+    def from_stationary_freestream(
+        cls,
+        reference_speed: VelocityType.Positive,
+        thermal_condition: ThermalCondition = ThermalCondition(),
+    ):
+        pass
+
```
benflexcompute, line 115: we need @property function to retrieve the alpha beta etc that are not stored.

flow360/component/simulation/operating_condition.py
```diff
@@ -1,50 +1,117 @@
...
+        # TODO: implement
+        return 1.013e5
+
+    @property
+    def dynamic_viscosity(self) -> ViscosityType:
+        # TODO: implement
+        return 1.825e-5
+
+
+class GenericOperatingCondition(Flow360BaseModel):
```
benflexcompute, line 61: Maybe call it GenericCondition? as:
operating_condition = GenericCondition(...) to shorten the name? We do not have anything else with name *Condition for now right?

examples/simulation_examples/Simulation_interface_illustration.py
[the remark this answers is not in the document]
benflexcompute: I see that this has not been touched. Let's sync tomorrow we need to be consistent on this.

flow360/component/simulation/operating_condition.py
yifan-flex: That's what @compute_field is for

flow360/component/simulation/operating_condition.py
yifan-flex: OperatingCondition itself is supposed to be reference. I'm not against changing the name, but thermal_condition should be reference_thermal_condition as well

flow360/component/simulation/operating_condition.py
yifan-flex: It's not a class attribute?

flow360/component/simulation/operating_condition.py
yifan-flex: I prefer not making this class name even more obscure. We can discuss

maciej-flexcompute (approved):

PR #279, opened 2024-05-24
[ModelRef] Add Field Definitions for **MeshingParameters**

benflexcompute:


benflexcompute:
We need to determine a balance between global singleton fields and per-entity override.

flow360/component/simulation/meshing_param/params.py
```diff
@@ -18,23 +19,58 @@ class MeshingParameters(Flow360BaseModel):
...
+        ge=0,
+        le=1,
+        description="Narrow gap treatment strength used when two surfaces are in close proximity. Use a value between 0 and 1, where 0 is no treatment and 1 is the most conservative treatment. This parameter has a global impact where the anisotropic transition into the isotropic mesh. However, the impact on regions without close proximity is negligible.",
+    )
+    volume_layer_growth_rate: pd.PositiveFloat = pd.Field(
+        gt=1, description="Growth rate for volume prism layers."
+    )
+    volume_layer_first_layer_thickness: pd.PositiveFloat = pd.Field(
+        description="Default first layer thickness for volumetric anisotropic layers."
+    )
```
maciej-flexcompute, line 57: I think this is not needed. It can be achieved by:
```python
MeshingParameters(refinements=[BoundaryLayerRefinement(entities="*", first_layer_thickness=...,  growth_rate=...)]
```

flow360/component/simulation/meshing_param/params.py
```diff
@@ -18,23 +19,58 @@ class MeshingParameters(Flow360BaseModel):
...
+    max_edge_length: LengthType.PositiveFloat = pd.Field(
+        description="Global maximum edge length for surface cells. This value will be overwritten by the local specification if provided."
+    )  # Note: Applied to faces without faceName and therefore has to exist in root level.
+    curvature_resolution_angle: pd.PositiveFloat = pd.Field(
+        description="""
+        Global maximum angular deviation in degrees. This value will restrict:
+        (1) The angle between a cell’s normal and its underlying surface normal
+        (2) The angle between a line segment’s normal and its underlying curve normal
+        """
+    )
```
maciej-flexcompute, line 69: Similarly, I think this is not needed. It can be achieved by:
```python
MeshingParameters(refinements=[FaceRefinement(entities="*", max_edge_length=...,  curvature_resolution_angle=...)]
```

flow360/component/simulation/meshing_param/params.py
```diff
@@ -18,23 +19,58 @@ class MeshingParameters(Flow360BaseModel):
...
+    surface_layer_growth_rate: float = pd.Field(
+        ge=1, description="Growth rate of the anisotropic layers grown from the edges."
+    )
```
maciej-flexcompute, line 72: similarly, AnisoSurfaceEdge could be used here

flow360/component/simulation/meshing_param/face_params.py
```diff
@@ -2,12 +2,39 @@
...
+
+class FaceRefinement(Flow360BaseModel):
+    """
+    These affects surface meshing.
+    """
+
+    entities: EntityList[Face] = pd.Field(alias="faces")
+    max_edge_length: LengthType.PositiveFloat = pd.Field(
+        description="Local maximum edge length for surface cells."
+    )
```
maciej-flexcompute, line 22: why `Face` not `Surface`?

flow360/component/simulation/meshing_param/params.py
benflexcompute: `entities="*"` will not capture faces that has no name. faceName is not a required property as of now. @rundaji showed me some customer does not always provide faceName so it is not possible to select all faces using `entities`/metadata.

flow360/component/simulation/meshing_param/face_params.py
benflexcompute: Because meshing parameter are applied to faces which may or may not align with boundary surface definition. These are two independent ways of grouping small patches.

flow360/component/simulation/meshing_param/params.py
benflexcompute: Yes, the issue is that:
1. We cannot just assume a value for default. Will the user be happy if he forget one face and that found that face grow with some strange first layer thickness for example?
2. Then We need to make sure the user have this line of code to provide the base line which is not intuitive. How is the user supposed to know he should always begin with:
```python
MeshingParameters(refinements=[BoundaryLayerRefinement(entities="*", first_layer_thickness=...,  growth_rate=...)]
```

flow360/component/simulation/meshing_param/params.py
benflexcompute: I asked Runda and he says specifying `growth_rate` per edge is not possible in Pointwise. The best pointwise can do is to have `surface_layer_growth_rate` per face.

For this one we can try move into the `AnisoSurfaceEdge` because for edges if they have no EdgeName we just ignore them and the mesher runs fine. So it is ok to have no parameter for unnamed edges.

rundaji: The best pointwise can do is to have `surface_layer_growth_rate` per "domain"

flow360/component/simulation/meshing_param/params.py
maciej-flexcompute: You are talking about implementation detail. We can convert "*" to global parameter as we have now. I don't understand why we need to expand "*". It is a global setting and is compatible with our current production mesher. It doesn't matter for the user. What only changes is input format.

I just don't like having 2 things achieving the same goal. The other thing is that we need to name all entities, if something is not named in CSM script, we will just use "unnamed" for it.

flow360/component/simulation/meshing_param/face_params.py
maciej-flexcompute: ok, I see. We can keep it like this now, but I think this distinction is not required when we introduced entities. It was supposed to solve this problem.

benflexcompute: Then the entities corresponds to the patches which are the smallest unit. Problem is we do not have name for patches so the entities wildcard matching does not work.

I think we only have some sort of IDs for the patches (resource: @rundaji). Currently all operations are done either through Face (when you specify meshing parameter) or Surface/Group (when you specify boundary conditions).

flow360/component/simulation/meshing_param/params.py
benflexcompute: Then there is a logic inconsistency right?
If the user do: 
```python
entities = [my_geometry["wing*], "*"]
```
it means wings plus all hidden faces. But if user do:
``python
entities = "*"
```
then it is for every thing?

Maybe an alternative is to ask user skip entities specification. Then we assume it is a global one. This is less logically in conflict with the "*" one.

flow360/component/simulation/meshing_param/params.py
benflexcompute: As discussed, we omit "entities" to specify global properties.

flow360/component/simulation/meshing_param/params.py
benflexcompute: As discussed, we omit "entities" to specify global properties.

flow360/component/simulation/meshing_param/params.py
benflexcompute: Keeping surface_layer_growth_rate as is because it does not make sense to move into a per-face definition.

flow360/component/simulation/meshing_param/face_params.py
benflexcompute: As discussed let's push the unification for surface and face.

maciej-flexcompute (approved):

PR #280, opened 2024-05-24
[ModelRef] Added Field Definitions for **ReferenceGeometry**

benflexcompute:


benflexcompute:
We need the unit system for the test to pass

flow360/component/simulation/primitives.py
```diff
@@ -7,6 +7,18 @@
...
 
+class ReferenceGeometry(Flow360BaseModel):
+    """
+    Contains all geometrical related refrence values
+    Note:
+    - mesh_unit is removed from here and will be a property
+    """
+
+    moment_center: Optional[LengthType.Point] = pd.Field()
+    moment_length: Optional[Union[LengthType.Positive, LengthType.Moment]] = pd.Field()
```
maciej-flexcompute, line 18: do we store float, or (float, float, float) depending what user provided, or we auto convert? The latter would be easier for solver, I think?

benflexcompute: yes, we convert to (float float float) in maybe a after validator.

PR #286, opened 2024-05-31
Add reference_mach to AerospaceCondition.from_mach

yifan-flex:


flow360/component/simulation/operating_condition.py
```diff
@@ -95,6 +95,7 @@ def from_mach(
         alpha: float = 0,
         beta: float = 0,
         atmosphere: ThermalState = ThermalState(),
+        reference_mach: Optional[PositiveFloat] = None,
```
benflexcompute, line 98: Do we still need `Optional` here? But I think we can keep it for now. We will figure out when writing unit tests.

benflexcompute (approved):

PR #288, opened 2024-05-31
Rename volume models: FluidDynamics->Fluid, HeatTransfer->Solid

yifan-flex:
Consideration is that `HeatTransfer` and `FluidDynamics` are not independent concepts and might cause confusion. Discussion is here: https://enreal.slack.com/archives/C01M8RXQPAP/p1717169023990279?thread_ts=1717162776.900719&cid=C01M8RXQPAP

### DIFF
--- a/flow360/component/simulation/models/volume_models.py
+++ b/flow360/component/simulation/models/volume_models.py
@@ -54,7 +54,7 @@ class PDEModelBase(Flow360BaseModel):
     initial_condition: Optional[dict] = pd.Field(None)
 
 
-class FluidDynamics(PDEModelBase):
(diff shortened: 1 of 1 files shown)


benflexcompute:
1. Renamed `HeatTransfer` to `Solid` so that user will not think that entities that are not in `HeatTransfer` does not include heat transfer. We do have heat transfer in FluidDynamics. We name it Solid because it may also contain other parallel physics like  deformation.
2. Renamed FluidDynamics to Fluid so that it is a parallel/same-level concept as `Solid`

benflexcompute (approved):

PR #289, opened 2024-05-31
Rename and separate FluidMaterialTypes and SolidMaterialTypes

yifan-flex:


flow360/component/simulation/models/material.py
```diff
@@ -40,7 +40,7 @@ def prandtl_number(self) -> PositiveFloat:
         return 0.72
 
 
-class Solid(MaterialBase):
+class SolidMaterial(MaterialBase):
```
benflexcompute, line 43: This got changed to not get confused with the Solid model.

flow360/component/simulation/models/material.py
```diff
@@ -63,4 +63,5 @@ def specific_heat_capacity(self) -> SpecificHeatCapacityType.Positive:
         return 903
 
 
-MaterialTypes = Union[Air, Solid, Aluminum]
+SolidMaterialTypes = Union[SolidMaterial, Aluminum]
```
benflexcompute, line 66: User can use SolidMaterial to construct their own solid material

benflexcompute (approved):

PR #290, opened 2024-05-31
Unify TranslationallyPeriodic and RotationallyPeriodic

yifan-flex:
Also removed `paired_patch` since it should be specified in `entities`

### DIFF
--- a/flow360/component/simulation/models/surface_models.py
+++ b/flow360/component/simulation/models/surface_models.py
@@ -92,25 +92,26 @@ class Inflow(BoundaryBaseWithTurbulenceQuantities):
     spec: Union[TotalPressure, MassFlowRate] = pd.Field()
 
 
-class Translationally(Flow360BaseModel):
-    pass
-
-
-class Rotationally(Flow360BaseModel):
-    axis_of_rotation: Optional[Axis] = pd.Field(None)
-
-
-class Periodic(BoundaryBase):
-    type: Literal["Periodic"] = pd.Field("Periodic", frozen=True)
-    spec: Union[Translationally, Rotationally] = pd.Field()
-
-
 class SlipWall(BoundaryBase):
     type: Literal["SlipWall"] = pd.Field("SlipWall", frozen=True)
 
 
 class SymmetryPlane(BoundaryBase):
     type: Literal["SymmetryPlane"] = pd.Field("SymmetryPlane", frozen=True)
+
+
+class Translational(Flow360BaseModel):
+    pass
+
+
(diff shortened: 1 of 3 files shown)


flow360/component/simulation/models/surface_models.py
```diff
@@ -92,17 +92,19 @@ class Inflow(BoundaryBaseWithTurbulenceQuantities):
     spec: Union[TotalPressure, MassFlowRate] = pd.Field()
 
 
-class TranslationallyPeriodic(BoundaryBase):
-    type: Literal["TranslationallyPeriodic"] = pd.Field("TranslationallyPeriodic", frozen=True)
-    paired_patch: Optional[Surface] = pd.Field(None)
+class Translationally(Flow360BaseModel):
```
benflexcompute, line 95: Can we remove "ly" to make it shorter?

flow360/component/simulation/models/surface_models.py
```diff
@@ -111,13 +100,26 @@ class SymmetryPlane(BoundaryBase):
     type: Literal["SymmetryPlane"] = pd.Field("SymmetryPlane", frozen=True)
 
 
+class Translational(Flow360BaseModel):
+    pass
+
+
+class Rotational(Flow360BaseModel):
+    axis_of_rotation: Optional[Axis] = pd.Field(None)
```
benflexcompute, line 108: Is this really optional?

flow360/component/simulation/primitives.py
```diff
@@ -98,4 +98,35 @@ class Surface(_SurfaceEntityBase):
...
+
+    def __eq__(self, other):
+        if isinstance(other, SurfacePair):
+            return tuple(sorted([self.pair[0].name, self.pair[1].name])) == tuple(
+                sorted([other.pair[0].name, other.pair[1].name])
+            )
+        return False
+
+    def __str__(self):
+        return "-".join(sorted([self.pair[0].name, self.pair[1].name]))
```
benflexcompute, line 129: how about instead " ,".join()

benflexcompute: so that it is more consistent with the user input

flow360/component/simulation/models/surface_models.py
yifan-flex: Yes. If user doesn't define we will try to automatically figure out the axis of rotation

benflexcompute: Ok then how does pydantic now which to construct when getting an empty dict? Maybe we need discriminator in this case? We will need to test this.

PR #295, opened 2024-06-03
Fix when support user is None

benflexcompute:


flow360/accounts_utils.py
```diff
@@ -58,6 +58,8 @@ def _get_supported_users():
         try:
             response = http.portal_api_get("auth/credential")
             supported_users = response["guestUsers"]
+            if supported_users is None:
```
lei-flex, line 61: if supported_users:
            	for entry in supported_users:
                    entry["userEmail"] = entry.pop("email")
                    entry["userIdentity"] = entry.pop("identity")
                return supported_users

lei-flex (approved):
fix and ship it.

PR #301, opened 2024-06-05
Add cache model serialization

yifan-flex:


flow360/component/simulation/operating_condition.py
```diff
@@ -1,60 +1,76 @@
...
             material=Air(),
         )
+        state._cached = ThermalStateCache(altitude=altitude, temperature_offset=temperature_offset)
+
+        return state
 
     @property
-    def altitude(self) -> LengthType.Positive:
-        return self._altitude
+    def altitude(self) -> Optional[LengthType.Positive]:
```
benflexcompute, line 51: Is the optional required here?

flow360/component/simulation/operating_condition.py
```diff
@@ -1,60 +1,76 @@
...
+        return self._cached.altitude
```
benflexcompute, line 52: Also would be nice if we tell the user that it was not given by user instead of return none?

flow360/component/simulation/operating_condition.py
```diff
@@ -87,29 +103,41 @@ class AerospaceCondition(Flow360BaseModel):
...
-    @validate_arguments
     @classmethod
+    @pd.validate_call
     def from_mach(
         cls,
-        mach: PositiveFloat,
+        mach: pd.PositiveFloat,
         alpha: float = 0,
         beta: float = 0,
         atmosphere: ThermalState = ThermalState(),
```
benflexcompute, line 113: is atmosphere a proper aerospace term here? I am not familiar

flow360/component/simulation/operating_condition.py
```diff
@@ -87,29 +103,41 @@ class AerospaceCondition(Flow360BaseModel):
...
     reference_velocity_magnitude: Optional[VelocityType.Positive] = None
 
-    @validate_arguments
     @classmethod
+    @pd.validate_call
     def from_mach(
         cls,
-        mach: PositiveFloat,
+        mach: pd.PositiveFloat,
         alpha: float = 0,
```
benflexcompute, line 111: lets use unit for angles. Could you add a TODO here?

flow360/component/simulation/unit_system.py
```diff
@@ -574,22 +574,6 @@ class _TemperatureType(_DimensionedType):
     dim = u.dimensions.temperature
     dim_name = "temperature"
 
-    @classmethod
```
benflexcompute, line 577: Why is this removed?

tests/simulation_framework/test_cache_model.py
```diff
@@ -0,0 +1,71 @@
...
+)
+
+
+class TempThermalStateCache(Flow360BaseModel):
+    altitude: Optional[LengthType.Positive] = None
+    temperature_offset: Optional[TemperatureType] = None
+
+
+class TempThermalState(CacheModelBase):
+    temperature: TemperatureType.Positive = pd.Field(288.15 * u.K, frozen=True)
```
benflexcompute, line 24: Could you add a todo here? We can probably somehow give user warning that the give altitude (if any) will not be valid anymore if the altitude was used and user try to change these fields.

tests/simulation_framework/test_cache_model.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".json", delete=False) as temp_file:
+        temp_file_name = temp_file.name
+        operating_condition.to_file(temp_file_name)
+
+    try:
+        with open(temp_file_name) as fp:
+            model_dict = json.load(fp)
+            assert model_dict["some_value"] == 1230
+            assert model_dict["thermal_state"]["_cached"]
```
benflexcompute, line 67: can you instead of assert here, check the dict is exactly the same as a handwritten one?

benflexcompute: or you can compare the two objects. The original one and the one loaded from JSON. Similar to what we do with SimulationParam seralizer.

flow360/component/simulation/framework/cache_model_base.py
```diff
@@ -0,0 +1,20 @@
+import abc
+from typing import Any, Dict
+
+import pydantic as pd
+
+from flow360.component.simulation.framework.base_model import Flow360BaseModel
+
+
+class CacheModelBase(Flow360BaseModel, metaclass=abc.ABCMeta):
```
benflexcompute, line 9: CachedModelBase

flow360/component/simulation/framework/cache_model_base.py
```diff
@@ -0,0 +1,20 @@
...
+    def __init__(self, **data):
+        cached = data.pop("_cached", None)
+        super().__init__(**data)
+        if cached:
+            self._cached = self.__annotations__["_cached"].model_validate(cached)
```
benflexcompute, line 14: It is going to be interesting when the `model_validate` gives error. Maybe we should catch this internally?

flow360/component/simulation/operating_condition.py
yifan-flex: optional is for returning none

flow360/component/simulation/operating_condition.py
yifan-flex: I think it's ok. If they haven't specified they shouldn't expect a value

flow360/component/simulation/operating_condition.py
benflexcompute: Right. If we want to return none.

flow360/component/simulation/operating_condition.py
yifan-flex: This was Maciej's suggestion in one of the iterations. @maciej-flexcompute I think it makes sense but not 100% sure

flow360/component/simulation/unit_system.py
yifan-flex: We shouldn't have a positive constrains on unit types as we discussed. This can be used for temperature offset for example

flow360/component/simulation/operating_condition.py
benflexcompute: Yes but it would be still better to let them know why it is none.

flow360/component/simulation/unit_system.py
benflexcompute: so did you add .Positive for existing TemperatureType?

flow360/component/simulation/operating_condition.py
benflexcompute: Ok then let's keep this for now.

tests/simulation_framework/test_cache_model.py
benflexcompute: @pytest.mark.usefixtures("array_equality_override")

flow360/component/simulation/operating_condition.py
yifan-flex: https://stackoverflow.com/questions/39429526/how-to-specify-nullable-return-type-with-type-hints
This post suggested using Optional

benflexcompute: Sure

benflexcompute (approved):

PR #310, opened 2024-06-10
Translator Utils

benflexcompute:


flow360/component/simulation/framework/base_model.py
```diff
@@ -104,6 +105,9 @@ def _type(self) -> str:
         conflicting_fields=[],
         include_hash=False,
         include_defaults_in_schema=True,
+        alias_generator=pd.AliasGenerator(
+            serialization_alias=to_camel,
+        ),
```
yifan-flex, line 110: How do I control if alias is used or not?

yifan-flex (approved):

PR #311, opened 2024-06-10
Ben y/volume meshing translator

benflexcompute:


flow360/component/simulation/translator/utils.py
```diff
@@ -44,23 +44,53 @@ def get_simulation_param_dict(input_params: SimulationParams | str | dict, mesh_
...
 
-def get_attribute_from_first_instance(obj_list: list, class_type, attribute_name: str):
+def get_attribute_from_first_instance(
+    obj_list: list, class_type, attribute_name: str, check_empty_entities: bool = False
+):
     """In a list loop and find the first instance matching the given type and retrive the attribute"""
     for obj in obj_list:
         if isinstance(obj, class_type):
+            if check_empty_entities and obj.entities is not None:
+                continue
```
yifan-flex, line 54: What does this do? It will only return the attribute if obj.entities is None?

yifan-flex (approved):

flow360/component/simulation/translator/utils.py
benflexcompute: Yes, for global setting the consensus is to ask user not specify entities at all so we know it is a globale setting. Otherwise it is not convenient to list all the entities.

PR #312, opened 2024-06-10
Add solver translator

yifan-flex:


flow360/component/simulation/framework/base_model.py
```diff
@@ -23,6 +22,17 @@
...
+    components = string.split("_")
+
+    camel_case_string = components[0]
+
+    for component in components[1:]:
+        camel_case_string += component[0].upper() + component[1:]
+
+    return camel_case_string
+
+
```
yifan-flex, line 35: pydantic to_camel doesn't preserve upper cases so I had to write this

flow360/component/simulation/operating_condition.py
```diff
@@ -112,6 +112,11 @@ def dynamic_viscosity(self) -> ViscosityType.Positive:
         # return self.material.speed_of_sound(self.temperature)
         return 1.825e-5 * u.Pa * u.s
 
+    @pd.validate_call
+    def mu_ref(self, mesh_unit: LengthType.Positive) -> pd.PositiveFloat:
+        """Computes nondimensional dynamic viscosity."""
+        return (self.dynamic_viscosity / (self.speed_of_sound * self.density * mesh_unit)).v.item()
```
maciej-flexcompute, line 118: I would not implement this logic here. Unit conversions should be independent from data models. I would prefer we do:
mu_ref = dynamic_viscosity.in_base(unit_system='flow360")
or this kind of interface. 

basically `mu_ref` is of the same type like `dynamic_viscosity` with only difference in unit. Thus I would not create special case and arithmetic for it here. See how it was done before: 
https://github.com/flexcompute/Flow360/blob/4821053f8a1f570f4d0b5884042f73c9e8936efa/flow360/component/flow360_params/flow360_params.py#L643

This `ExtraDimensionedProperty` was introduced for values that di not exist originally in the model. We might not have this issue in operating conditions.

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -0,0 +1,180 @@
...
+
+@preprocess_input
+def get_solver_json(
+    input_params: Union[str, dict, SimulationParams],
+    mesh_unit: LengthType.Positive,
+):
+    """
+    Get the surface mesh json from the simulation parameters.
+    """
+    outputs = input_params.outputs
```
maciej-flexcompute, line 73: if input_params can be "str" or "dict" this line will fail

maciej-flexcompute: oh, I just saw the preprocess code, never mind

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -0,0 +1,180 @@
...
+
+    op = input_params.operating_condition
+    translated["freestream"] = {
+        "alphaAngle": op.alpha.to("degree").v.item(),
+        "betaAngle": op.beta.to("degree").v.item(),
+        "Mach": op.velocity_magnitude.v.item(),
+        "Temperature": op.thermal_state.temperature.to("K").v.item(),
+        "muRef": op.thermal_state.mu_ref(mesh_unit),
+    }
+    if "reference_velocity_magnitude" in op and op.reference_velocity_magnitude:
```
benflexcompute, line 104: This `"reference_velocity_magnitude" in op` works? op is an object and "reference_velocity_magnitude" is string?

flow360/component/simulation/translator/solver_translator.py
benflexcompute, line 1: Would be nice if there is clear separation between each part of solver JSON. This helps maintenance in the future. You can add comment like:
`######### Output ##########`
There are so many things in here.

flow360/component/simulation/operating_condition.py
yifan-flex: The difficulty is really how I access this nondimensional viscosity in the converter. muRef used to be part of what we store in Freestream for flow360_params but that's not the case here. We consider making muRef a private attribute, but that requires special treatment for serialization and deserialization. I'm not sure if it's worth the effort

flow360/component/simulation/translator/solver_translator.py
benflexcompute: Would be even better if we can encapsulate each part into separate file/function like what we did with solver Scripts/preprocessing.

tests/simulation/translator/ref/Flow360_om6Wing.json
benflexcompute, line 1: Does this JSON run om6Wing?

flow360/component/simulation/operating_condition.py
benflexcompute: @yifan-flex muRef I think is not part of user interface. We use `ExtraDimensionedProperty` to compute it actually. Similar to what you did here.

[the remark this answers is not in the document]
yifan-flex: This calculation should still work right? Shouldn't the unit system handle it?

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -0,0 +1,180 @@
...
+    output_dict = to_dict(input_params)
+    for key, value in output_dict.items():
+        output_dict[key] = (
+            value["value"] if isinstance(value, dict) and value.keys() == unit_keys else value
+        )
+    return output_dict
+
+
+def remove_empty_keys(input_dict):
+    # TODO: implement
```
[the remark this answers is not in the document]
yifan-flex: get_attribute_from_first_instance can return None. I don't want to do the check before setting an entry every time, so I thought we can have a function to remove keys for None values when translation is done

flow360/component/simulation/operating_condition.py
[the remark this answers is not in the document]
benflexcompute: I am just not sure if the operation is supported by unyt.

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def to_dict(input_params):
+    return input_params.model_dump(by_alias=True, exclude_none=True)
+
+
+def to_dict_without_unit(input_params):
+    unit_keys = {"value", "units"}
+    output_dict = to_dict(input_params)
+    for key, value in output_dict.items():
```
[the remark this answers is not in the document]
benflexcompute: So if we are not changing this then a better name would be needed. The function name looks like it is going to remove units for all.

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -0,0 +1,180 @@
...
+    return {
+        "animationFrequency": get_attribute_from_first_instance(obj_list, class_type, "frequency"),
+        "animationFrequencyOffset": get_attribute_from_first_instance(
+            obj_list, class_type, "frequency_offset"
+        ),
+        "outputFormat": get_attribute_from_first_instance(obj_list, class_type, "output_format"),
+    }
+
+
+def get_output_fields(obj_list, class_type):
```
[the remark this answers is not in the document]
yifan-flex: I added this before I realized surface and slice won't use it for now, but for prepopulation maybe in the future? There's no harm in keeping a function right

flow360/component/simulation/translator/solver_translator.py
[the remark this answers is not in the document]
benflexcompute: So I think in theory get_attribute_from_first_instance should never return None since we work on preprocessed param. Otherwise we are missing required information.

flow360/component/simulation/translator/solver_translator.py
yifan-flex: Right. I think I meant to write __fields__

flow360/component/simulation/translator/solver_translator.py
yifan-flex: Sure. This function is far from complete. This is just an initial scoping

tests/simulation/translator/ref/Flow360_om6Wing.json
yifan-flex: Yes

flow360/component/simulation/translator/solver_translator.py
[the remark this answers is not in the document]
benflexcompute: What do you mean by prepopulation in the future?

From my understanding this only works for volumeOutput and we do not have many "volumeOutput" so there is not much point moving stuff in to a function that is only going to be used once. 

Maybe I am lazy but I prefer having inline codes rather than jumping around to see what a function actually does (especially if that function is only used once).

Also the name should then be get_output_fields_from_first_instance.

flow360/component/simulation/operating_condition.py
yifan-flex: But a Freestream object is returned in to_solver? How do I do something similar?

flow360/component/simulation/translator/solver_translator.py
benflexcompute: Yes. We can polish this in the fututre

flow360/component/simulation/operating_condition.py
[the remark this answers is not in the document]
yifan-flex: I think it's what unyt is designed to do. Look at the example for adding length in different units
https://unyt.readthedocs.io/en/stable/usage.html#an-example-from-high-school-physics

flow360/component/simulation/translator/solver_translator.py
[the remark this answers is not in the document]
yifan-flex: It's possible for some fields to be optional right?

flow360/component/simulation/translator/solver_translator.py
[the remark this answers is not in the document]
yifan-flex: Maybe we want to use this function to prepopulation output_fields in the future. It's ok I'll remove it

flow360/component/simulation/translator/solver_translator.py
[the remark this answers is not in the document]
benflexcompute: They can be optional but should have been populated by the time they reach translator. But let's see as we go.

flow360/component/simulation/operating_condition.py
benflexcompute: It is possible because Flow360Param accept other models that has mu_ref. I do think we should move away from letting user compute the nondimensional parameter.

benflexcompute: You can try something like: https://github.com/flexcompute/Flow360/blob/a95b0d442eb477e72d8f1763a1aba49794279e52/tests/simulation/translator/test_surface_meshing_translator.py#L42C23-L42C24

yifan-flex: Adding TODO for using unit system for nondimensionalization

PR #318, opened 2024-06-11
Update cached model to store all constructor function arguments

yifan-flex:


flow360/component/simulation/framework/cached_model_base.py
```diff
@@ -1,12 +1,35 @@
...
+        def wrapper(cls, *args, **kwargs):
+            sig = inspect.signature(func)
+            result = func(cls, *args, **kwargs)
+            defaults = {
+                k: v.default
+                for k, v in sig.parameters.items()
+                if v.default is not inspect.Parameter.empty
+            }
+            result._cached = result.__annotations__["_cached"](
+                **{**result._cached.model_dump(), **defaults, **kwargs}
```
benflexcompute, line 26: What happens if **result._cached.model_dump(), **defaults, **kwargs have same keyword args?

benflexcompute (commented):
Will continue review later.

flow360/component/simulation/framework/cached_model_base.py
yifan-flex: cache will be updated. This order should be correct: existing cache < defaults < kwargs

maciej-flexcompute (approved):

PR #328, opened 2024-06-19
Entity Registry in SimulationParams

benflexcompute:
- [x] Add required metadata private attributes to all terminal entities.
- [x] Corrected the merging logic between non-conflicting entities. Generic ones are still lower priority.
- [x] Replaced removing of duplicate entities to merging the entities so there is one type of operation to maintain.
- [x] Removed `supplied_registry` as I expect this will not be how `EntityList` should be used.
- [x] Added control on whether EntityList expansion should be a hardcopy or not. Only when we replace `stored_entities` to expanded entity do we do the hardcopy.
- [x] Fixed problem in `array_equality_override` that it did not reset back the original eq definition.


tests/simulation/framework/test_entities_v2.py
```diff
@@ -29,22 +47,105 @@ class TempVolumeMesh(AssetBase):
...
+                    },
+                },
+                "surfaces": {
+                    "farfield/farfield": {},
+                    "innerZone/interface": {},
+                    "mostinnerZone/interface": {},
+                    "my_wall_1": {},
+                    "my_wall_2": {},
+                    "my_wall_3": {},
+                },
```
yifan-flex, line 126: Reminder to fix this unit test

yifan-flex (approved):

PR #333, opened 2024-06-24
Update input for MassInflow boundary condition 

malhardp333:
* Currently, the `MassInflow` boundary condition requires `totalTemperatureRatio` in the calculations, but it is not set as required input in the Python API. 
* Hence, it has been added as a required input for this boundary condition, and the corresponding tests have been updated as part of this PR.

### DIFF
--- a/flow360/component/flow360_params/boundaries.py
+++ b/flow360/component/flow360_params/boundaries.py
@@ -234,6 +234,7 @@ class MassInflow(BoundaryWithTurbulenceQuantities):
 
     type: Literal["MassInflow"] = pd.Field("MassInflow", const=True)
     mass_flow_rate: PositiveFloat = pd.Field(alias="massFlowRate")
+    total_temperature_ratio: PositiveFloat = pd.Field(alias="totalTemperatureRatio")
     ramp_steps: Optional[PositiveInt] = pd.Field(alias="rampSteps")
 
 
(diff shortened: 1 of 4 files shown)


flow360/component/flow360_params/boundaries.py
```diff
@@ -234,6 +234,7 @@ class MassInflow(BoundaryWithTurbulenceQuantities):
 
     type: Literal["MassInflow"] = pd.Field("MassInflow", const=True)
     mass_flow_rate: PositiveFloat = pd.Field(alias="massFlowRate")
+    total_temperature_ratio: PositiveFloat = pd.Field(alias="totalTemperatureRatio")
```
benflexcompute, line 237: is there a range for this? Or is positive enough?

benflexcompute:
@yifan-flex Let's add a memo to incorporate this into Simulation as well.

flow360/component/flow360_params/boundaries.py
malhardp333: Just positive is enough. In the formula where totalTemperatureRatio gets used, I don't see any other restriction which limits its range.

benflexcompute (commented):
Can you take a look at the unit test:
`tests.test_updater.py::test_updater_from_files`
and add the MassInflow into one of the solver JSON and make sure the totalTermeratureRatio is properly handled by legacy updater and the new model?

yifan-flex:
> @yifan-flex Let's add a memo to incorporate this into Simulation as well.

It has already been incorporated

benflexcompute (approved):

PR #335, opened 2024-06-25
Maciej/workbench service

maciej-flexcompute:


flow360/component/simulation/services.py
```diff
@@ -96,10 +97,14 @@ def get_default_params(unit_system_name, length_unit) -> SimulationParams:
             reference_geometry=ReferenceGeometry(
                 area=1, moment_center=(0, 0, 0), moment_length=(1, 1, 1)
             ),
-            models=[Fluid()],
+            operating_condition=AerospaceCondition(velocity_magnitude=1),
```
benflexcompute, line 100: Personally I am against adding a default for operating condition. We should not try to describe the physical problem for the customer. 
We need to be careful to not accidentally ALWAYS trigger a case execution because we insert too many defaults.

flow360/component/simulation/services.py
```diff
@@ -114,6 +119,8 @@ def validate_model(params_as_dict, unit_system_name):
     validation_warnings = None
     validated_param = None
 
+    params_as_dict = remove_properties_by_name(params_as_dict, "_id")
```
benflexcompute, line 122: _id only comes from the webUI?

flow360/component/simulation/simulation_params.py
```diff
@@ -231,12 +233,41 @@ class SimulationParams(_ParamModelBase):
...
+            v = []
+        assert isinstance(v, list)
+        if not any(isinstance(item, Fluid) for item in v):
+            v.append(Fluid())
+        return v
+
+    @pd.field_validator("outputs", mode="after")
+    @classmethod
+    def apply_defult_output_settings(cls, v):
+        """apply default SurfaceOutput settings if not found in outputs"""
```
benflexcompute, line 267:
```suggestion
        """[Solver Capability Related] apply default SurfaceOutput settings if not found in outputs"""
```

benflexcompute: Adds a tag so we easily identify which are related to solver capability.

flow360/component/simulation/unit_system.py
```diff
@@ -178,6 +178,8 @@ def _unit_object_parser(value, unyt_types: List[type]):
                     return unyt_type(value["value"], value["units"], dtype=np.float64)
                 except u.exceptions.UnitParseError:
                     pass
+                except RuntimeError:
```
benflexcompute, line 181: What is this for?

flow360/component/simulation/services.py
maciej-flexcompute: yes, this is right

flow360/component/simulation/unit_system.py
maciej-flexcompute: when parsing scalar or vector - moment length type, unyt is not throwing error as expected (`u.exceptions.UnitParseError` but RuntimeError) and could infer type correctly.

flow360/component/simulation/services.py
maciej-flexcompute: do you think we should require always: alpha, beta, density, temperature, material? Frontend is not ready and also does not hold any defaults. This way we can carry on default. Note that velocity_magnitude is removed from the dumped model.

benflexcompute: Let's discuss more details later. I do not understand the frontend statement.
We can merge this now. 
The defaults should be consistent with python client. I personally think very high level defaults should be as little as possible but once after the highest level we can provide as much defaults as possible. This allow user to describe the problem with minimal effort but still describe the problem.

benflexcompute (approved):

PR #339, opened 2024-06-27
Add tranlator for porous media

yifan-flex:


flow360/component/simulation/translator/solver_translator.py
```diff
@@ -74,6 +91,70 @@ def rotation_translator(model: Rotation):
...
+    porous_medium.pop("type")
+    porous_medium["DarcyCoefficient"] = list(porous_medium.pop("darcyCoefficient"))
+    porous_medium["ForchheimerCoefficient"] = list(porous_medium.pop("forchheimerCoefficient"))
+    return porous_medium
+
+
+def boundary_translator(model: SurfaceModelTypes, op_acousitc_to_static_pressure_ratio):
+    """Boundary translator"""
+    boundary = remove_units_in_dict(dump_dict(model))
+    boundary.pop("surfaces")
```
benflexcompute, line 126: Is surfaces the alias of entities? If yes let's avoid using aliases anywhere as alias is frequently subject to change. For example here we will not have surface concept in SimulationParams very soon.

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -74,6 +91,70 @@ def rotation_translator(model: Rotation):
...
+    """Porous media translator"""
+    porous_medium = remove_units_in_dict(dump_dict(model))
+    porous_medium.pop("volumes")
+    porous_medium.pop("type")
+    porous_medium["DarcyCoefficient"] = list(porous_medium.pop("darcyCoefficient"))
+    porous_medium["ForchheimerCoefficient"] = list(porous_medium.pop("forchheimerCoefficient"))
+    return porous_medium
+
+
+def boundary_translator(model: SurfaceModelTypes, op_acousitc_to_static_pressure_ratio):
```
benflexcompute, line 123: can you rename it to `boundary_spec_translator`? `boundary_translator `sondes like it does everything including the entity part.

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -74,6 +91,70 @@ def rotation_translator(model: Rotation):
...
+            boundary["type"] = "SubsonicOutflowPressure"
+            boundary["staticPressureRatio"] = spec * op_acousitc_to_static_pressure_ratio
+        elif isinstance(model.spec, Mach):
+            pass
+        elif isinstance(model.spec, MassFlowRate):
+            pass
+    elif isinstance(model, Periodic):
+        pass
+
+    return boundary
```
benflexcompute, line 155: I would actually recommend changing how the translation is done at a high level.

`boundary` here initially contains all the V2.0 model fields which basically means that it inherits from the Simulation boundary model. This introduces coupling which we learned a very painful lesson with the legacy model.

Also, by looking at this function, because it is initially already populated, you do not know what keys exactly the returned boundary will have which is perfect place to hide bugs.

Can you return a newly constructed dict instead and enforce this throughout the rest of translation?

flow360/component/simulation/models/material.py
```diff
@@ -81,9 +82,13 @@ def prandtl_number(self) -> pd.PositiveFloat:
...
-    def speed_of_sound_from_temperature(
-        self, temperature: TemperatureType.Positive
-    ) -> VelocityType.Positive:
+    def pressure(
+        self, density: DensityType.Positive, temperature: TemperatureType.Positive
+    ) -> PressureType.Positive:
+        return density * self.gas_constant * temperature
+
+    @pd.validate_call
+    def speed_of_sound(self, temperature: TemperatureType.Positive) -> VelocityType.Positive:
```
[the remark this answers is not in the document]
yifan-flex: This should belong to material because this is an ideal gas relationship. In the future when we add other type's of gas they will have different implementations

flow360/component/simulation/translator/solver_translator.py
yifan-flex: model_dump writes alias. How do I change that behavior?

benflexcompute (approved):

PR #350, opened 2024-07-08
Flow360Param add `UserDefinedFields`

benflexcompute:


flow360/component/flow360_params/flow360_output.py
```diff
@@ -197,8 +184,11 @@ def to_solver(self, params, **kwargs) -> TimeAverageAnimatedOutput:
 class Surface(Flow360BaseModel):
     """:class:`Surface` class"""
 
-    output_fields: Optional[SurfaceOutputFields] = pd.Field(
-        alias="outputFields", displayed="Output fields", default=[]
+    output_fields: Optional[List[str]] = pd.Field(
+        alias="outputFields",
+        displayed="Output fields",
+        default=[],
+        natively_supported=SurfaceFields,
```
maciej-flexcompute, line 191: I think adding custom fields is not good solution. It will not have type hints and we need custom handling of it. Also, pydantic v2 will not support it:
```
pydantic.warnings.PydanticDeprecatedSince20: Using extra keyword arguments on `Field` is deprecated and will be removed. Use `json_schema_extra` instead. (Extra keys: 'custom_property'). Deprecated in Py...
```

I think better would be better to do:
```
Optional[List[Union[SurfaceFields, str]]]
```

tests/params/test_outputs.py
```diff
@@ -86,11 +87,6 @@ def test_surface_output():
             output_fields=["Cp", "qcriterion"],
         )
 
-    with pytest.raises(pd.ValidationError):
-        output = SurfaceOutput(
-            output_fields=["invalid_field", "qcriterion"],
-        )
-
```
maciej-flexcompute, line 93: why do we need to remove this test?

flow360/component/flow360_params/flow360_fields.py
```diff
@@ -146,8 +146,7 @@ def get_aliases(name) -> List[str]:
     if name in full:
         i = full.index(name)
         return [name, short[i]]
-
-    raise ValueError(f"{name} is not a valid output field name.")
+    return [name, name]
```
maciej-flexcompute, line 149: I think it was better to raise here. I understand the motivation was for unknown field names, whereas this here serves for "known" field names. But I think it should be handled by caller function. Optionally, a parameter can be use `raise_on_not_found=True`

tests/params/test_outputs.py
benflexcompute: The validation is deferred to a root validator so by just instantiating the `SurfaceOutput` obj it is no enough to tell "invalid_field" is invalid. e.g. a user defined field called "invalid_field".

flow360/component/flow360_params/flow360_output.py
benflexcompute: Sure. Just FYI this does not have hints either:
Screenshot 2024-07-09 094808

tests/params/test_outputs.py
benflexcompute: Yes, I added invalid field test.

flow360/component/flow360_params/flow360_fields.py
benflexcompute: Done

flow360/component/flow360_params/flow360_output.py
maciej-flexcompute: oh, I see. Interesting why if doesn't show hints...

maciej-flexcompute (approved):

PR #355, opened 2024-07-11
Fixed lowMach threshold for velocity freestream

NasserFlexCompute:


flow360/component/flow360_params/solvers.py
```diff
@@ -204,7 +204,10 @@ def to_solver(self, params, **kwargs) -> NavierStokesSolver:
 
         if self.low_mach_preconditioner:
             if self.low_mach_preconditioner_threshold is None:
-                self.low_mach_preconditioner_threshold = params.freestream.Mach
+                if hasattr(params.freestream, "Mach"):
+                    self.low_mach_preconditioner_threshold = params.freestream.Mach
+                else:
+                    self.low_mach_preconditioner_threshold = params.freestream.getMach()
```
maciej-flexcompute, line 210: I think enough would be to just do:
```suggestion
                self.low_mach_preconditioner_threshold = params.freestream.to_solver(params, **kwargs).Mach                    
```
and other changes are not required. The `to_solver()` returns object that has Mach calculated

flow360/component/flow360_params/flow360_params.py
benflexcompute, line 1: Can we add a unit test to ensure that the low_mach_preconditioner_threshold is properly set depending on the freestream condition?

flow360/component/flow360_params/solvers.py
maciej-flexcompute: but the issue is that Mach does not exist in all types of Freestream. Only after to_solver() we have access to Mach.

benflexcompute: right we need the value. Ignore my comment then.

PR #356, opened 2024-07-11
added type discriminator to meshing->volume zones

maciej-flexcompute:
the discriminator helps with integration with webUI

### DIFF
--- a/flow360/component/simulation/meshing_param/volume_params.py
+++ b/flow360/component/simulation/meshing_param/volume_params.py
@@ -75,6 +75,7 @@ class RotationCylinder(CylindricalRefinementBase):
 
     """
 
+    type: Literal["RotationCylinder"] = pd.Field("RotationCylinder", frozen=True)
     name: Optional[str] = pd.Field(None)
     entities: EntityList[Cylinder] = pd.Field()
     enclosed_entities: Optional[EntityList[Cylinder, Surface]] = pd.Field(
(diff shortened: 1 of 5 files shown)


flow360/component/simulation/meshing_param/volume_params.py
```diff
@@ -102,10 +103,11 @@ class AutomatedFarfield(Flow360BaseModel):
                     Note: "user-defined" are left out due to scarce usage and will not be implemented.
     """
 
+    type: Literal["AutomatedFarfield"] = pd.Field("AutomatedFarfield", frozen=True)
     name: Optional[str] = pd.Field(None)
     method: Literal["auto", "quasi-3d"] = pd.Field(default="auto", frozen=True)
     private_attribute_entity: GenericVolume = pd.Field(
-        GenericVolume(name="__farfield_zone_name_not_properly_set_yet"), frozen=True
+        GenericVolume(name="__farfield_zone_name_not_properly_set_yet"), frozen=True, exclude=True
```
benflexcompute, line 110: Why do we exclude here?

maciej-flexcompute: what is the reason to dump entities to JSON? Are they intended to be used by some other process? I found all these additional properties confusing for webUI team because they are trying to set them but they seem to be redundant.

benflexcompute (approved):

PR #359, opened 2024-07-12
Remove reference geometry default from translator

yifan-flex:


maciej-flexcompute:
what is the reason for removing default?

benflexcompute:
Related to https://github.com/flexcompute/Flow360Core/pull/1499

benflexcompute (approved):

PR #361, opened 2024-07-12
Add actuator disk translator

yifan-flex:


flow360/component/simulation/translator/solver_translator.py
```diff
@@ -447,6 +447,48 @@ def porous_media_translator(model: PorousMedium):
...
+        "thickness": v["height"],
+    }
+
+
+def bet_disk_translator(model: BETDisk):
+    """BET disk translator"""
+    disk_param = convert_tuples_to_lists(remove_units_in_dict(dump_dict(model)))
+    replace_dict_key(disk_param, "machNumbers", "MachNumbers")
+    replace_dict_key(disk_param, "reynoldsNumbers", "ReynoldsNumbers")
+    disk_param.pop("volumes")
```
benflexcompute, line 466: This introduces coupling between BETDisk class and the translated JSON. It is probably fine because BETDisk is huge but we need good unit test.

benflexcompute (approved):

benflexcompute:
Who is translating AD localtest? Can we add that to unit test before merging?

yifan-flex:
> Who is translating AD localtest? Can we add that to unit test before merging?

We have a unit test although it's not from local tests. AD doesn't have that many entries so this test should be good enough.

flow360/component/simulation/translator/solver_translator.py
yifan-flex: Right, I can decouple them

benflexcompute: Ok then please do that.

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -447,6 +447,60 @@ def porous_media_translator(model: PorousMedium):
...
+        "radius": v["outerRadius"],
+        "thickness": v["height"],
+    }
+
+
+def bet_disk_translator(model: BETDisk):
+    """BET disk translator"""
+    model_dict = convert_tuples_to_lists(remove_units_in_dict(dump_dict(model)))
+    disk_param = {
+        "rotationDirectionRule": model_dict["rotationDirectionRule"],
```
benflexcompute, line 465: Have we ensured that all of these are not Optional?

yifan-flex: They are not

PR #362, opened 2024-07-15
fixed comparing to not None in if statements

maciej-flexcompute:


flow360/component/simulation/translator/solver_translator.py
```diff
@@ -459,15 +459,15 @@ def boundary_spec_translator(model: SurfaceModelTypes, op_acousitc_to_static_pre
                 "velocityType": model.velocity_type,
             }
         )
-        if model.velocity:
```
maciej-flexcompute, line 462: if velocity is array, this if statement is ambiguous

PR #366, opened 2024-07-16
John/remove velocity type

johnmoore4:
removed velocity_type everywhere, because not velocity will always be prescribed in the inertial reference frame.


yifan-flex (approved):
Local tests still pass right?

PR #369, opened 2024-07-16
Added decomposedMesh to legacy

NasserFlexCompute:


flow360/component/flow360_params/flow360_params.py
```diff
@@ -479,7 +479,7 @@ class Geometry(Flow360BaseModel):
     mesh_unit: Optional[LengthType.Positive] = pd.Field(alias="meshUnit")
 
     if Flags.beta_features():
-        decomposed_mesh: Optional[bool] = pd.Field(alias="decomposedMesh", default=False)
+        decomposed_mesh: Optional[bool] = pd.Field(alias="decomposedMesh")
```
benflexcompute, line 482: What is the reason that the default is moved out of the field definition?

benflexcompute (approved):

flow360/component/flow360_params/flow360_params.py
NasserFlexCompute: Just to be consistent with mesh_unit and the rest.

benflexcompute: Ok the GeometryLegacy one has the default in the field definition.

PR #372, opened 2024-07-17
Give warning that `velocityType` is no longer supported.

benflexcompute:
And also allow deseralization to continue instead of erroring out.

Also separated pd.ValidationError handling out of _self_named_property_validator to show more details on why validation failed.

### DIFF
--- a/flow360/component/flow360_params/validations.py
+++ b/flow360/component/flow360_params/validations.py
@@ -25,10 +25,12 @@ from .volume_zones import HeatTransferVolumeZone
 def _ignore_velocity_type_in_boundaries(values):
     """values here is actually json dict."""
     for boundary_name, obj in values.items():
+
         if "velocityType" in obj:
+            bcType = obj.get("type", "")
             log.warning(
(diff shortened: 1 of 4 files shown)


flow360/component/flow360_params/validations.py
```diff
@@ -22,6 +22,19 @@
 from .volume_zones import HeatTransferVolumeZone
 
 
+def _ignore_velocity_type_in_boundaries(values):
+    """values here is actually json dict."""
+    for boundary_name, obj in values.items():
+        if "velocityType" in obj:
+            log.warning(
+                f"In {boundary_name}: 'velocityType' is no longer supported and will be ignored."
```
johnmoore4, line 30: Can you change it to this message to be consistent with the solver:
f"Specifying velocityType for boundary condition {bcType} is no longer supported. The boundary velocity type \
must now always be prescribed relative to to the inertial reference frame."

johnmoore4 (approved):

flow360/component/flow360_params/validations.py
benflexcompute: Done

PR #383, opened 2024-07-18
Enabled DDES sheilding output for 24.7

benflexcompute:


maciej-flexcompute:
no sure if still needed

PR #384, opened 2024-07-19
Added lowDissipationControlFactors

benflexcompute:


flow360/component/flow360_params/solvers.py
```diff
@@ -195,6 +195,9 @@ class NavierStokesSolver(GenericFlowSolverSettings):
     low_mach_preconditioner_threshold: Optional[NonNegativeFloat] = pd.Field(
         alias="lowMachPreconditionerThreshold"
     )
+    low_dissipation_control_factors: Optional[List[float]] = pd.Field(
+        default=[], alias="lowDissipationControlFactors"
+    )
```
yifan-flex, line 200: Is the length 4? Can we add that restriction?

yifan-flex (approved):

flow360/component/flow360_params/solvers.py
benflexcompute: No, arbitrary length

PR #398, opened 2024-07-26
Fix missing value error in __init__ signature

benflexcompute:


tests/simulation/service/test_translator_service.py
```diff
@@ -381,6 +381,7 @@ def test_simulation_to_all_translation():
                     entities=[Surface(name="wing1"), Surface(name="wing2")],
                 ),
                 Freestream(entities=[Surface(name="farfield")]),
+                Rotation(),
```
yifan-flex, line 384: This doesn't throw validation error?

benflexcompute: This is a debug code. Included by mistake

yifan-flex (approved):

PR #415, opened 2024-08-08
Added linear residual output

NasserFlexCompute:


flow360/component/flow360_params/flow360_output.py
```diff
@@ -780,8 +780,11 @@ class LegacyOutputFormat(pd.BaseModel, metaclass=ABCMeta):
     primitive_vars: Optional[bool] = pd.Field(alias="primitiveVars")
     q_criterion: Optional[bool] = pd.Field(alias="qcriterion")
     residual_navier_stokes: Optional[bool] = pd.Field(alias="residualNavierStokes")
+    linear_residual_navier_stokes: Optional[bool] = pd.Field(alias="linearResidualNavierStokes")
```
benflexcompute, line 783: Why do we need to add these for legacy output?

NasserFlexCompute: This is mostly for internal use and thought would be good to be able to add them in old jsons as well. Should we remove them?

benflexcompute: These old JSONs are still using boolean to indicate what fields to output? Like 
`"primitiveVars": True`?

benflexcompute: If they already use "outputFields" then we do not need these

PR #419, opened 2024-08-12
Move non-global meshing settings to global setting.

benflexcompute:
Addressing issues from: https://flow360.atlassian.net/browse/SCFD-2501


benflexcompute:
Change of plan. We now prefer having settings inside the per-item instance.

PR #422, opened 2024-08-13
add trip region to transition solver

miles-flexcompute:


flow360/component/simulation/translator/solver_translator.py
```diff
@@ -616,6 +619,23 @@ def boundary_spec_translator(model: SurfaceModelTypes, op_acousitc_to_static_pre
     return boundary
 
 
+def set_aft_ncrit(ncrit_input, turb_intensity_percent_input):
```
benflexcompute, line 622: Can you instead put this as a after model validator for transitionModelSovler? The translator would be best not to process param too much as it is designed to be disposable. It should only contain logics to "translate",

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -709,6 +737,38 @@ def get_solver_json(
                 translated["turbulenceModelSolver"]["modelConstants"] = translated[
                     "turbulenceModelSolver"
                 ].pop("modelingConstants")
+
+            if not isinstance(model.transition_model_solver, NoneSolver):
+                # baseline dictionary dump for transition model object
+                translated["transitionModelSolver"] = dump_dict(model.transition_model_solver)
+                transition_dict = translated["transitionModelSolver"]
+                replace_dict_key(transition_dict, "typeName", "modelType")
```
benflexcompute, line 745: There are so many duplicated processing in NS and turb solver, in future we should abstract this out. It is fine not to change this in this PR. Just a note

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -709,6 +737,38 @@ def get_solver_json(
...
+                    model.transition_model_solver.turbulence_intensity_percent,
+                )
+                transition_dict.pop("turbulenceIntensityPercent", None)
+                transition_dict.pop("NCrit", None)
+                transition_dict["Ncrit"] = ncrit
+                # build trip region(s) if applicable
+                if "tripRegions" in transition_dict:
+                    transition_dict.pop("tripRegions")
+                    for trip_region in model.transition_model_solver.trip_regions:
+                        trip_region.convert_axis_and_angle_to_coordinate_axes()
```
benflexcompute, line 761: What is the reason we call this again?

flow360/component/simulation/models/solver_numerics.py
```diff
@@ -358,11 +362,40 @@ class TransitionModelSolver(GenericSolverSettings):
     type_name: Literal["AmplificationFactorTransport"] = pd.Field(
         "AmplificationFactorTransport", frozen=True
     )
+    CFL_multiplier: PositiveFloat = pd.Field(2.0)
     absolute_tolerance: PositiveFloat = pd.Field(1e-7)
     equation_evaluation_frequency: PositiveInt = pd.Field(4)
-    turbulence_intensity_percent: pd.confloat(ge=0.03, le=2.5) = pd.Field(1.0)
-    N_crit: pd.confloat(ge=1, le=11) = pd.Field(8.15)
+    turbulence_intensity_percent: Optional[pd.confloat(ge=0.03, le=2.5)] = pd.Field(None)
+    # pylint: disable=C0103
```
benflexcompute, line 369: Can we use the "invalid-name" instead of the code? It provides more readability

flow360/component/simulation/primitives.py
```diff
@@ -252,11 +252,7 @@ def from_principal_axes(
         )
 
     @pd.model_validator(mode="after")
-    def convert_axis_and_angle_to_coordinate_axes(self) -> Self:
-        """
-        Converts the axis and angle transformation represnetation stored in Box
-        to the coordinate axes representation in BoxCache.
-        """
+    def _convert_axis_and_angle_to_coordinate_axes(self) -> Self:
```
benflexcompute, line 255:
```suggestion
    def _convert_axis_and_angle_to_coordinate_axes(self) -> Self:
    """	
    Converts the axis and angle transformation representation stored in Box	
    to the coordinate axes representation and store in Box object.	
    """
```

benflexcompute (approved):

PR #423, opened 2024-08-13
Changed meshing interface design principle to adopt the future proof route

benflexcompute:


flow360/component/simulation/meshing_param/edge_params.py
```diff
@@ -54,12 +47,41 @@ class SurfaceEdgeRefinement(_BaseEdgeRefinement):
...
+                )
+            if self.method is None:
+                raise ValueError("`method` must be specified for per-item surface edge refinement.")
+        else:
+            # Is Global edge refinement
+            if self.growth_rate is None:
+                self.growth_rate = 1.2  # Applying default
+            if self.method is not None:
+                raise ValueError("`method` can be only specified in per-item manner, not global.")
+        return self
```
maciej-flexcompute, line 87: how about we move this validator to `refinements` validator which will be applied on refinements of type SurfaceEdgeRefinement. And then there is only one check we need to do:
check if set(growth rates) is length 1 (need to exclude None). If it is not, we can send warning instead of error that first appearance of growth_rate will be used.

flow360/component/simulation/meshing_param/face_params.py
```diff
@@ -41,15 +42,18 @@ class SurfaceRefinement(Flow360BaseModel):
...
+        if self.entities is not None:
+            # Is per-item refinement
+            if self.curvature_resolution_angle is not None:
+                raise ValueError(
+                    "`curvature_resolution_angle` can be only specified in global manner."
+                )
+        else:
+            # Is Global refinement
+            if self.curvature_resolution_angle is None:
+                self.curvature_resolution_angle = 12 * u.deg  # Applying default
```
maciej-flexcompute, line 56: same strategy here. Only first value will be respected.

flow360/component/simulation/meshing_param/face_params.py
```diff
@@ -70,25 +74,29 @@ class BoundaryLayer(Flow360BaseModel):
...
+            if self.growth_rate is not None:
+                raise ValueError("`growth_rate` can be only specified in global manner.")
+            if self.first_layer_thickness is None:
+                raise ValueError(
+                    "`first_layer_thickness` must be specified for per-item refinement."
+                )
+        else:
+            # Is Global refinement
+            if self.growth_rate is None:
+                self.growth_rate = 1.2  # Applying default
```
maciej-flexcompute, line 99: same here

flow360/component/simulation/meshing_param/edge_params.py
benflexcompute: I see this relieves the requirement a bit but two issues:
1. If we have:
```python
[SurfaceEdgeRefinement(entities=[edge1, edge2], growth_rate = 1.4), SurfaceEdgeRefinement(growth_rate = 1.2)]
```
Then we can not say we use the 1st appearance? Since the eneities-is-none one is supposed to be global but is second in place?
2. Personally I think if user did not specify a global setting it might be arbitrary for us to tell them that we are going to use whatever is the first in the list? The result is going to change depending on the list ordering then.

maciej-flexcompute: ok, so how about we raise error instead of warning and ask user to provide only one value?

benflexcompute: It is on todo list. See https://flow360.atlassian.net/browse/SCFD-2552. There are some values that does not require a global setting. For example max_edge_length.

maciej-flexcompute: right. My problem is that now:
```python
[SurfaceEdgeRefinement(entities=[edge1, edge2], growth_rate = 1.4)]
```
will raise an error, although it is perfectly valid setting. We can use this 1.4 as global, it will affect only these two edges anyway, regardless if local or global.


What I am proposing it to just move this check one level above and make sure that only one value is used. It will be also very easy to remove this check if we decide to support it.

benflexcompute: The global SurfaceEdgeRefinement-> growth_rate does not only affect the listed/mentioned edges. Both meshers that we have will use this global setting for auto-detected edges that does not even exist in the geometry file. So we still require an explicit global `growth_rate`.  Source: @rundaji 

For other global setting fields. We cannot use arbitrary global values for the same reason that they make difference.

benflexcompute: I do appreciate your effort to reduce validation error due to capability limitation.

benflexcompute: I talked with Tengda this morning and the UI will mask most of these to make it harder to error out. See here: https://flow360.atlassian.net/browse/SCFD-2501#:~:text=Change%20of%20plan%3A%20We%20will%20have%20interface%20that%20aims%20to%20be%20future%2Dproof.

flow360/component/simulation/meshing_param/edge_params.py
```diff
@@ -67,6 +68,8 @@ class SurfaceEdgeRefinement(Flow360BaseModel):
         ]
     ] = pd.Field(None, discriminator="type")
 
+    private_attribute_is_global_setting: bool = pd.Field(default=False, frozen=True)
```
benflexcompute, line 71: Why do we need this?
Front end need to know which ones are global settings. Front end use this to detect if it should 1. mask out non-global settings
2. disable "delete", "duplicate" action.
3. disable selecting entities for these global settings.

Why cannot we just see if entities is empty?
Because all user added refinements **initially** does not have `entities` and therefore all are considered global by this criterion. Therefore user cannot add entities to make it non-global...

flow360/component/simulation/meshing_param/edge_params.py
maciej-flexcompute: I see. This is indeed the problem. So this design needs to stay. I goes this is the same problem with `curvature_resolution_angle` that is applied to all surfaces, not only mentioned ones?

How about boundary layer growth rate? This seems only be applied on mentioned patches?

flow360/component/simulation/meshing_param/edge_params.py
```diff
@@ -54,12 +48,50 @@ class SurfaceEdgeRefinement(_BaseEdgeRefinement):
...
+    def _check_valid_setting_combination(self):
+        """Check if the settings are valid in global or per-item context."""
+        if self.entities is not None:
+            # Is per-item edge refinement
+            if self.growth_rate is not None:
+                raise ValueError(
+                    "`growth_rate` can be only specified in global manner, not per edge."
+                )
+            if self.method is None:
+                raise ValueError("`method` must be specified for per-item surface edge refinement.")
```
maciej-flexcompute, line 83:
```suggestion
                raise ValueError(
                    "`growth_rate` can be only specified in global manner, not per edge. Don't use entities (edges) parameter when setting growth rate"
                )
            if self.method is None:
                raise ValueError("`method` must be specified for per-item surface edge refinement. You must provide entities (edges).")
```

benflexcompute:
Overshadowed by https://github.com/flexcompute/Flow360/pull/426

PR #424, opened 2024-08-14
Debug Navier Stokes

NasserFlexCompute:


flow360/component/flow360_params/solvers.py
```diff
@@ -200,6 +203,22 @@ class NavierStokesSolver(GenericFlowSolverSettings):
...
+                "minPressure",
+                "maxVelocity",
+                "maxResCont",
+                "maxResMomX",
+                "maxResMomY",
+                "maxResMomZ",
+                "maxResEnergy",
+            ]
+        ] = pd.Field(alias="debugType")
+        debug_point: Optional[Coordinate] = pd.Field(alias="debugPoint")
```
NasserFlexCompute, line 220: If I want to implement a logic that if debug_type=="point", then debug_point has to be specified, should I do it in Flow360 in in validations under the scripts?

benflexcompute: We should do it in the python client to ensure errors are reported as early as possible

benflexcompute: We also need to ensure when other debug_types are selected there should be no debug_point. I assume these are conflicting?

benflexcompute: I would recommend remove "point" debug_type and use debug_point when specified. If not specified then check debug_type for other options. This way you do not need the `if debug_type=="point", then debug_point has to be specified` logic

benflexcompute (approved):

PR #425, opened 2024-08-15
Added validation for entity properties

benflexcompute:


flow360/component/simulation/unit_system.py
```diff
@@ -552,6 +562,14 @@ def Vector(self):
         """
         return self._VectorType.get_class_object(self)
 
+    # pylint: disable=invalid-name
+    @classproperty
+    def PositiveVector(self):
```
yifan-flex, line 567: Can we have a better name for it? The length is restricted to 3 right?

yifan-flex (approved):

flow360/component/simulation/unit_system.py
benflexcompute: Keeping it as is for now.

PR #430, opened 2024-08-19
Removed  logic of checking entities is None . Changes to use explicit `ForAll` test

benflexcompute:


benflexcompute:
"Converted to DRAFT status since All is not needed for now (we decided to not add the default items under refinements)"

benflexcompute:
Not useful for now.

PR #435, opened 2024-08-20
Added validation to ensure entities to Rotation and PorousMedium are valid

benflexcompute:


tests/simulation/params/test_generic_volume_with_sufficient_properties.py
```diff
@@ -0,0 +1,43 @@
+import pytest
```
yifan-flex, line 1: I prefer that these be separated into test_rotation and test_porous_medium, more unit test should be added to each in the future

yifan-flex (approved):

tests/simulation/params/test_generic_volume_with_sufficient_properties.py
benflexcompute: Done

PR #437, opened 2024-08-21
Get the length unit from project's draft

benflexcompute:


flow360/component/simulation/cloud.py
```diff
@@ -139,6 +141,11 @@ def update_simulation_params(self, params: SimulationParams):
             method="simulation/file",
         )
 
+    def retrieve_simulation_dict(self) -> dict:
+        """retrieve the SimulationParams of the draft"""
+        response = self.get(method="simulation-config")
+        return json.loads(response["simulationJson"])
+
```
maciej-flexcompute, line 148:
```suggestion
    def get_simulation_dict(self) -> dict:
        """retrieve the SimulationParams of the draft"""
        response = self.get(method="simulation-config")
        return json.loads(response["simulationJson"])

```

flow360/component/simulation/cloud.py
```diff
@@ -183,6 +190,18 @@ def _run(
...
         )
 
+    ##-- Getting the project length unit from draft and store in the SimulationParams
+    _resp = RestApi(ProjectInterface.endpoint, id=source_asset.project_id).get()
+    last_opened_draft_id = _resp["lastOpenDraftId"]
+    assert last_opened_draft_id is not None
+    _draft_with_length_unit = Draft.from_cloud(last_opened_draft_id)
+    length_unit = _draft_with_length_unit.retrieve_simulation_dict()[
+        "private_attribute_asset_cache"
+    ]["project_length_unit"]
```
maciej-flexcompute, line 200: I guess we need some error handling here

flow360/component/simulation/cloud.py
```diff
@@ -183,6 +190,18 @@ def _run(
             f"params argument must be a SimulationParams object but is of type {type(params)}"
         )
 
+    ##-- Getting the project length unit from draft and store in the SimulationParams
+    _resp = RestApi(ProjectInterface.endpoint, id=source_asset.project_id).get()
+    last_opened_draft_id = _resp["lastOpenDraftId"]
+    assert last_opened_draft_id is not None
+    _draft_with_length_unit = Draft.from_cloud(last_opened_draft_id)
+    length_unit = _draft_with_length_unit.retrieve_simulation_dict()[
```
maciej-flexcompute, line 198: this adds 3 more http requests. It will significantly slow down overall experience. We need to talk to backend team on how limit number of http requests

flow360/component/simulation/cloud.py
maciej-flexcompute: How user provides this value for the first time, before any project or draft is created?

flow360/component/simulation/cloud.py
benflexcompute: Done

flow360/component/simulation/cloud.py
benflexcompute: When calling `from_file` method user has to specify the length unit. Then it passed to the project API.

benflexcompute: Right. Raised key error if the length unit is not found.

flow360/component/simulation/cloud.py
benflexcompute: right, it would be best to get the length unit from just Geometry get so that all three here is eliminated.

maciej-flexcompute (approved):

PR #438, opened 2024-08-23
Added consistency check for wall function + surface output [SCFD-2205]

andrzej-krupka:


flow360/component/simulation/validation/validation_simulation_params.py
```diff
@@ -0,0 +1,37 @@
...
+    has_wall_function_model = False
+
+    models = v.models
+
+    if models is None:
+        return v
+
+    for model in models:
+        if isinstance(model, Wall) and model.use_wall_function:
+            has_wall_function_model = True
```
benflexcompute, line 20: maybe we can already return here?

benflexcompute (approved):

flow360/component/simulation/validation/validation_simulation_params.py
andrzej-krupka: True! Added a `break` to avoid unnecessary loops. Implemented in `34d5dfe`

PR #440, opened 2024-08-26
Add DDES - volume output consistency validator [SCFD-2206]

andrzej-krupka:
Two points that came into mind while I was implementing this:

1. Output fields named `SpalartAllmaras_DDES` and `kOmegaSST_DDES` were present in the old version of the validator but they didnt exist in the field definition file under `flow360_fields.py`, I added them into `VolumeOutput` names. Seems like this validator didn't work as intended in the previous version, as files with extra fields wouldn't pass validation.

2. We are still using the dual definition of shorthand / full field names with conversion functions between the two. This was originally written by me when we were generating our web UI from autogenerated schemas. We've since moved away from this approach so I think we might want to consider letting the frontend map the full names / localization to shorthand field names in simulation.json and only keeping shorthands in the python client. I think having two massive lists of literal union types, e.g.

```
VolumeFieldNamesFull = Literal[
    CommonFieldNamesFull,
    "BET Metrics",
    "BET Metrics per Disk",
    "linear residual of Navier-Stokes solver",
    "linear residual of turbulence solver",
    "linear residual of transition solver",
    "DDES output for Spalart-Allmaras solver",
    "DDES output for kOmegaSST solver",
]

VolumeFieldNames = Literal[
    CommonFieldNames,
    "betMetrics",
    "betMetricsPerDisk",
    "linearResidualNavierStokes",
    "linearResidualTurbulence",
    "linearResidualTransition",
    "SpalartAllmaras_DDES",
    "kOmegaSST_DDES",
]
```

Is not very elegant from the user side - in particular I don't think anyone will bother typing out the full names when the shorthands exist and are documented - the full description of the field names could perhaps go into docstrings?

Let me know what you guys think. Removing full field names and simplifying `flow360_fields.py` could probably be a separate small task.

@benflexcompute @maciej-flexcompute


flow360/component/simulation/validation/validation_simulation_params.py
```diff
@@ -39,6 +40,46 @@ def _check_consistency_wall_function_and_surface_output(v):
...
+    outputs = v.outputs
+
+    if not outputs:
+        return v
+
+    for output in outputs:
+        if isinstance(output, VolumeOutput) and output.output_fields is not None:
+            output_fields = output.output_fields.items
+            if "SpalartAllmaras_DDES" in output_fields and not (
+                model_type == "SpalartAllmaras" and run_ddes
```
benflexcompute, line 69: `run_ddes` and `model_type` may be referenced without definition when DDES is not activated or when user uses NoneSolver for turb solver.

tests/simulation/params/test_validators_params.py
```diff
@@ -74,6 +93,29 @@ def test_consistency_wall_function_validator(
...
+            models=[fluid_model_with_DDES], outputs=[volume_output_with_SA_DDES]
+        )
+
+    assert params
+
+    message = "kOmegaSST_DDES output can only be specified with kOmegaSST turbulence model and DDES turned on."
+
+    # Invalid simulation params (wrong output type)
+    with SI_unit_system, pytest.raises(ValueError, match=re.escape(message)):
+        _ = SimulationParams(
```
benflexcompute, line 114:
```suggestion
        SimulationParams(
```

benflexcompute: Could you also add a test when DDES is disabled?

benflexcompute:
> Two points that came into mind while I was implementing this:
> 
> 1. Output fields named `SpalartAllmaras_DDES` and `kOmegaSST_DDES` were present in the old version of the validator but they didnt exist in the field definition file under `flow360_fields.py`, I added them into `VolumeOutput` names. Seems like this validator didn't work as intended in the previous version, as files with extra fields wouldn't pass validation.
> 2. We are still using the dual definition of shorthand / full field names with conversion functions between the two. This was originally written by me when we were generating our web UI from autogenerated schemas. We've since moved away from this approach so I think we might want to consider letting the frontend map the full names / localization to shorthand field names in simulation.json and only keeping shorthands in the python client. I think having two massive lists of literal union types, e.g.
> 
> ```
> VolumeFieldNamesFull = Literal[
>     CommonFieldNamesFull,
>     "BET Metrics",
>     "BET Metrics per Disk",
>     "linear residual of Navier-Stokes solver",
>     "linear residual of turbulence solver",
>     "linear residual of transition solver",
>     "DDES output for Spalart-Allmaras solver",
>     "DDES output for kOmegaSST solver",
> ]
> 
> VolumeFieldNames = Literal[
>     CommonFieldNames,
>     "betMetrics",
>     "betMetricsPerDisk",
>     "linearResidualNavierStokes",
>     "linearResidualTurbulence",
>     "linearResidualTransition",
>     "SpalartAllmaras_DDES",
>     "kOmegaSST_DDES",
> ]
> ```
> 
> Is not very elegant from the user side - in particular I don't think anyone will bother typing out the full names when the shorthands exist and are documented - the full description of the field names could perhaps go into docstrings?
> 
> Let me know what you guys think. Removing full field names and simplifying `flow360_fields.py` could probably be a separate small task.
> 
> @benflexcompute @maciej-flexcompute

benflexcompute:
Regarding #2 good point. We should remove the long names. The explanation has to be still exist somewhere serving as documentation but yes we do not want the user to type those.

tests/simulation/params/test_validators_params.py
andrzej-krupka: Added in [0a08e8b](https://github.com/flexcompute/Flow360/pull/440/commits/0a08e8bc2a82aa448bb7c4c1831f355ffe2266f0)

maciej-flexcompute (approved):

PR #442, opened 2024-08-26
[SCFD-2681] [part1] Geometry workflow use entityInfo.json

benflexcompute:
What is done in this PR?

- [x] Geometry class now use entityInfo.json to know what entities are available (to replace geometry.json)
- [x] Geometry class now support `__setitem__` access to its entities (after grouping)
- [x] cloud._run method now injects the parsed entityInfo.json into the submited simulation.json so the 2nd stage validators (still pending in https://github.com/flexcompute/Flow360/pull/439) can use this to know a full picture of the entities.
- [x] Some new helper function / refactor for Geometry class and EntityRegistry class.


flow360/component/resource_base.py
```diff
@@ -112,6 +113,33 @@ def wrapper(obj, *args, **kwargs):
     return wrapper
 
 
+class AssetMetaBaseModelV2(pd_v2.BaseModel):
```
benflexcompute, line 116: Had to go with V2 pydantic since we cannot mix fields of V1 and V2 in same model.

flow360/component/simulation/entity_info.py
```diff
@@ -0,0 +1,142 @@
...
+        description="A full list of attribute names that the user can"
+        "select to achieve grouping of faces. It has same length as `grouped_faces`",
+        alias="faceAttributeNames",
+    )
+    grouped_faces: List[List[Surface]] = pd.Field(
+        [[]],
+        description="The resulting list "
+        "of `Surface` entities after grouping using the attribute name.",
+        alias="groupedFaces",
+    )
```
maciej-flexcompute, line 52: instead of list of lists, can we have better structure? Maybe:
```python
grouped_faces: List[FaceGroup]

class FaceGroup:
    name: str
    items: List[Surface]
```
so we can access this group by name

By the way, we mix all, Face, Surface, Boundary now, for example: `grouped_faces: List[List[Surface]]`. Is is necessary?

flow360/component/simulation/entity_info.py
```diff
@@ -0,0 +1,142 @@
...
+        attribute_name: str,
+        registry: EntityRegistry,
+    ) -> List[Union[Surface, Edge]]:
+        """
+        Group items with given attribute_name.
+        """
+        entity_list = self._get_full_list_of_given_entity_type(attribute_name, entity_type_name)
+        for item in entity_list:
+            registry.register(item)
+        return registry
```
maciej-flexcompute, line 84: it returns `registry`, why is it called `group_items_with_given_tag`? Maybe `group_in_registry`? Also, there is a mix of `tag` and `attribute`.

flow360/component/simulation/entity_info.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def _get_full_list_of_given_entity_type(
```
maciej-flexcompute, line 86: I think name of this function is not good. `full_list` and `given` are redundant.
```suggestion
    def _get_list_of_entities(
```

flow360/component/simulation/entity_info.py
```diff
@@ -0,0 +1,142 @@
...
+    ) -> list:
+        # Validations
+        if entity_type_name is None:
+            raise ValueError("Entity type name is required.")
+        if entity_type_name not in ["face", "edge"]:
+            raise ValueError(
+                f"Invalid entity type name, expected 'face' or 'edge' but got {entity_type_name}."
+            )
+        entity_attribute_names = getattr(self, f"{entity_type_name}_attribute_names")
+        entity_full_list = getattr(self, f"grouped_{entity_type_name}s")
```
maciej-flexcompute, line 99: overall `getattr` is not good solution. It cannot be checked by linter and IDE. If this is larger change, leave a ticket to improve

flow360/component/simulation/framework/entity_registry.py
```diff
@@ -85,7 +85,9 @@ def get_bucket(self, by_type: EntityBase) -> EntityRegistryBucket:
             by_type.model_fields["private_attribute_registry_bucket_name"].default,
         )
 
-    def find_by_naming_pattern(self, pattern: str) -> list[EntityBase]:
+    def find_by_naming_pattern(
+        self, pattern: str, enforce_output_as_list: bool = True, error_when_no_match: bool = False
+    ) -> list[EntityBase]:
         """
         Finds all registered entities whose names match a given pattern.
```
maciej-flexcompute, line 92: what "pattern" formats do we support? Is it glob or regex? Would be good to include some examples

flow360/component/simulation/framework/entity_registry.py
```diff
@@ -105,13 +107,21 @@ def find_by_naming_pattern(self, pattern: str) -> list[EntityBase]:
...
 
-    def find_by_name(self, name: str):
+    def find_single_entity_by_name(self, name: str):
         """Retrieve the entity with the given name from the registry."""
-        entities = self.find_by_naming_pattern(name)
-        if not entities:
-            raise ValueError(f"No entity found in registry with given name: '{name}'.")
+        entities = self.find_by_naming_pattern(
+            name, enforce_output_as_list=True, error_when_no_match=True
+        )
```
maciej-flexcompute, line 124: I think `get_entity(name=)` or `find_by_name(name=)` would be better. Can we ensure no 2 entities with the same name lands in registry? them we don't need this error handling

flow360/component/geometry.py
```diff
@@ -341,35 +342,86 @@ def _show_avaliable_entity_groups(
...
+            log.warning(f"Grouping already exists for {entity_type_name}. Resetting the grouping.")
+            self._reset_grouping(entity_type_name)
+
         self.internal_registry = self._webapi.metadata.group_items_with_given_tag(
-            Surface, attribute_tag=tag_name, registry=self.internal_registry
+            entity_type_name, attribute_name=tag_name, registry=self.internal_registry
         )
+        setattr(self, f"_{entity_type_name}_has_been_grouped", True)
+
+    def group_faces_by_tag(self, tag_name: str) -> None:
```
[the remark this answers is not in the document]
benflexcompute: no difference, which one do we use? Let's unify this.

flow360/component/geometry.py
```diff
@@ -122,7 +123,7 @@ def get_metadata(self, force: bool = False):
             # Windows OS complains when a file is opened in write mode and then read mode. So we need to close it first.
             # pylint: disable=protected-access
             self._download_file(
-                "results/geometry.json",
+                "metadata/entityInfo.json",
```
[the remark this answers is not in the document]
benflexcompute: https://flow360.atlassian.net/browse/SCFD-2717

flow360/component/geometry.py
```diff
@@ -86,27 +87,27 @@ def is_final(self):
...
-class GeometryMeta(AssetMetaBaseModel):
+class GeometryMeta(AssetMetaBaseModelV2):
     """
     GeometryMeta component
     """
 
     project_id: str = pd.Field(alias="projectId")
     deleted: bool = pd.Field()
-    metadata: Optional[_GeometryMetadataModel] = pd.Field(None)
+    metadata: Optional[GeometryEntityInfo] = pd.Field(None)
```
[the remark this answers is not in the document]
benflexcompute: done

flow360/component/geometry.py
```diff
@@ -341,35 +342,86 @@ def _show_avaliable_entity_groups(
             raise Flow360ValueError(
                 f"entity_type_name: {entity_type_name} is invalid. Valid options are: ['faces', 'edges']"
             )
-        group_name_to_items = self._webapi.metadata.process_metadata_for_given_type(
-            entity_type_name
-        )
+
+        self._webapi.get_entity_info()
```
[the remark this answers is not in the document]
benflexcompute: I just wrote a new entity_info() property function under Geometry class to hide the webapi interaction.

flow360/component/geometry.py
```diff
@@ -341,35 +342,86 @@ def _show_avaliable_entity_groups(
...
+        if getattr(self, f"_{entity_type_name}_has_been_grouped", None) is True:
+            # pylint: disable=fixme
+            # TODO: We need to make sure only 1 grouping is used in simluationParams.
+            log.warning(f"Grouping already exists for {entity_type_name}. Resetting the grouping.")
+            self._reset_grouping(entity_type_name)
+
         self.internal_registry = self._webapi.metadata.group_items_with_given_tag(
-            Surface, attribute_tag=tag_name, registry=self.internal_registry
+            entity_type_name, attribute_name=tag_name, registry=self.internal_registry
         )
```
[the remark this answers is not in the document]
benflexcompute: This has been replaced by checking if the face_group_tag/edge_group_tag.

flow360/component/simulation/entity_info.py
```diff
@@ -0,0 +1,142 @@
...
+        ##:: Pydantic kwargs
+        extra="ignore",
+        frozen=False,
+        populate_by_name=True,
+        validate_assignment=True,
+        validate_default=True,
+    )
+
+    @abstractmethod
+    def get_full_list_of_boundaries(self, attribute_name: str = None) -> list:
```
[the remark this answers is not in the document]
benflexcompute: Changed

flow360/component/simulation/entity_info.py
benflexcompute: I prefer separating the tags from their grouping:
1. Front end/ python client can directly know what tag keys are available.
2. Even if we put the name under FaceGroup like this we will need to loop through this list to get the index so there is no improvement in terms of usability.

flow360/component/simulation/entity_info.py
benflexcompute: Changed to group_in_registry

flow360/component/simulation/entity_info.py
benflexcompute: Changed

flow360/component/simulation/entity_info.py
benflexcompute: Removed with a if-else

flow360/component/simulation/framework/entity_registry.py
benflexcompute: glob or regex. We can explain better later.

flow360/component/simulation/framework/entity_registry.py
benflexcompute: It will be hard to prevent the same naming because we may have Surface entity and a Box entity for example have the same name. This is allowed.

maciej-flexcompute (approved):

PR #445, opened 2024-08-27
Meshing translator with grouping

benflexcompute:
https://flow360.atlassian.net/browse/SCFD-2638


tests/simulation/translator/test_surface_meshing_translator.py
```diff
@@ -84,14 +86,106 @@ def _get_meta_data(self):
...
+    def _get_entity_info(self):
+        if self.fname == "om6wing.csm":
+            return GeometryEntityInfo(
+                face_ids=["wing"],
+                edge_ids=[
+                    "wingLeadingEdge",
+                    "wingTrailingEdge",
+                    "rootAirfoilEdge",
+                    "tipAirfoilEdge",
+                ],
```
rundaji, line 98: In real scenario, the face_ids and edge_ids will look like `body01_face0001` and `body01_edge0001`, respectively. I talked to @feilin-flexcompute, since this is just a test/mock/dummy data, we don't have to be extremely rigorous --> but it would be nice if we can add some comments to avoid confusion

feilin-flexcompute (approved):

PR #452, opened 2024-09-06
Fix JSON translation for multiple trip zones

miles-flexcompute:
I was resetting the list for each zone by mistake.


flow360/component/simulation/translator/solver_translator.py
```diff
@@ -742,9 +742,9 @@ def get_solver_json(
                 # build trip region(s) if applicable
                 if "tripRegions" in transition_dict:
                     transition_dict.pop("tripRegions")
```
yifan-flex, line 744: This line is not necessary anymore right?

yifan-flex (approved):

PR #453, opened 2024-09-11
Fixed allowed isosurface fields

maciej-flexcompute:
Fixed allowed isosurface fields for one of "p", "rho", "Mach", "qcriterion", "s", "T", "Cp", "mut", "nuHat"


andrzej-krupka (approved):
LGTM

PR #464, opened 2024-09-19
Add SA and SST Coefficients to Simulation API

miles-flexcompute:


flow360/component/flow360_params/solvers.py
```diff
@@ -312,14 +312,14 @@ class KOmegaSSTModelConstants(Flow360BaseModel):
...
     C_d2: Optional[pd.NonNegativeFloat] = pd.Field(3.0)
-    C_sigma_k1: Optional[pd.PositiveFloat] = pd.Field(0.85)
-    C_sigma_k2: Optional[pd.PositiveFloat] = pd.Field(1.0)
-    C_sigma_omega1: Optional[pd.PositiveFloat] = pd.Field(0.5)
-    C_sigma_omega2: Optional[pd.PositiveFloat] = pd.Field(0.856)
-    C_alpha1: Optional[pd.PositiveFloat] = pd.Field(0.31)
-    C_beta1: Optional[pd.PositiveFloat] = pd.Field(0.075)
-    C_beta2: Optional[pd.PositiveFloat] = pd.Field(0.0828)
-    C_beta_star: Optional[pd.PositiveFloat] = pd.Field(0.09)
+    C_sigma_k1: Optional[pd.NonNegativeFloat] = pd.Field(0.85)
```
NasserFlexCompute, line 315: why this change?

miles-flexcompute: I believe I made a mistake when I added the SST variables the first time. I accidentally made the new SST variables PositiveFloat, while all other model constants in both APIs are NonNegativeFloat. This seems to make sense since you could probably turn on and off terms as a user if they are allowed to be zero (if that's a reasonable thing to want).

NasserFlexCompute (approved):

PR #468, opened 2024-09-23
Improved translation efficiency

benflexcompute:
List of changes:

- [x] Merge the 3 validate_model() calls from 3 translations into 1.
- [x] removed the deepcopy() during unit conversion (Inherited from Flow360Param implementation why do we have this?)
- [x] removed the copy_current() during SimulationParam __init__ (Inherited from Flow360Param implementation why do we have this?)

### DIFF
--- a/flow360/component/simulation/framework/base_model.py
+++ b/flow360/component/simulation/framework/base_model.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import hashlib
 import json
+from itertools import chain
 from typing import Any, List
 
 import pydantic as pd
@@ -539,15 +540,16 @@ class Flow360BaseModel(pd.BaseModel):
         if required_by is None:
             required_by = []
 
+        additional_fields = {}
         if extra is not None:
             for extra_item in extra:
                 # Note: we should not be expecting extra field for SimulationParam?
                 require(extra_item.dependency_list, required_by, params)
-                self_dict[extra_item.name] = extra_item.value_factory()
+                additional_fields[extra_item.name] = extra_item.value_factory()
 
         assert mesh_unit is not None
 
-        for property_name, value in self_dict.items():
+        for property_name, value in chain(self_dict.items(), additional_fields.items()):
             if property_name in [COMMENTS, TYPE_TAG_STR]:
                 continue
             loc_name = property_name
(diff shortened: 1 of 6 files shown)


benflexcompute:
Tested on dev001 and the total translation time (surface mesh + volume mesh + case) is reduced from 1.0 sec to 0.09 sec.

maciej-flexcompute:
I know that deepcopy was used to not modify values in-place. But as soon as tests can pass, I think this is good to remove

benflexcompute:
> I know that deepcopy was used to not modify values in-place. But as soon as tests can pass, I think this is good to remove

There is currently no change in the object so deepcopy should be safe to be removed.

flow360/component/simulation/simulation_params.py
```diff
@@ -111,7 +111,7 @@ def _init_with_context(self, **kwargs):
         Initializes the simulation parameters with the given unit context.
         """
         kwargs = self._init_check_unit_system(**kwargs)
-        super().__init__(unit_system=unit_system_manager.copy_current(), **kwargs)
+        super().__init__(unit_system=unit_system_manager.current, **kwargs)
```
maciej-flexcompute, line 114: are we sure it will not overwrite unit system when two SimulationParams will be created in different unit systems one after the other?

flow360/component/simulation/framework/base_model.py
```diff
@@ -532,7 +531,7 @@ def _convert_dimensions_to_solver(
         extra: List[Any] = None,
     ) -> dict:
         solver_values = {}
-        self_dict = deepcopy(self.__dict__)
+        self_dict = self.__dict__
```
maciej-flexcompute, line 534: this change can be in conflict with line 547 (modifying self). I think we can create classmethod for any extra dimensional fields and remove this logic from here.

benflexcompute: I put these extra field in a separate dict and we loop both of the dicts together.

flow360/component/simulation/simulation_params.py
benflexcompute: Yes, I added a unit test.

maciej-flexcompute (approved):

PR #477, opened 2024-09-25
Add time averaged probe output

yifan-flex:
- [x] Add TimeAverageSurfaceProbe

### DIFF
--- a/flow360/component/simulation/outputs/outputs.py
+++ b/flow360/component/simulation/outputs/outputs.py
@@ -223,6 +223,18 @@ class SurfaceSliceOutput(_AnimationAndFileFormatSettings):
         return _check_unique_probe_type(value, "SurfaceSliceOutput")
 
 
+class TimeAverageProbeOutput(ProbeOutput):
+    """Time average probe monitor output settings."""
+
+    # pylint: disable=abstract-method
+    frequency: int = pd.Field(default=1, ge=1)
+    frequency_offset: int = pd.Field(default=0, ge=0)
+    start_step: Union[pd.NonNegativeInt, Literal[-1]] = pd.Field(
+        default=-1, description="Physical time step to start calculating averaging"
+    )
+    output_type: Literal["TimeAverageProbeOutput"] = pd.Field("TimeAverageProbeOutput", frozen=True)
+
+
 class AeroAcousticOutput(Flow360BaseModel):
     """AeroAcoustic output settings."""
 
@@ -251,6 +263,7 @@ OutputTypes = Annotated[
         ProbeOutput,
         SurfaceProbeOutput,
         SurfaceSliceOutput,
+        TimeAverageProbeOutput,
         AeroAcousticOutput,
     ],
     pd.Field(discriminator="output_type"),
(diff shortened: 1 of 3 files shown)


flow360/component/simulation/outputs/outputs.py
```diff
@@ -201,6 +201,32 @@ def check_unique_probe_type(cls, value):
         return _check_unique_probe_type(value, "SurfaceProbeOutput")
 
 
+class TimeAverageProbeOutput(ProbeOutput):
+    """Time average probe monitor output settings."""
+
+    # pylint: disable=abstract-method
```
benflexcompute, line 207: What abstract-method?

flow360/component/simulation/outputs/outputs.py
```diff
@@ -201,6 +201,32 @@ def check_unique_probe_type(cls, value):
...
+    frequency: int = pd.Field(default=1, ge=1)
+    frequency_offset: int = pd.Field(default=0, ge=0)
+    start_step: Union[pd.NonNegativeInt, Literal[-1]] = pd.Field(
+        default=-1, description="Physical time step to start calculating averaging"
```
benflexcompute, line 211: Is there effective difference in solver between start_step == 0 vs start_step == -1?

tests/simulation/translator/test_output_translation.py
```diff
@@ -403,11 +403,17 @@ def probe_output_config():
         {
             "monitors": {
                 "prb 10": {
+                    "animationFrequency": 1,
```
benflexcompute, line 406: Can you add a test where `TimeAverageProbeOutput` & `TimeAverageSurfaceProbeOutput` are used along with Non-average counterparts? Just to make sure the merge in translator works.

flow360/component/simulation/outputs/outputs.py
yifan-flex: `Method 'load_point_location_from_file' is abstract in class 'ProbeOutput' but is not overridden in child class 'TimeAverageProbeOutput' (abstract-method)`

flow360/component/simulation/outputs/outputs.py
yifan-flex: Yes, if start_step is 0 the first time step won't be written

tests/simulation/translator/test_output_translation.py
yifan-flex: Done

flow360/component/simulation/outputs/outputs.py
benflexcompute: Hmmm. We still want these model to have this functions though. Could you just add implementation which is just throwing not implemented error? This reminds us that we need to add these in the future. Otherwise nobody will remember this.

yifan-flex: Shouldn't the implementation be in the base class? It is not meant to be an abstract method, we have this pylint error only because it isn't implemented yet.

benflexcompute: You are right. Can you add the fake implementation in the base class?

benflexcompute (approved):

PR #482, opened 2024-10-02
added Wall and Freestream into default BCs

maciej-flexcompute:
This PR is for providing default boundary conditions when creating new project:
all geometry is marked as Wall
farfield is marked as Freestream

https://flow360.atlassian.net/browse/SCFD-3032

### DIFF
--- a/flow360/component/simulation/services.py
+++ b/flow360/component/simulation/services.py
@@ -179,8 +179,8 @@ def get_default_params(
                 ],
             )
         # cleaning up stored entities in default settings to let user decide:
-        params.models[0].entities.stored_entities = [] # pylint: disable=unsubscriptable-object
-        params.models[1].entities.stored_entities = [] # pylint: disable=unsubscriptable-object
+        params.models[0].entities.stored_entities = []  # pylint: disable=unsubscriptable-object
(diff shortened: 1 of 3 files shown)


flow360/component/simulation/services.py
```diff
@@ -120,26 +135,28 @@ def get_default_params(
...
-            # pylint: disable=assigning-non-slot,no-member
-            params.private_attribute_asset_cache.project_length_unit = LengthType.validate(
-                length_unit
-            )
     if root_item_type == "Geometry":
+        automated_farfield = AutomatedFarfield(name="Farfield")
+        with unit_system:
+            params = SimulationParams(
+                reference_geometry=reference_geometry,
+                meshing=MeshingParams(
```
andrzej-krupka, line 148: Why initialize unit_system context twice here? In particular, should the farfield be initialized outside this context? This looks like a change compared to how it was before..

maciej-flexcompute: No, `unit_system` context is initialised one time each, there is no nested `unit_system` context. SimulationParams are always in one unit context and stores it. Other objects do not store unit system but only apply units.

andrzej-krupka (approved):
LGTM

PR #489, opened 2024-10-08
[SCFD-2744] [SCFD-2069] Added BC completeness check

benflexcompute:


flow360/component/simulation/validation/validation_context.py
```diff
@@ -185,7 +185,7 @@ def decorator(func: Callable):
         def wrapper(self: Any, *args, **kwargs):
             current_levels = get_validation_levels()
             # Run the validator only if the current levels matches the specified context or is ALL
-            if current_levels is None or any(lvl in (context, ALL) for lvl in current_levels):
+            if current_levels is not None and any(lvl in (context, ALL) for lvl in current_levels):
```
maciej-flexcompute, line 188: that was not mistake. The motivation was the following:
if user is constructing SimulationParams and uses any parameter, we always validate this parameter. It is most likely optional or belongs to optional parent. User is free to remove it to pass validation. It is context irrelevant that any explicit inputs must be valid.

Thus we always run validator when running SimulationParams without validation context. But for developer user, for example web GUI integration or some python client functionalities, we can ignore some errors as we understand why some errors can be ignored. That is the motivation of this logic here.

flow360/component/simulation/validation/validation_simulation_params.py
```diff
@@ -264,3 +268,47 @@ def _validate_cht_has_heat_transfer(params):
...
+            # pylint: disable=protected-access
+            for entity in model.entities._get_expanded_entities(create_hard_copy=False):
+                param_boundary_registry.register(entity)
+        elif hasattr(model, "entity_pairs"):  # Periodic BC
+            for surface_pair in model.entity_pairs.items:
+                param_boundary_registry.register(surface_pair.pair[0])
+                param_boundary_registry.register(surface_pair.pair[1])
+
+    for boundary in asset_boundaries:
+        if param_boundary_registry.contains(boundary) is False:
```
maciej-flexcompute, line 299: `.contain` seems be linear and this is also linear loop, which makes it quadratic complexity. `.contain` at least can be improved to be O(log(N)).

Overall I think we can do 2 checks to improve performance here:
1. check for duplication, O(N)
2. check for array equality, we can use Counter as items are hashable and that would be O(N)

so for most checks we will have O(N) cost

only if the above checks fail, we run O(N*log(N)) check to indicate which boundaries are not set or set multiple times.

flow360/component/simulation/validation/validation_context.py
benflexcompute: But then what is the point of `context_validator` other than providing the context of error? Or is this exactly the purpose?

flow360/component/simulation/validation/validation_simulation_params.py
benflexcompute: I am confused. Can we have a meeting tomorrow?

flow360/component/simulation/validation/validation_context.py
maciej-flexcompute: that is the purpose

flow360/component/simulation/validation/validation_simulation_params.py
benflexcompute: Using set to speed things up. Checking the length of the two sets does not really help since there might be different items in two sets.

flow360/component/simulation/validation/validation_simulation_params.py
```diff
@@ -264,3 +267,55 @@ def _validate_cht_has_heat_transfer(params):
...
 
     return params
+
+
+def _check_complete_boundary_condition_and_unknown_surface(params):
+    ## Step 1: Get all boundaries patches from asset cache
+    asset_boundary_entities = params.private_attribute_asset_cache.boundaries
+    if asset_boundary_entities is None or asset_boundary_entities == []:
+        return params
+    asset_boundaries = {boundary.name for boundary in asset_boundary_entities}
```
maciej-flexcompute, line 277:
```suggestion
    asset_boundaries = set(asset_boundary_entities)
```

flow360/component/simulation/validation/validation_simulation_params.py
```diff
@@ -264,3 +267,55 @@ def _validate_cht_has_heat_transfer(params):
...
+
+    ## Step 2: Collect all used boundaries from the models
+    if len(params.models) == 1 and isinstance(params.models[0], Fluid):
+        return params
```
maciej-flexcompute, line 281: if there are no BC defined at all, why don't we raise?

flow360/component/simulation/validation/validation_simulation_params.py
```diff
@@ -264,3 +267,55 @@ def _validate_cht_has_heat_transfer(params):
...
+            entities = model.entities._get_expanded_entities(create_hard_copy=False)
+        elif hasattr(model, "entity_pairs"):  # Periodic BC
+            entities = [
+                pair for surface_pair in model.entity_pairs.items for pair in surface_pair.pair
+            ]
+
+        for entity in entities:
+            if isinstance(entity, GhostSurface):
+                continue
+            used_boundaries.add(entity.name)
```
maciej-flexcompute, line 301: should we also check that boundary is used multiple times?

flow360/component/simulation/validation/validation_simulation_params.py
```diff
@@ -264,3 +267,55 @@ def _validate_cht_has_heat_transfer(params):
...
+
+    ## Step 3: Use set operations to find missing and unknown boundaries
+    missing_boundaries = asset_boundaries - used_boundaries
+    unknown_boundaries = used_boundaries - asset_boundaries
+
+    if missing_boundaries:
+        missing_list = ", ".join(sorted(missing_boundaries))
+        raise ValueError(
+            f"The following boundaries do not have a boundary condition: {missing_list}. "
+            "Please add them to a model in the `models` section."
```
maciej-flexcompute, line 311:
```suggestion
            "Please add them to a boundary condition model in the `models` section."
```

flow360/component/simulation/validation/validation_simulation_params.py
benflexcompute: because even if user composes a SimulationParams for meshing _only_ this validator will still get triggered. The context validator does not control when validator is called.

flow360/component/simulation/validation/validation_simulation_params.py
benflexcompute: It is done in this validator: https://github.com/flexcompute/Flow360/blob/971359b892f2125fa2ae3d1e32ca416efd242112/flow360/component/simulation/validation/validation_simulation_params.py#L57

flow360/component/simulation/validation/validation_simulation_params.py
benflexcompute: Did not commit this suggestion in the end because `Surface` obj is not hashable

flow360/component/simulation/validation/validation_simulation_params.py
benflexcompute: Changed this validator to adapt with current validation context. We only raise if we have CASE ValidationContext

maciej-flexcompute (approved):

PR #491, opened 2024-10-09
import models from simulation

feilin-flexcompute:


flow360/new_simulation_models.py
```diff
@@ -0,0 +1,125 @@
...
+from flow360.component.simulation.primitives import Edge, ReferenceGeometry, Surface
+from flow360.component.simulation.simulation_params import SimulationParams
+from flow360.component.simulation.time_stepping.time_stepping import (
+    AdaptiveCFL,
+    RampCFL,
+    Steady,
+    Unsteady,
+)
+from flow360.component.simulation.unit_system import SI_unit_system
+from flow360.component.simulation.utils import _model_attribute_unlock
```
benflexcompute, line 70: This is internal used only. We should not expose this.

feilin-flexcompute: removed.

benflexcompute (approved):

PR #493, opened 2024-10-10
 Remove logic for applying global setting when no entity is specified

yifan-flex:


flow360/component/simulation/services.py
```diff
@@ -177,10 +185,18 @@ def get_default_params(
                         name="Freestream", surfaces=[Surface(name="placeholder2")]
                     ),  # to make it consistent with geo
                 ],
+                outputs=[
+                    SurfaceOutput(
+                        name="Surface output 1",
+                        entities=[Surface(name="placeholder")],
+                        output_fields=["Cp", "yPlus", "Cf", "CfVec"],
+                    ),
+                ],
```
maciej-flexcompute, line 194: I think here we want to select "all" as well by default, right? The reason why we don't want to have all for wall, is that this setting will be always incorrect. But of output, I think this is ok. The down side is that we will export some data on freestream surfaces.

If we want all, then we can move this code outside the ifs to avoid duplication

flow360/component/simulation/simulation_params.py
```diff
@@ -229,19 +229,6 @@ def apply_defult_fluid_settings(cls, v):
...
-        """[Solver Capability Related] apply default SurfaceOutput settings if not found in outputs"""
-        if v is None:
-            v = []
-        assert isinstance(v, list)
-        if not any(isinstance(item, SurfaceOutput) for item in v):
-            v.append(
-                SurfaceOutput(name="Surface output 1", output_fields=["Cp", "yPlus", "Cf", "CfVec"])
-            )
-        return v
-
```
maciej-flexcompute, line 244: can you write in the description of PR why this cannot be done this way anymore?

tests/ref/simulation/simulation_json_with_multi_constructor_used.json
```diff
@@ -353,24 +353,6 @@
...
-                    "Cf",
-                    "CfVec"
-                ]
-            },
-            "output_format": "paraview",
-            "output_type": "SurfaceOutput",
-            "write_single_file": false
-        }
-    ],
     "time_stepping": {
```
maciej-flexcompute, line 356: I think this will be a problem when we release python. Currently, user doesn't need to specify surface outputs explicitly. But after this PR, using python user needs to specify it which is different experience than in the past. I would suggest we address this issue with this PR or soon after.

cc: @benflexcompute

benflexcompute: We can add a `SurfaceOutput` instance if user did not specify during `SimulationParam.preprocess()` function. And populate all boundaries.

benflexcompute: The real question is do user rely on these auto-added surface output or not?

flow360/component/simulation/outputs/outputs.py
```diff
@@ -64,7 +64,7 @@ class SurfaceOutput(_AnimationAndFileFormatSettings):
     # pylint: disable=fixme
     # TODO: entities is None --> use all surfaces. This is not implemented yet.
     name: Optional[str] = pd.Field(None)
-    entities: Optional[EntityList[Surface, GhostSurface]] = pd.Field(None, alias="surfaces")
+    entities: EntityList[Surface, GhostSurface] = pd.Field(None, alias="surfaces")
```
benflexcompute, line 67:
```suggestion
    entities: EntityList[Surface, GhostSurface] = pd.Field(alias="surfaces")
```

flow360/component/simulation/outputs/outputs.py
```diff
@@ -129,7 +129,7 @@ class SliceOutput(_AnimationAndFileFormatSettings):
     """Slice output settings."""
 
     name: Optional[str] = pd.Field(None)
-    entities: Optional[EntityList[Slice]] = pd.Field(None, alias="slices")
+    entities: EntityList[Slice] = pd.Field(None, alias="slices")
```
benflexcompute, line 132: Same here. Can you remove all "None" defaults?

tests/ref/simulation/simulation_json_with_multi_constructor_used.json
yifan-flex: I thought we didn't want SurfaceOutput as default for the python client. On webUI the user will have the option to remove this default, but not with python client. We didn't use to always generate SurfaceOutput with v1 python or json.

tests/simulation/translator/ref/Flow360_CHT_three_cylinders.json
```diff
@@ -104,49 +104,19 @@
...
-                    "Cp",
-                    "primitiveVars",
-                    "T",
-                    "heatFlux",
-                    "lowMachPreconditionerSensor"
+                    "T"
                 ]
             },
-            "solid-2/isothermal-2": {
+            "solid-3/Interface_fluid": {
```
benflexcompute, line 117: Why the reference surface name  changed?

tests/simulation/translator/test_output_translation.py
```diff
@@ -216,141 +208,43 @@ def avg_surface_output_config_with_global_setting():
...
-        )
+        param = SimulationParams(outputs=surface_output_config[0] + avg_surface_output_config)
     translated = {"boundaries": {}}
     translated = translate_output(param, translated)
     ref = {
         "animationFrequency": 123,
         "animationFrequencyOffset": 321,
-        "animationFrequencyTimeAverage": 23,
-        "animationFrequencyTimeAverageOffset": 21,
+        "animationFrequencyTimeAverage": -1,
```
benflexcompute, line 229: Why are the reference value changed? Because these were previously "global setting"?

yifan-flex: Yeah, they were previous the "global" values.

tests/simulation/translator/ref/Flow360_CHT_three_cylinders.json
yifan-flex: Just realized these have been the wrong surfaces specified for output. For the CHT case we are only interested in the fluid-solid interfaces.

tests/ref/simulation/simulation_json_with_multi_constructor_used.json
maciej-flexcompute: this is a good point. How do you remove surface output at all? Maybe we can have a custom default constructor for outputs?

benflexcompute (approved):

tests/ref/simulation/simulation_json_with_multi_constructor_used.json
yifan-flex: @maciej-flexcompute For v1 python and json if surfaceOutput is empty we generate surface output for all surfaces with empty fields. After discussing with @benflexcompute we think it's unnecessary to replicate this behavior. The user experience should be consistent since output with empty fields is meaningless.

PR #523, opened 2024-10-29
Add VolumeMesh from local storage and Update module import for local tests

angranl-flex:
- Add all the modules that needs to be imported for localTests in `src/Flow360Core/Flow360/flow360/new_simulation_models.py` file.
- Update AssetBase to allow id=None for running tests locally.

### DIFF
--- a/flow360/component/simulation/web/asset_base.py
+++ b/flow360/component/simulation/web/asset_base.py
@@ -31,7 +31,8 @@ class AssetBase(metaclass=ABCMeta):
     _entity_info: EntityInfoModel = None
 
     # pylint: disable=redefined-builtin
-    def __init__(self, id: str = None):
+    def __init__(self, id: Union[str, None]):
+        """When id is None, the asset is meant to be operating in local mode."""
         # pylint: disable=not-callable
         self.id = id
         self.internal_registry = None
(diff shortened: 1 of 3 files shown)


flow360/component/simulation/web/asset_base.py
```diff
@@ -31,21 +31,23 @@ class AssetBase(metaclass=ABCMeta):
     _entity_info: EntityInfoModel = None
 
     # pylint: disable=redefined-builtin
-    def __init__(self, id: str):
+    def __init__(self, id: str = None):
```
benflexcompute, line 34: Let's not add default for id.

It does not bring any convenience but makes the interface more error-prone.

Also the default/most likely use case should not be when id is None.

```suggestion
    def __init__(self, id: Union[str, None]):
```

benflexcompute: Add in docstring that when id is none the asset is meant to be operating in local mode.

flow360/component/volume_mesh.py
```diff
@@ -1005,6 +1008,44 @@ def from_cloud(cls, id: str) -> VolumeMeshV2:
 
         return asset_obj
 
+    @classmethod
+    def from_local_storage(cls, case_id: str = None, local_storage_path="") -> VolumeMeshV2:
```
benflexcompute, line 1012:
```suggestion
    def from_local_storage(cls, mesh_id: str = None, local_storage_path="") -> VolumeMeshV2:
```

angranl-flex: Updated

flow360/component/simulation/web/asset_base.py
angranl-flex: Updated and also added the docstring

benflexcompute (approved):

PR #526, opened 2024-10-29
Added quick-start and 2d-CRM tutorials

maciej-flexcompute:


flow360/component/simulation/translator/solver_translator.py
```diff
@@ -730,8 +732,7 @@ def get_solver_json(
             if isinstance(op.thermal_state.material.dynamic_viscosity, Sutherland)
             else -1
         ),
-        # pylint: disable=protected-access
-        "muRef": op.thermal_state._mu_ref(mesh_unit),
+        "muRef": op.thermal_state.dynamic_viscosity.v.item(),
```
yifan-flex, line 735: This will actually return the nondimensional dynamic_viscosity even when it's not a field? @benflexcompute and I spent so much time looking for a good solution when we weren't aware that this would work.

maciej-flexcompute: yes, it will. It will call either Sutherlands reference viscosity or Viscosity directly. The units are converted to flow360 in the preprocess stage making sure here they will be flow360_viscosity_units

flow360/component/geometry.py
```diff
@@ -199,7 +199,8 @@ def submit(self, description="", progress_callback=None) -> Geometry:
         heartbeat_thread.join()
         ##:: kick off pipeline
         geometry._webapi._complete_upload()
-        log.info("Geometry successfully submitted.")
+        log.info(f"Geometry successfully submitted: {geometry.short_description()}")
+        self._id = info.id
```
benflexcompute, line 203: is the _id of the GeometryDraft used anywhere?

flow360/component/simulation/web/draft.py
```diff
@@ -79,8 +79,9 @@ def submit(self) -> Draft:
         """
         Submit draft to cloud and under a given project
         """
-        draft_id = RestApi(DraftInterface.endpoint).post(self._request.model_dump(by_alias=True))
-        return Draft.from_cloud(draft_id["id"])
+        draft_meta = RestApi(DraftInterface.endpoint).post(self._request.model_dump(by_alias=True))
+        self._id = draft_meta["id"]
```
benflexcompute, line 83: again is `_id` used?

flow360/examples/tutorial_2dcrm/geometry.csm
benflexcompute, line 1: Should we get this from cloud instead? We did the same thing for the mesh I think?

tests/simulation/translator/test_solver_translator.py
```diff
@@ -80,6 +84,9 @@
     create_XV15HoverMRF_param,
     rotation_cylinder,
 )
+
+assertions = unittest.TestCase("__init__")
```
benflexcompute, line 88: I am not familiar with unnittest. What is this line of code doing?

flow360/component/geometry.py
maciej-flexcompute: yes, it will remove the "remember to submit draft" warning message

flow360/component/simulation/web/draft.py
maciej-flexcompute: same, see here: https://github.com/flexcompute/Flow360/blob/3d28af7ea2221967165e45a06aeb3be37e5e4cf9/flow360/component/resource_base.py#L176

flow360/examples/tutorial_2dcrm/geometry.csm
maciej-flexcompute: I think it is ok to keep it here, it is very small file. The reason for cloud it size of the files

tests/simulation/translator/test_solver_translator.py
maciej-flexcompute: unittest is an alternative framework to pytest. They can be used simultaneously, and unittest gives a bit fancier assertions, for example here: https://github.com/flexcompute/Flow360/blob/a3a98fc0e1968cb37db5682fdb8c0a20b56cd053/tests/simulation/params/test_simulation_params.py#L167

flow360/component/simulation/web/draft.py
benflexcompute, line 83:
```suggestion
        self._id = draft_meta["id"] # silencing submission warning.
```

flow360/component/geometry.py
benflexcompute: Ok. Can we add a comment here to indicate this purpose?

flow360/component/simulation/web/draft.py
maciej-flexcompute: fixed in: https://github.com/flexcompute/Flow360/pull/526/commits/515cf1ea13850341d9b5ee25da3e4d494f3e7248

flow360/component/geometry.py
maciej-flexcompute: fixed in: https://github.com/flexcompute/Flow360/pull/526/commits/515cf1ea13850341d9b5ee25da3e4d494f3e7248

benflexcompute (approved):

PR #528, opened 2024-10-30
[SCFD-2695] Added util function to reset initial condition of Fluid model when forking before handling to user.

benflexcompute:


maciej-flexcompute:
closing because decided to go back to InitialCondition and RestartManipulation approach

PR #529, opened 2024-10-30
Grouping V1 python client components so they are not exposed directly to user.

benflexcompute:


.gitignore
```diff
@@ -312,11 +312,6 @@ $RECYCLE.BIN/
 
 # End of https://www.toptal.com/developers/gitignore/api/python,pycharm+all,windows,macos
 
-
-# mesh files
-*.cgns
-*.ugrid
```
benflexcompute, line 318: This results in moving of the mesh files being ignored by git.

flow360/component/validator.py
maciej-flexcompute, line 1: I think this validator is v1 concept. Do we still run validation service when submitting to release-24.11? I think we shouldn't

flow360/examples/cylinder2D/cylinder.cgns
maciej-flexcompute, line 1: we need to remove these files. The will be fetched from cloud when used

flow360/examples/om6wing/wing_tetra.1.lb8.ugrid
maciej-flexcompute, line 1: we need to remove these files. The will be fetched from cloud when used

tests/v1/_test_cli.py
maciej-flexcompute, line 1: CLI is not v1 concept

tests/v1/_test_http.py
maciej-flexcompute, line 1: http is not v1 concept

flow360/__init__.py
```diff
@@ -1,276 +1,116 @@
...
-from .component.surface_mesh import SurfaceMesh
-from .component.surface_mesh import SurfaceMeshList as MySurfaceMeshes
-from .component.volume_mesh import VolumeMesh
-from .component.volume_mesh import VolumeMeshList as MyVolumeMeshes
-from .environment import Env
-from .flags import Flags
-from .user_config import UserConfig
-from .version import __solver_version__, __version__
-
-__all__ = [
```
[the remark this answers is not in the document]
benflexcompute: fixed

flow360/component/project.py
```diff
@@ -32,6 +31,7 @@
     ProjectAssetCache,
     match_file_pattern,
 )
+from flow360.component.v1 import Case, SurfaceMesh, __solver_version__
```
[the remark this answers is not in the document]
benflexcompute: Case is already in the non-V1 folder. The import is misleading.

Moving `SurfaceMesh` to non-V1.

flow360/component/validator.py
benflexcompute: For Case and SurfaceMesh class yes they still use Validator class. They are not fully 24.11 ready.

flow360/examples/cylinder2D/cylinder.cgns
benflexcompute: Removed

tests/v1/_test_cli.py
benflexcompute: This test actually delete the flow360 configuration in home menu. Seems dangerous to execute. Also it has underscore in file name so it is not executed. Wonder if that is the reason.

benflexcompute: I tried to make the test work but it does not. I think API Key is set before the test runs so it is impossible for the test to change the API key.

Here is what I do:
1. I added _test_cli_tmp.py which contains my effort.
2. keeping _test_cli.py as is.
3. Neither tests is run ( is commented out by file name) to be consistent with previous behavior.

tests/v1/_test_http.py
benflexcompute: I think this test has same problem as the above one. Let's fix later since we need to rush a release.

flow360/examples/om6wing/wing_tetra.1.lb8.ugrid
benflexcompute: Removed

flow360/examples/cylinder2D/cylinder.cgns
maciej-flexcompute: this files seems still be present in this PR

benflexcompute: Deleted in 0b630783eaf1c9791fac4bb46fd9ade6b6f5fe8e

maciej-flexcompute (approved):

PR #538, opened 2024-10-31
Fetching simulation params from cloud

maciej-flexcompute:
- [ ] fetching simulation.json from cloud with Case(...).params interface
- [ ] ported results to pydantic v2


flow360/component/results/case_results.py
```diff
@@ -27,8 +27,10 @@
 )
 from ...exceptions import Flow360ValueError
 from ...log import log
-from ..flow360_params.conversions import unit_converter
+from ..flow360_params.conversions import unit_converter as unit_converter_v1
 from ..flow360_params.flow360_params import Flow360Params
+from ..simulation.conversion import unit_converter as unit_converter_v2
```
benflexcompute, line 32: Does this work? I think both modules (unit_converter_v2,unit_converter_v1) are setting up same-name global variables so it might be that here the v1 unit converter is silently overwritten by v2 ones.

flow360/examples/actuatorDisk/flow360.json
```diff
@@ -25,7 +25,7 @@
       "startAverageIntegrationStep": -1
     },
     "surfaceOutput": {
-      "outputFormat": "both",
+      "outputFormat": "tecplot",
```
benflexcompute, line 28: why are these changed?

benflexcompute (approved):

flow360/examples/actuatorDisk/flow360.json
maciej-flexcompute: when I tried to run this example, it failed saying "both" is not supported

flow360/component/results/case_results.py
maciej-flexcompute: good point, I am not sure. But unit tests would fail I suppose...

PR #541, opened 2024-11-01
Rename `__init__.py` from v1 folder as modules.py

benflexcompute:


tests/test_current_flow360_version.py
```diff
@@ -1,4 +1,4 @@
-from flow360.component.v1 import __version__
+from flow360.v1 import __version__
```
maciej-flexcompute, line 1: why version is v1?

tests/test_environment.py
```diff
@@ -1,4 +1,4 @@
-from flow360.component.v1 import Env
+from flow360.v1 import Env
```
maciej-flexcompute, line 1: why env is v1?

tests/test_utils.py
```diff
@@ -12,9 +12,9 @@
     shared_account_confirm_proceed,
     validate_type,
 )
-from flow360.component.v1 import Accounts
 from flow360.component.volume_mesh import VolumeMeshMeta
 from flow360.exceptions import Flow360TypeError, Flow360ValueError
+from flow360.v1 import Accounts
```
maciej-flexcompute, line 17: why Accounts is v1?

tests/test_current_flow360_version.py
benflexcompute: It will also be in V2

tests/test_environment.py
benflexcompute: It will also be in B2

tests/test_utils.py
benflexcompute: It will also be in V2

tests/test_current_flow360_version.py
benflexcompute: Change to import from absolute path

tests/test_environment.py
benflexcompute: Change to import from absolute path

maciej-flexcompute (approved):

PR #544, opened 2024-11-01
Fix wrong item type for forking a case

benflexcompute:


maciej-flexcompute (approved):

benflexcompute:
TODO: Add test with mock API.

PR #548, opened 2024-11-03
[Tracking] beta-24.11 Python client beta release

benflexcompute:


flow360/component/simulation/models/surface_models.py
```diff
@@ -42,63 +42,94 @@ class BoundaryBase(Flow360BaseModel, metaclass=ABCMeta):
...
+        None,
+        description="The turbulence related quantities definition."
+        + "See :func:`TurbulenceQuantities` documentation.",
+    )
 
 
 class HeatFlux(SingleAttributeModel):
-    """Heat flux"""
+    """
+    :class:`HeatFlux` class to specify the heat flux for `Wall` boundary condition
```
yifan-flex, line 54: Why is there a link in the doc under every class in this file? It just links to the same page.

flow360/component/simulation/models/surface_models.py
```diff
@@ -112,83 +143,140 @@ class Rotational(Flow360BaseModel):
...
 class Wall(BoundaryBase):
-    """Replace Flow360Param:
-    - NoSlipWall
-    - IsothermalWall
-    - HeatFluxWall
-    - WallFunction
-    - SolidIsothermalWall
-    - SolidAdiabaticWall
+    """
+    :class:`Wall` class defines the Wall boundary conditions below based on the input:
```
yifan-flex, line 147: Are we exposing these old boundary condition values to users?

flow360/component/simulation/models/surface_models.py
```diff
@@ -112,83 +143,140 @@ class Rotational(Flow360BaseModel):
...
     """
 
-    name: Optional[str] = pd.Field(None)
+    name: Optional[str] = pd.Field(None, description="Name of the `Periodic` boundary condition.")
     type: Literal["Periodic"] = pd.Field("Periodic", frozen=True)
-    entity_pairs: UniqueItemList[SurfacePair] = pd.Field(alias="surface_pairs")
-    spec: Union[Translational, Rotational] = pd.Field(discriminator="type_name")
+    entity_pairs: UniqueItemList[SurfacePair] = pd.Field(
+        alias="surface_pairs",
+        description=" Patch name of the matching pair of a :class:`~flow360.Surface` patch. "
```
yifan-flex, line 272:
```suggestion
        description="List of matching pairs of :class:`~flow360.Surface`. "
```

flow360/component/simulation/models/surface_models.py
```diff
@@ -112,83 +143,140 @@ class Rotational(Flow360BaseModel):
...
+        + "Needs to be specified for one side of each of the periodic boundary pairs.",
```
yifan-flex, line 273:
```suggestion
```

flow360/component/simulation/models/surface_models.py
benflexcompute: because it looks good with the link style (the underline). No actually because most of the classes are already like this when I reviewed the PR and I thought it is too much work to change.

we can change this if requested.

flow360/component/simulation/models/surface_models.py
benflexcompute: good catch. we should not.

PR #551, opened 2024-11-05
fix(): windows will use \ as path separator, but aws s3 only accept / as path separator

JunchengXue:


flow360/cloud/s3_utils.py
```diff
@@ -393,7 +393,7 @@ def _get_s3_sts_token(self, resource_id: str, file_name: str) -> _S3STSToken:
             return token
         token = _s3_sts_tokens[session_key]
         return token.copy(
-            deep=True, update={"cloud_path": os.path.join(token.cloud_path_prefix, file_name)}
+            deep=True, update={"cloud_path": token.cloud_path_prefix + "/" + file_name}
```
maciej-flexcompute, line 396: what if `filename` is already a path with `\` ?

maciej-flexcompute (approved):

PR #554, opened 2024-11-05
Add from_local_storage function for creating geometry instance with local geometry file

angranl-flex:


flow360/component/geometry.py
```diff
@@ -289,6 +292,44 @@ def show_available_groupings(self, verbose_mode: bool = False):
             show_ids_in_each_group=verbose_mode,
         )
 
+    @classmethod
+    def from_local_storage(cls, mesh_id: str = None, local_storage_path="") -> Geometry:
```
benflexcompute, line 296:
```suggestion
    def from_local_storage(cls, geometry_id: str = None, local_storage_path="") -> Geometry:
```

benflexcompute: Is this function exactly the same as the one in VolumeMeshV2? If so can we consider moving it to the Asset base instead?

benflexcompute: Let's keep duplicate code to minimum.

angranl-flex: OK, I move the function to Asset base now.

flow360/component/geometry.py
```diff
@@ -289,6 +289,26 @@ def show_available_groupings(self, verbose_mode: bool = False):
             show_ids_in_each_group=verbose_mode,
         )
 
+    @classmethod
+    # pylint: disable=redefined-builtin
+    def from_local_storage(cls, id: str = None, local_storage_path="") -> Geometry:
```
benflexcompute, line 294: Pylint here is complaining the `id` being redefined. Can we use something that does not conflict? Like geometry_id?

I know we have other places that did not enforce this... But let's not introduce new ones :)

flow360/component/simulation/web/asset_base.py
```diff
@@ -159,6 +162,39 @@ def from_file(
             length_unit=length_unit,
         )
 
+    @classmethod
+    # pylint: disable=too-many-arguments
+    def from_local_storage(cls, id: str = None, local_storage_path=""):
```
benflexcompute, line 167: Interesting that pylint did not complain here. But still can you change this to asset_id?

benflexcompute (approved):

PR #561, opened 2024-11-06
added support for x slicing force distribution and y slicing force distribution

maciej-flexcompute:
- [x] x slicing force distribution results
- [x] y slicing force distribution results


flow360/component/results/case_results.py
```diff
@@ -42,6 +43,53 @@ def _temp_file_generator(suffix: str = ""):
     return file_path
 
 
+def _find_by_pattern(all_entities: list, pattern):
```
benflexcompute, line 46:
```suggestion
def _find_entities_by_pattern(all_entities: list, pattern):
```

flow360/component/results/case_results.py
```diff
@@ -42,6 +43,53 @@ def _temp_file_generator(suffix: str = ""):
...
+
+    matched_entities = []
```
benflexcompute, line 48: Can we use some other name than entities? `Entities` are usually associated with surface slices etc in context of python client.

flow360/component/results/case_results.py
```diff
@@ -42,6 +43,53 @@ def _temp_file_generator(suffix: str = ""):
...
+    else:
+        regex_pattern = f"^{pattern}$"  # Exact match if no '*'
+
+    regex = re.compile(regex_pattern)
+    # pylint: disable=no-member
+    matched_entities.extend(filter(lambda x: regex.match(x), all_entities))
+    return matched_entities
+
+
+def _filter_headers(
```
benflexcompute, line 60:
```suggestion
def _filter_headers_by_prefix(
```

flow360/component/results/case_results.py
```diff
@@ -362,6 +412,24 @@ def as_dataframe(self):
 
         return pandas.DataFrame(self.values)
 
+    def wait(self, timeout_minutes=60):
```
benflexcompute, line 415: We have many wait functions. I remember we or the tidy 3D team made it a wrapper which can be used. I will add to tickets. We do not have to do it here.

flow360/component/results/case_results.py
```diff
@@ -422,21 +490,24 @@ class MaxResidualLocationResultCSVModel(ResultCSVModel):
     remote_file_name: str = pd.Field(CaseDownloadable.MAX_RESIDUAL_LOCATION.value, frozen=True)
 
 
+class ResultOperations:
+    @classmethod
+    def average_last_fraction(obj: ResultCSVModel, column, avarage_fraction):
+        df = obj.as_dataframe()
+        last_10_percent = df.tail(int(len(df) * avarage_fraction))
```
benflexcompute, line 497:
```suggestion
        selected_fraction = df.tail(int(len(df) * avarage_fraction))
```

flow360/component/results/case_results.py
```diff
@@ -462,26 +533,104 @@ class SurfaceForcesResultCSVModel(ResultCSVModel):
...
+        headers = _filter_headers(self._raw_values.keys(), include_resolved, exclude_resolved)
+        self._values = {
+            key: val for key, val in self.as_dict().items() if key in [*headers, *self._x_columns]
+        }
+        self._sum()
+
+    def _sum(self):
+        for variable in self._variables:
+            new_col_name = "total" + variable
+            self._values.pop(new_col_name, None)
```
benflexcompute, line 603: Is there a way to achieve this with regexpr instead of popping it out?

flow360/component/results/case_results.py
```diff
@@ -462,26 +533,104 @@ class SurfaceForcesResultCSVModel(ResultCSVModel):
...
+            chain.from_iterable(_find_by_pattern(self.entities, exc) for exc in exclude)
+        )
+
+        headers = _filter_headers(self._raw_values.keys(), include_resolved, exclude_resolved)
+        self._values = {
+            key: val for key, val in self.as_dict().items() if key in [*headers, *self._x_columns]
+        }
+        self._sum()
+
+    def _sum(self):
```
benflexcompute, line 600:
```suggestion
    def _filtered_sum(self):
```

flow360/component/results/case_results.py
```diff
@@ -462,26 +533,104 @@ class SurfaceForcesResultCSVModel(ResultCSVModel):
...
+        """
+        if self._entities is None:
+            pattern = re.compile(rf"(.*)_({'|'.join(self._variables)})$")
+            prefixes = {
+                match.group(1) for col in self.as_dict().keys() if (match := pattern.match(col))
+            }
+            self._entities = sorted(prefixes)
+        return self._entities
+
+    def filter(self, include: list = None, exclude: list = None):
```
benflexcompute, line 569: As a side note. What do you think of the CSV file format that we use for now? The CSV seems does not have enough capability to achieve what we want.

For example here properly associating entities with a field ( we have to do hard coded naming patter like entity_name+field_name for now). Does alternative have better data structure to store metadata? Like HDF5?

maciej-flexcompute: yes, I think HDF5 would be better idea. User can convert this data to CSV with column names he prefers to use. I think HDF5 + pandas DataFrame + saving as CSV would be better and more scalable solution. And user interface will be not affected by internal data storage conventions.

flow360/component/results/case_results.py
maciej-flexcompute: addressed in: https://github.com/flexcompute/Flow360/pull/561/commits/7e8543c64dc6fbf027741165109bec18ccd51a6c

flow360/component/results/case_results.py
maciej-flexcompute: addressed in: https://github.com/flexcompute/Flow360/pull/561/commits/7e8543c64dc6fbf027741165109bec18ccd51a6c

flow360/component/results/case_results.py
maciej-flexcompute: addressed in: https://github.com/flexcompute/Flow360/pull/561/commits/7e8543c64dc6fbf027741165109bec18ccd51a6c

flow360/component/results/case_results.py
maciej-flexcompute: addressed in: https://github.com/flexcompute/Flow360/pull/561/commits/7e8543c64dc6fbf027741165109bec18ccd51a6c

PR #572, opened 2024-11-13
added new tutorial cases for airfoils

wiktor-flex:


examples/tutorials/tutorial_2D_30p30n.py
```diff
@@ -0,0 +1,140 @@
...
+                fl.SurfaceRefinement(
+                    name="flap-slat",
+                    max_edge_length=0.55,
+                    faces=[geometry["flap"], geometry["slat"]],
+                ),
+                fl.SurfaceRefinement(
+                    name="trailing",
+                    max_edge_length=0.36,
+                    faces=[
+                        geometry["wingTrailing"],
```
angranl-flex, line 59: For the selection of entities, could you try to use the wildcard/regex more and demonstrate here? You can add a comment after detailing which exactly entites will be selected.

examples/tutorials/tutorial_2D_30p30n.py
```diff
@@ -0,0 +1,140 @@
...
+                        geometry["flapTrailing"],
+                        geometry["slatTrailing"],
+                    ],
+                ),
+                fl.SurfaceEdgeRefinement(
+                    name="edges",
+                    method=fl.HeightBasedRefinement(value=0.0007),
+                    edges=[
+                        geometry["wingtrailingEdge"],
```
angranl-flex, line 68: Same suggestion for the enetity selection here, and please also check the other places.

flow360/examples/tutorial_2D_crm.py
```diff
@@ -1,12 +1,12 @@
...
 from .base_test_case import BaseTestCase
 
 
-class Tutorial_2dcrm(BaseTestCase):
-    name = "tutorial_2dcrm"
+class Tutorial2DCRM(BaseTestCase):
+    name = "tutorial2DCrm"
 
     class url:
         geometry = "local://geometry.csm"
```
angranl-flex, line 12: For the geometry file, can you use the link from the original tutorial to download the input file instead of putting it in the python client repo?

flow360/examples/tutorial_2D_crm.py
angranl-flex, line 1: Regarding the json files (surface_parms.json, volume_params.json, case_params.json), what is the purpose of storing them locally? We only need the geometry file to create the project and then define the simulation params in the python script to run the case, right?

examples/tutorials/tutorial_2D_crm.py
```diff
@@ -84,7 +81,7 @@
         reference_geometry=fl.ReferenceGeometry(
             moment_center=[0.25, 0.005, 0], moment_length=[1, 1, 1], area=0.01
```
angranl-flex, line 82: The moment_center setting seems to be different from the old json file, could you double check all the cases to make sure the parameter settings are consistent?

examples/tutorials/tutorial_2D_30p30n.py
```diff
@@ -0,0 +1,140 @@
...
+                        geometry["slatFrontLEadingEdge"],
+                    ],
+                ),
+                fl.SurfaceEdgeRefinement(
+                    name="symmetry", method=fl.ProjectAnisoSpacing(), edges=[geometry["symmetry"]]
+                ),
+            ],
+        ),
+        reference_geometry=fl.ReferenceGeometry(
+            moment_center=[0.25, 0.005, 0], moment_length=[1, 1, 1], area=0.01
```
angranl-flex, line 82: The moment_center setting seems to be different from the old json file, could you double check all the cases to make sure the parameter settings (such as values, entities and etc.) are consistent?

examples/tutorials/tutorial_2D_30p30n.py
```diff
@@ -0,0 +1,140 @@
...
+                surfaces=[
+                    geometry["wing"],
+                    geometry["flap"],
+                    geometry["slat"],
+                    geometry["wingTrailing"],
+                    geometry["flapTrailing"],
+                    geometry["slatTrailing"],
+                ],
+                name="wall",
+            ),
```
angranl-flex, line 101: The entities for the fl.Wall() are different from the original case json. Could you also double check to ensure consistent? Same for the other two cases.

examples/tutorials/tutorial_2D_crm.py
angranl-flex, line 82: Same comment as in the tutorial_2D_30p30n.py file.

examples/tutorials/tutorial_2D_gaw2.py
```diff
@@ -0,0 +1,135 @@
...
+                        geometry["flapleadingEdge"],
+                    ],
+                ),
+                fl.SurfaceEdgeRefinement(
+                    name="symmetry", method=fl.ProjectAnisoSpacing(), edges=[geometry["symmetry"]]
+                ),
+            ],
+        ),
+        reference_geometry=fl.ReferenceGeometry(
+            moment_center=[0.25, 0.005, 0], moment_length=[1, 1, 1], area=0.01
```
angranl-flex, line 79: Same comment as in the tutorial_2D_30p30n.py file.

examples/tutorials/tutorial_2D_30p30n.py
wiktor-flex: Okay, I will work on that

examples/tutorials/tutorial_2D_30p30n.py
wiktor-flex: 👍

flow360/examples/tutorial_2D_crm.py
wiktor-flex: They are unnecessary and a relic of v1, so they can be deleted

examples/tutorials/tutorial_2D_30p30n.py
wiktor-flex: You are correct, the moment center should be equal to 0, I will check other cases as well

examples/tutorials/tutorial_2D_30p30n.py
wiktor-flex: This is because of the grouping setting that was used, I can change it to a different one where only 3 groupings will exist (wing, flap, slat)

examples/tutorials/tutorial_2D_crm.py
wiktor-flex: Yes

examples/tutorials/tutorial_2D_gaw2.py
wiktor-flex: Yes

examples/tutorials/tutorial_2D_30p30n.py
wiktor-flex: Would you mind going a little more into detail as to what you would expect by using regex here? I am not really sure what the purpose of that would be

flow360/examples/tutorial_2D_crm.py
angranl-flex: I think Maciej should know how to upload those files to aws.

examples/tutorials/tutorial_2D_30p30n.py
angranl-flex: We would like to show the user that we support using name pattern to match group name for multiple geometry groups selection. As a result, we do not need to manually type each group name and have a long script.
For instance, when selecting Trailing faces here, we could use `geometry["*Trailing"]` to select three face groups all together. 
Also, later when selecting trailing edges and leading edges for SurfaceEdgeRefinement, we could use `[geometry["*trailingEdge"], geometry["*(?i)leadingedge"]]` to cover all 6 edge groups.

wiktor-flex: Okay that makes sense now, thanks

angranl-flex (approved):

PR #581, opened 2024-11-17
upload mapbc file if it exists

feilin-flexcompute:
If the mapbc file is in the same directory of the ugrid mesh file and their file name's stem is the same, the mapbc file will be  uploaded.

### DIFF
--- a/flow360/component/utils.py
+++ b/flow360/component/utils.py
@@ -361,9 +361,9 @@ def get_mapbc_from_ugrid(ugrid):
     """
     return associated mapbc file name from the ugrid mesh file
     """
-    mapbc = ugrid.lower().replace(".lb8.ugrid", ".mapbc")
-    mapbc = mapbc.lower().replace(".b8.ugrid", ".mapbc")
-    mapbc = mapbc.lower().replace(".ugrid", ".mapbc")
+    mapbc = ugrid.replace(".lb8.ugrid", ".mapbc")
+    mapbc = mapbc.replace(".b8.ugrid", ".mapbc")
+    mapbc = mapbc.replace(".ugrid", ".mapbc")
     return mapbc
 
 
@@ -536,7 +536,11 @@ class MeshNameParser:
         return self.format in [MeshFileFormat.UGRID, MeshFileFormat.CGNS]
 
     # pylint: disable=missing-function-docstring
-    def get_associated_mapbc_file_name(self):
+    def get_associated_mapbc_file(self):
+        if not self.is_ugrid():
+            raise RuntimeError("Invalid operation to get mapbc file,  since the mesh is not in UGRID format.")
+        mapbc_file = get_mapbc_from_ugrid(self.file_name_no_compression)
+        return mapbc_file
 
     # pylint: disable=missing-function-docstring
     @staticmethod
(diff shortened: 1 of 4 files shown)


flow360/component/volume_mesh.py
```diff
@@ -955,6 +955,14 @@ def submit(self, description="", progress_callback=None, compress=True) -> Volum
                 renamed_file_on_remote, self.file_name, progress_callback=progress_callback
             )
 
+        if mesh_parser.is_ugrid() and os.path.isfile(mesh_parser.get_associated_mapbc_file()):
+            remote_mesh_parser = MeshNameParser(renamed_file_on_remote)
+            volume_mesh._webapi._upload_file(
+                remote_mesh_parser.get_associated_mapbc_file(),
+                mesh_parser.get_associated_mapbc_file(),
+                progress_callback=progress_callback,
+            )
```
benflexcompute, line 964: Can you add logging here to notify user when they submit a ugrid but we cannot find its mapbc file?

flow360/component/utils.py
```diff
@@ -535,6 +535,15 @@ def is_valid_surface_mesh(self):
     def is_valid_volume_mesh(self):
         return self.format in [MeshFileFormat.UGRID, MeshFileFormat.CGNS]
 
+    # pylint: disable=missing-function-docstring
+    def get_associated_mapbc_file(self):
```
benflexcompute, line 539:
```suggestion
    def get_associated_mapbc_filename(self):
```

benflexcompute:
If user uploads mapbc with ugrid, have you tested if we can do volume_mesh["leftWing"] instead of volume_mesh["1"]?

Does the entire workflow runs without error (for example does case run or not? Is there anywhere in casePipeline/solver where we still use the numbers instead of names?)

feilin-flexcompute:
> If user uploads mapbc with ugrid, have you tested if we can do `volume_mesh["leftWing"]` instead of `volume_mesh["1"]`?
> 
> Does the entire workflow runs without error (for example does case run or not? Is there anywhere in casePipeline/solver where we still use the numbers instead of names?)

yes. The admin link attached did exactly what you said.

benflexcompute:
> > If user uploads mapbc with ugrid, have you tested if we can do `volume_mesh["leftWing"]` instead of `volume_mesh["1"]`?
> > Does the entire workflow runs without error (for example does case run or not? Is there anywhere in casePipeline/solver where we still use the numbers instead of names?)
> 
> yes. The admin link attached did exactly what you said.

It only shows the mesh pipelines works. Is there a case run with it?

feilin-flexcompute:
> > > If user uploads mapbc with ugrid, have you tested if we can do `volume_mesh["leftWing"]` instead of `volume_mesh["1"]`?
> > > Does the entire workflow runs without error (for example does case run or not? Is there anywhere in casePipeline/solver where we still use the numbers instead of names?)
> > 
> > 
> > yes. The admin link attached did exactly what you said.
> 
> It only shows the mesh pipelines works. Is there a case run with it?

the case admin link was attached.

flow360/component/utils.py
feilin-flexcompute: fixed

flow360/component/volume_mesh.py
feilin-flexcompute: added

benflexcompute (approved):

PR #584, opened 2024-11-17
misc. improvements

benflexcompute:
- [x] Skipped submission reminder for all V2 workflow paths.
- [x] Better unit system message
- [x] fix(): volume mesh from cloud always expect VolumeMeshEntityInfo which conflict with Geo/SufMesh workflows.
- [x] fix(): bug in regexpr mimicing glob pattern.
- [x] Added id property function for project

### DIFF
--- a/flow360/component/resource_base.py
+++ b/flow360/component/resource_base.py
@@ -5,11 +5,9 @@ Flow360 base Model
 import os
 import re
 import shutil
-import threading
 import time
 import traceback
 from abc import ABCMeta
-from contextlib import contextmanager
 from datetime import datetime
 from enum import Enum
 from functools import wraps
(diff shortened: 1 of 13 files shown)


flow360/component/resource_base.py
```diff
@@ -142,6 +144,27 @@ def handle_none_str(cls, value):
...
+        _thread_local.submission_reminder_suppressed = False
+
+    previous_state = _thread_local.submission_reminder_suppressed
+    _thread_local.submission_reminder_suppressed = True
+    try:
+        yield
+    finally:
+        _thread_local.submission_reminder_suppressed = previous_state
+
+
```
maciej-flexcompute, line 167: maybe we can just remove the reminder instead of this? This python API is supposed to be used for project submissions so reminder is not needed at all

benflexcompute: Sure!

benflexcompute: Removed

maciej-flexcompute (approved):

PR #587, opened 2024-11-18
Add default value for Wall.heat_spec

angranl-flex:
Set the default heat_spec of Wall to heatFlux=0 now

### DIFF
--- a/flow360/component/simulation/models/surface_models.py
+++ b/flow360/component/simulation/models/surface_models.py
@@ -7,6 +7,7 @@ from typing import Literal, Optional, Union
 
 import pydantic as pd
 
+import flow360.component.simulation.units as u
 from flow360.component.simulation.framework.base_model import Flow360BaseModel
 from flow360.component.simulation.framework.entity_base import EntityList
 from flow360.component.simulation.framework.expressions import StringExpression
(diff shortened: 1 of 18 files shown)


flow360/component/simulation/models/surface_models.py
```diff
@@ -165,7 +165,7 @@ class Wall(BoundaryBase):
         None, description="Prescribe a tangential velocity on the wall."
     )
     heat_spec: Optional[Union[HeatFlux, Temperature]] = pd.Field(
-        None,
+        HeatFlux("0"),
```
benflexcompute, line 168: Can we not use expression here? Why do we use expression when we are trying to set a constant value?

angranl-flex: Sure, use a constant value now.

flow360/component/simulation/models/surface_models.py
```diff
@@ -169,8 +170,9 @@ class Wall(BoundaryBase):
     velocity: Optional[VelocityVectorType] = pd.Field(
         None, description="Prescribe a tangential velocity on the wall."
     )
+    # pylint: disable=no-member
     heat_spec: Optional[Union[HeatFlux, Temperature]] = pd.Field(
```
benflexcompute, line 174: Remove Optional

benflexcompute (commented):
Approved but let's wait for front end to make change and then merge.

benflexcompute (approved):

PR #589, opened 2024-11-19
Added support for restart interpolation onto given mesh

benflexcompute:
WIP: Need a simpler example. Current one comes from testcases
Also need to wait for MC to push to preprod (Done)


flow360/component/project.py
```diff
@@ -1312,6 +1315,7 @@ def _run(
             source_item_type=source_item_type,
             solver_version=solver_version if solver_version else self.solver_version,
             fork_case=fork_from is not None,
+            interpolation_volume_mesh_id=interpolate_to_mesh.id if interpolate_to_mesh else None,
```
angranl-flex, line 1318: Not very familiar with the interpolation implementation. Just curious if there is any validation can be performed before running the interpolation? Like if the interpolated mesh is from a very different geometry, we can prevent the case from running? Or we just let the case run and error out if the interpolation failed?

angranl-flex (approved):

flow360/component/project.py
benflexcompute: We just let it run. The validation you proposed is very expensive.

angranl-flex: Oh okay, thanks for the explanation.

PR #595, opened 2024-11-22
changed operating_condition_from_mach_reynolds func to calc density

wiktor-flex:


flow360/component/simulation/operating_condition/operating_condition.py
```diff
@@ -401,21 +404,34 @@ def operating_condition_from_mach_reynolds(
...
-            dynamic_viscosity=Sutherland(
-                reference_temperature=temperature,
-                reference_viscosity=(mach / reynolds) * u.flow360_viscosity_unit,
-                effective_temperature=110.4 * u.K,
-            )
-        ),
+    if temperature == 288.15 * u.K:
+        log.info("Default value of 288.15 K will be used as temperature.")
+
+    material = Air(
```
yifan-flex, line 417: You can just use default Air(). The rest looks good to me.

wiktor-flex: True, and reference_temperature in Sutherland should be a default value instead of temperature variable. Thanks!

flow360/component/simulation/operating_condition/operating_condition.py
```diff
@@ -401,21 +400,28 @@ def operating_condition_from_mach_reynolds(
...
-        temperature=temperature,
-        material=Air(
-            dynamic_viscosity=Sutherland(
-                reference_temperature=temperature,
-                reference_viscosity=(mach / reynolds) * u.flow360_viscosity_unit,
-                effective_temperature=110.4 * u.K,
-            )
-        ),
+    if temperature == 288.15 * u.K:
+        log.info("Default value of 288.15 K will be used as temperature.")
```
yifan-flex, line 411: We should display a similar warning for default length unit.

flow360/component/simulation/operating_condition/operating_condition.py
```diff
@@ -401,21 +400,28 @@ def operating_condition_from_mach_reynolds(
...
+
+    material = Air()
+
+    velocity = mach * material.get_speed_of_sound(temperature)
+
+    density = (
+        reynolds * material.get_dynamic_viscosity(temperature) / (velocity * project_length_unit)
+    )
+
+    thermal_state = ThermalState(temperature=temperature, density=density, material=material)
```
yifan-flex, line 421: Actually default material for ThermalState is Air(), so it doesn't need to be specified.

flow360/component/simulation/operating_condition/operating_condition.py
yifan-flex: Or we should remove the default. Since the length unit appears in two places the default might cause a major error if it is used unintentionally.

flow360/component/simulation/operating_condition/operating_condition.py
wiktor-flex: True

flow360/component/simulation/operating_condition/operating_condition.py
wiktor-flex: You are right, the user should have to specify the length unit, I will change it

yifan-flex (approved):

PR #601, opened 2024-11-26
Fix(): Remove usage of unit_system_name from request

angranl-flex:
The unit system name is obtained from simulation.json now when `WorkbenchValidateRequest` or `WorkbenchGenerateProcessJsonsRequest` received.

### DIFF
--- a/flow360/component/simulation/services.py
+++ b/flow360/component/simulation/services.py
@@ -46,7 +46,10 @@ from flow360.component.simulation.unit_system import (
     imperial_unit_system,
     unit_system_manager,
 )
-from flow360.component.simulation.utils import model_attribute_unlock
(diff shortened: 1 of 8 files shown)


benflexcompute (approved):

flow360/component/simulation/services.py
```diff
@@ -207,7 +210,6 @@ def get_default_params(
 
 def validate_model(
```
benflexcompute, line 211: Lets starting from this function enforce the "too-many-positional-arguments" linting.
```suggestion
def validate_model(*,
```

angranl-flex: OK! I just add asterisk to validate_model and generate_process_json functions, which are both updated to fix the unit_system_name.

PR #603, opened 2024-11-30
feat: switch to use docker cli to generate shutter pdf filel

lei-flex:


flow360/plugins/report/uvf_shutter.py
```diff
@@ -572,56 +574,69 @@ class Shutter(Flow360BaseModel):
...
+                '-v', sequence_file_path + ':/sequence.json',
+                'shutter', 'sequence', 'run', '/sequence.json'
+            ]
+            log.debug(f"execute the docker command: {' '.join(docker_command)}" )
+            try:
+                subprocess.run(docker_command, shell=False, check=True)
+            except Exception as error:
+                log.debug('-----------------------------start docker exception-----------------------------')
+                log.debug(error)
+                log.debug('-----------------------------end docker exception-----------------------------')
```
maciej-flexcompute, line 610: if there are errors, we should do `log.error`

flow360/plugins/report/uvf_shutter.py
```diff
@@ -572,56 +574,69 @@ class Shutter(Flow360BaseModel):
...
+            file_names = [os.path.join(output_folder, entry) for entry in entries if os.path.isfile(os.path.join(output_folder, entry))]
+            return file_names
+
+        tasks = []
+        for case_id, img_folder, uvf_request in screenshots:
+            os.makedirs(img_folder, exist_ok=True)
+            tasks.append(
+                _build_image_from_local(
+                    uvf_request=uvf_request,
+                    output_folder=os.path.join(img_folder,'shutter')
```
maciej-flexcompute, line 625: remove `shutter` from path

flow360/plugins/report/uvf_shutter.py
```diff
@@ -572,56 +574,69 @@ class Shutter(Flow360BaseModel):
...
-            for _, _, shutter_request in screenshots:
-                tasks.append(
-                    _get_image_sequence(
-                        session=session,
-                        url=urljoin(self.url, "/sequence/run"),
-                        shutter_request=shutter_request,
-                    )
-                )
-                log.debug(
-                    f"request to shutter: {json.dumps(shutter_request)}"
```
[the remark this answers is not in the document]
lei-flex: let's do not keep such legacy code in our branch. in case we need it, we could easily get them back from history. and then add the logic back.
otherwise, such code will be left there and probably never be used as dead end zone.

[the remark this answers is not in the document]
maciej-flexcompute: you are removing code which I am using for development. How am I supposed to test report locally? It is not "legacy" code, it is still valid good which is using a different backend.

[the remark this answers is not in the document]
lei-flex: for local test,
1.  I would suggest to config docker and run it.
2. or in you test code,  hijack _get_image_sequence function to use the different backend.
If we don't the code in prod, i would like not keep it in main branch code.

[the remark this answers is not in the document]
MatthewSwords: I'm with @maciej-flexcompute we should leave it. the container still supports http so its usable code. just maybe put a comment on it explain this.

flow360/plugins/report/uvf_shutter.py
```diff
@@ -572,56 +573,79 @@ class Shutter(Flow360BaseModel):
...
+
+        async def _build_image_from_local(uvf_request: list[dict], output_folder: str) -> list[str]:
```
MatthewSwords, line 577: This is pretty big can we break this up.  Also this is async so does that mean we will spin up multiple containers?  Don't we need to set some kind of limit on the number of containers we would spin up?  If that limit already exists I just didn't see it.

flow360/plugins/report/uvf_shutter.py
```diff
@@ -572,56 +573,79 @@ class Shutter(Flow360BaseModel):
...
+            input_folder = os.path.join(output_folder, 'input')
+            os.makedirs(input_folder, exist_ok=True)
+
+            log.info(f"dumping the uvf request {uvf_request} to local folder {input_folder}/sequence.json")
```
MatthewSwords, line 581: Small nit pick but we should call it shutter where ever we can.  We had a whole thing about renaming the repos because they had UVF

flow360/plugins/report/uvf_shutter.py
```diff
@@ -572,56 +573,79 @@ class Shutter(Flow360BaseModel):
...
+            os.makedirs(img_folder, exist_ok=True)
+            tasks.append(
+                _build_image_from_local(
+                    uvf_request=uvf_request,
+                    output_folder=img_folder
                 )
+            )
+
+
+        responses = await asyncio.gather(*tasks, return_exceptions=True)
```
MatthewSwords, line 638: Ohhh is this how your preventing multiple jobs running you just have an await at the end?  If you want it to run async then this will work but you should probably just not have it be an async function then.

MatthewSwords (commented):
Just a few things I had questions on

flow360/plugins/report/uvf_shutter.py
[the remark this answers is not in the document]
lei-flex: we could add it back when we really setup the prod rest api on it. 
but for now, I would like keep the code clean. 
I will create another PR to hold such web api version in right place. 
also such docker cli code need to be moved to pipeline repo. 
most likely, I will define ShutterFetcher with WebShutterFetcher (in Flow360 repo) and CliShutterFetch(in PIpelineRepo) to keep both.

flow360/plugins/report/uvf_shutter.py
lei-flex: responses = await asyncio.gather(*tasks, return_exceptions=True)
@maciej-flexcompute is the person to put it as async call. since he is in vacation. I would like to keep the change no change. we could raise the question back when @maciej-flexcompute merge it back to main branch.

flow360/plugins/report/uvf_shutter.py
lei-flex: @MatthewSwords any suggestion for the appropriate lines for each function? 
the function has around 60 lines including all comments. 
I just get the move files to separated one

flow360/plugins/report/uvf_shutter.py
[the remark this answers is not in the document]
MatthewSwords: it will get lost if we squash the history... don't we usually squash

flow360/plugins/report/uvf_shutter.py
MatthewSwords: this could use some work but if you just want to get it out we can leave it but i would put some todos here that this is needs to get cleaned up. This isn't really async call from what i see

maciej-flexcompute: with rest API calls, it was async meaning every case could be processed in parallel depending how docker instance scales. With CLI, I don't know

maciej-flexcompute:
merged this branch manually with base branch due to many conflicts, this is the merge commit: https://github.com/flexcompute/Flow360/pull/559/commits/109118817556d5f1d96856872264d3742b85a9e0

PR #607, opened 2024-12-03
fixed tutorial cases

wiktor-flex:


examples/tutorials/tutorial_BET_Disk.py
```diff
@@ -96,14 +85,14 @@
...
             thermal_state=fl.ThermalState(
                 temperature=288.15,
                 material=fl.Air(
                     dynamic_viscosity=fl.Sutherland(
                         reference_temperature=288.15,
-                        reference_viscosity=4.29166e-08 * u.flow360_viscosity_unit,
+                        reference_viscosity=4.29166e-08 * fl.u.flow360_viscosity_unit,
                         effective_temperature=110.4,
                     )
                 ),
```
angranl-flex, line 98: We need to update this with the `operating_condition_from_mach_muref` function, right?

wiktor-flex: I used AerospaceCondition.from_mach instead as it is closer to how it should be defined

examples/migration_guide/README.md
```diff
@@ -0,0 +1,28 @@
+# Migration Guide
+
+This folder contains useful tools for helping you adjust your V1 cases to be V2 compatible.
```
angranl-flex, line 3:
```suggestion
This folder contains useful tools to help migrating your old cases to use the latest Flow360 V2 API. 
```

angranl-flex (approved):

PR #609, opened 2024-12-04
new function that creates operating conditions from mach an muref

wiktor-flex:


angranl-flex (approved):
There is a typo in line 278 of test_simulation_params.py: test_mach_reynodls_op_cond -> test_mach_reynolds_op_cond

benflexcompute (changes requested):
I talked with Maciej and the proposal is:
1. Remove all mentioning of the muRef everywhere in documentation since muRef is a Flexcompute thing and it sounds dimensional. Let's use Reynolds throughout instead.
2. We move all the migration functions (Operating condition 2 plus BETDisk 1) into a dedicated file/folder on the python side and we only show them under "migration guide" section of the documentation.
CC: @angranl-flex

examples/migration_guide/README.md
```diff
@@ -25,4 +27,38 @@ In order to save the BET Disk and Cylinder files as jsons, be sure to set `save=
...
+The function takes in the following inputs:
+- mach
+- reynolds
+- project_length_unit
+- temperature
+- alpha
+- beta
+- reference_mach
+
+and returns an `AerospaceCondition` class instatiated based on given parameters.
```
benflexcompute, line 45:
```suggestion
and returns an `AerospaceCondition` class instantiated based on given parameters.
```

examples/migration_guide/README.md
```diff
@@ -25,4 +27,38 @@ In order to save the BET Disk and Cylinder files as jsons, be sure to set `save=
...
+The function takes in the following inputs:
+- mach
+- muRef
+- project_length_unit
+- temperature
+- alpha
+- beta
+- reference_mach
+
+and returns an `AerospaceCondition` class instatiated based on given parameters.
```
benflexcompute, line 62:
```suggestion
and returns an `AerospaceCondition` class instantiated based on given parameters.
```

flow360/__init__.py
```diff
@@ -2,6 +2,10 @@
 This module is flow360 for simulation based models
 """
 
+from examples.migration_guide.extra_operating_condition import (
```
benflexcompute, line 5: Let's not expose these directly under flow360 namespace. Otherwise user will see auto-prompt of these functions in IDE which is not ideal.
can you do 
```suggestion
import examples.migration_guide as migration_helper
```
and import the functions there?

benflexcompute (approved):

flow360/__init__.py
wiktor-flex: But don't we want operating_condition_from_mach_reynolds to actually be accessible normally? From what I know it is a way of describing operating conditions often used during workshop cases so I don't think it should be labeled as just migration helper.

benflexcompute: @maciej-flexcompute What do you think? Should we expose these function at root namespace?

maciej-flexcompute: `operating_condition_from_mach_reynolds` we want to have available from flow360 namespace as this one of standard usage (workshops, airfoils etc). `muRef` should be only migration concept

wiktor-flex: Changed it so that `operating_condition_from_mach_reynolds` is under the flow360 namespace, `operating_condition_from_muref` is available via importing from `extra_operating_condition.py` in migration_guide folder

PR #626, opened 2024-12-13
fix(): add update mechanism for input cache

benflexcompute:


flow360/component/simulation/operating_condition/operating_condition.py
```diff
@@ -202,19 +202,20 @@ class GenericReferenceCondition(MultiConstructorBaseModel):
     ====
     """
 
+    private_attribute_input_cache: GenericReferenceConditionCache = GenericReferenceConditionCache()
```
maciej-flexcompute, line 205: how does it impact type hints? I think IDE will list the options in order of definition, thus private_... and type_name should be last?

flow360/component/simulation/primitives.py
```diff
@@ -263,16 +263,25 @@ class Box(MultiConstructorBaseModel, _VolumeEntityBase):
     ====
     """
 
+    # private_attribute_input_cache has to be defined first for field validators to work.
+    private_attribute_input_cache: BoxCache = pd.Field(BoxCache(), frozen=True)
```
maciej-flexcompute, line 267: oh, interesting. Question how we can modify type hints..

flow360/component/simulation/operating_condition/operating_condition.py
benflexcompute: You are right. input_cache will become the first.

flow360/component/simulation/primitives.py
benflexcompute: Never mind, I just tested and it seems that even though fields are validated by definition ordering but it is still accessible and has definition (None).

I thought it is like C++ where you reference an uninitialized pointer.

reference: https://docs.pydantic.dev/latest/concepts/validators/#field-validators:~:text=Validation%20is%20done%20in%20the%20order%20fields%20are%20defined%2C%20so%20you%20have%20to%20be%20careful%20when%20using%20ValidationInfo.data%20to%20not%20access%20a%20field%20that%20has%20not%20yet%20been%20validated/populated

flow360/component/simulation/operating_condition/operating_condition.py
benflexcompute: Reverted the ordering of private_attribute_input_cache to make type hint a bit cleaner.

maciej-flexcompute (approved):

PR #634, opened 2024-12-20
[FL-71] BET translator

wiktor-flex:
Done
- unittest comparing both individual values and dicts against reference files
- C81
- DFDC
- XFoil
- XROTOR
- MultiConstructorBaseModel for BETDisk class
- original input file contents as private attributes
- entitylist working properly
- alias for generate_xrotor_bet_json so that dfdc doesn't use the same name

Some files are left in order to decide which unittests to delete

There are 2 kinds of unittests for BET translator currently:
- compare individual values between a translated output and original reference json
- compare translated output json with a new reference json

BET will mostly likely need to be separated form volume_models.py as it takes a considerable amount of lines of code


flow360/component/simulation/models/bet/bet_translator_interface.py
```diff
@@ -0,0 +1,1529 @@
...
+    for i in range(xrotor_dict["nGeomStations"]):
+        # pylint: disable=no-member
+        r = xrotor_dict["rRGeom"][i] * multiplier * u.m / mesh_unit
+        twist = xrotor_dict["beta0Deg"][i]
+        twist_vec.append({"radius": r * length_unit, "twist": twist * angle_unit})
+
+    return twist_vec
+
+
+def generate_chords(xrotor_dict, mesh_unit, length_unit):
```
benflexcompute, line 1086: What is the difference between `length_unit` and `mesh_unit`?

flow360/component/simulation/models/bet/utils.py
```diff
@@ -0,0 +1,90 @@
+"""
+Utility functions for bet_translator_interface
+"""
+
+import operator
+from math import exp, log
+
+
+# pylint: disable=missing-class-docstring
+class Array(list):
```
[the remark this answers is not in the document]
wiktor-flex: I don't think we really need that, I just wanted to retain most of the original code. I think that it can most likely be replaced with np.array as you suggested

flow360/component/simulation/models/bet/utils.py
```diff
@@ -0,0 +1,90 @@
...
+        super().__init__(*args, **kwargs)
+
+    # pylint: disable=missing-function-docstring
+    def operator(self, op, value):
+        result = self.copy()
+        for i, x in enumerate(result):
+            result[i] = op(x, value)
+        return Array(result)
+
+    # pylint: disable=missing-function-docstring
```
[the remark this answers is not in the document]
wiktor-flex: Yes, if we decide to use those structures, then I can definitely add docstrings describing their functionality

flow360/component/simulation/models/volume_models.py
```diff
@@ -476,8 +486,86 @@ class BETDiskSectionalPolar(Flow360BaseModel):
     )
 
 
+class XROTORFile(Flow360BaseModel):
+    file_name: str
+    type_name: Literal["XRotorFile"] = pd.Field("XRotorFile", frozen=True)
+    content: str = pd.Field()
+
+    @pd.model_validator(mode="before")
```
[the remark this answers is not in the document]
wiktor-flex: This is something that I discussed with Maciej that we want to have a complete record of what was the user's original input, in this case contents of the provided file, so that if at any time the user wants to check what was used as an input to their simulation, they can do so without having to keep track of their local files.

flow360/component/simulation/models/volume_models.py
```diff
@@ -476,8 +486,86 @@ class BETDiskSectionalPolar(Flow360BaseModel):
...
+    @classmethod
+    def _extract_content(cls, input_data):
```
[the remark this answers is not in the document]
wiktor-flex: Good point

flow360/component/simulation/models/bet/bet_translator_interface.py
wiktor-flex: Yes, I think this can be simplified and mesh_unit will be unnecessary, thank you

flow360/component/simulation/models/volume_models.py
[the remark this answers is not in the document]
benflexcompute: I am not suggesting we get rid of the `content`. I am just saying using validator to read the file will break things.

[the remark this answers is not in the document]
benflexcompute: Because validator is executed every time deserialization happens.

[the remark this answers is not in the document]
wiktor-flex: So for example calling something like:
`file.content = extract_content(file.file_name)`
in from_c81/from_dfdc/from_xfoil/from_xrotor BETDisk methods?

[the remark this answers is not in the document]
benflexcompute: Yes similar to that. I will send a PR based on yours addressing some of the comments and you can take a look tmr.

tests/simulation/translator/data/c81/Xv15_c81_section1Polars.csv
benflexcompute, line 1: Can we move all these files to a new location? These are translators but not "the" translator. tests/simulation/translator contains tests for the SimulationParam --> Flow360.json translations.

See my new PR for proposed new location.

flow360/component/simulation/models/bet/utils.py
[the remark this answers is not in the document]
maciej-flexcompute: correct, the original code tried to get rid of numpy dependecy, but now it is not needed anymore.

flow360/component/simulation/models/volume_models.py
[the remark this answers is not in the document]
maciej-flexcompute: I would think otherwise, let me explain:
1. Scenario 1, user creates BET from file
2. Scenario 2, BET is created from json, e.g., from webUI or loaded from local storage

Currently, the principle we are following is that our models are validated upon construction, so in theory we can make everything immutable. `XRotorFile` without content is NOT a "valid" model, because file might not exist or can be invalid file.

In both scenarios I mentioned, after construction of `XRotorFile` object, it should be complete. In scenario 1, read the file, it should be accessible according to current CWD, if file not fund, user can provide absolute path etc. In scenario 2, the content is already provided, so skip validation.

This is also stateless design - we don't have states *before* `read_file()` explicitly called and *after*.

flow360/component/simulation/models/volume_models.py
```diff
@@ -476,8 +486,86 @@ class BETDiskSectionalPolar(Flow360BaseModel):
     )
 
 
+class XROTORFile(Flow360BaseModel):
+    file_name: str
```
maciej-flexcompute, line 490: this is very close to `filename` which is `Flow360BaseModel` parameter. I bet people will make this mistake and the behaviour will be different. I propose adding alias here or just calling it `filename`

flow360/component/simulation/models/volume_models.py
```diff
@@ -476,8 +486,86 @@ class BETDiskSectionalPolar(Flow360BaseModel):
...
+    omega: Optional[AngularVelocityType.NonNegative] = None
+    chord_ref: Optional[LengthType.Positive] = None
+    n_loading_nodes: Optional[pd.StrictInt] = None
+    entities: Optional[EntityList[Cylinder]] = None
+    angle_unit: Optional[AngleType] = None
+    length_unit: Optional[LengthType.NonNegative] = None
+    mesh_unit: Optional[LengthType.NonNegative] = None
+    number_of_blades: Optional[pd.StrictInt] = None
+    initial_blade_direction: Optional[Axis] = None
+    blade_line_chord: Optional[LengthType.NonNegative] = None
```
maciej-flexcompute, line 564: this class is very similar to the original `BETDisk` but without polars, can we use base abstract and then `BETDisk` and `BETDiskCache` inherit?

flow360/component/simulation/models/volume_models.py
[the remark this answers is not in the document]
benflexcompute: Then let's do:
1. Make All BETFile models' fields immutable.
2. Add an after model validator to read the file and populate `content` if `content` is None.
Is this what you think? @maciej-flexcompute

flow360/component/simulation/models/volume_models.py
benflexcompute: The `Optional`s are different between `BETDiskCache` and `BETDisk` though.

If we make them all `Optional` also in `BETDisk` then there will be a lot of validation to do.

flow360/component/simulation/models/volume_models.py
```diff
@@ -476,8 +486,86 @@ class BETDiskSectionalPolar(Flow360BaseModel):
...
+
+
 # pylint: disable=no-member
-class BETDisk(Flow360BaseModel):
+class BETDisk(MultiConstructorBaseModel):
```
benflexcompute, line 568: @wiktor-flex Can you add some test as in https://github.com/flexcompute/Flow360/blob/develop/tests/simulation/framework/test_multi_constructor_model.py to make sure that the dumped JSON can be deserialized back to proper BETDisk object?

For starter you need to overload the type_name field in BETDisk so the multiconstructor knows what class/model to use for deseralizing.

flow360/component/simulation/models/volume_models.py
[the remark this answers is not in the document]
maciej-flexcompute: I think it should be *before* validator and make the *content* required. I think *after* validator in general cannot assign because it re-evaluate itself. I am not saying that we need to go immutable here, but just saying we can simply turn it on if we decide one day.

flow360/component/simulation/models/volume_models.py
maciej-flexcompute: Why optionals are different?

benflexcompute: I see. Yes, most optional can be removed. Ignore my previous comment.

flow360/component/simulation/models/bet/utils.py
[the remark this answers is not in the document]
benflexcompute: Got rid of it.

flow360/component/simulation/models/bet/utils.py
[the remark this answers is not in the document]
benflexcompute: Removed

tests/simulation/translator/data/c81/Xv15_c81_section1Polars.csv
benflexcompute: Moved

flow360/component/simulation/models/volume_models.py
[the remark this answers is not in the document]
benflexcompute: Ok. I just do not like before validators since we will have to deal with all the crazy input we might get. (e.g. typos like file_Name instead of file_name). But fine this is not a critical part and let's just do before validator.

flow360/component/simulation/models/volume_models.py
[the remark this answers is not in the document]
benflexcompute: Done

flow360/component/simulation/models/volume_models.py
benflexcompute: Added such tetss

flow360/component/simulation/models/volume_models.py
benflexcompute: I just realized by removing "Optional" the BETDiskCache has to be constructed with non-None values.

For example the default for `BETDisk.private_attribute_input_cache` is now `BETDiskCache()` which will fail because we have None for `omega`

benflexcompute: @maciej-flexcompute Can we keep this as is?

flow360/component/simulation/models/bet/bet_translator_interface.py
```diff
@@ -0,0 +1,1529 @@
...
+    Attributes
+    ----------
+    return: alpha_list, mach_list, cl_list, cd_list
+    """
+    cl_alphas = []
+    cl_values = {}
+    cd_values = {}
+
+    # pylint: disable=unspecified-encoding, consider-using-with
+    xfoil_fid = open(polar_file, "r")
```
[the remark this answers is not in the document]
benflexcompute: Using with

flow360/component/simulation/models/volume_models.py
benflexcompute: I changed to file_path. This is actually what it is.

flow360/component/simulation/models/volume_models.py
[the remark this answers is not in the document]
benflexcompute: And the proposition cannot handle one case:
If user changed the file_name like:
```python
file = (
        fl.C81File(
            file_path=(
                os.path.join(
                    os.path.dirname(os.path.abspath(__file__)), "data/c81", "Xv15_geometry.csv"
                )
            )
        ),
    )

file.file_path = "non-exist_file.csv"
```
Then the file content would not update since the validator sees that content is not empty.

flow360/component/simulation/migration/BETDisk.py
```diff
@@ -0,0 +1,218 @@
...
+        raise ValueError(
+            "The supplied Flow360 input for BETDisk hsa invalid format. Details: " + str(err) + "."
+        ) from err
+
+
+@validate_call
+def read_all_v1_BETDisks(
+    file_path: str,
+    mesh_unit: LengthType.NonNegative,  # pylint: disable = no-member
+    time_unit: TimeType.Positive,  # pylint: disable = no-member
```
maciej-flexcompute, line 174: I don't think this is an issue, but what other value than "second" user might want to input here?

flow360/component/simulation/models/bet/bet_translator_interface.py
```diff
@@ -0,0 +1,1529 @@
...
+    """
+    Generate the list of Alphas that the BET 2d section polar is for in 1 degree steps from -180 to 180.
+
+    Attributes
+    ----------
+    return: list of floats
+    """
+
+    neg_ang = float_range(-180, -9)
+    pos_ang = [
```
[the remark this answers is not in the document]
maciej-flexcompute: I meant this:
```python
pos_ang = np.concatenate([np.arange(-9, -1, 0.5), np.arange(-1, 3, 0.25), np.arange(3, 9.5, 0.5)])
```

flow360/component/simulation/models/volume_models.py
```diff
@@ -617,6 +698,335 @@ def check_bet_disk_3d_coefficients_in_polars(self):
...
+        ...     n_loading_nodes=20,
+        ...     entities=bet_cylinder,
+        ...     length_unit=fl.u.m,
+        ...     mesh_unit=fl.u.m,
+        ...     angle_unit=fl.u.deg,
+        ... )
+        """
+
+        params = generate_dfdc_bet_json(
+            dfdc_file_name=file.file_path,
```
maciej-flexcompute, line 851: this will not work with webUI, we need to pass in `content`

flow360/component/simulation/models/volume_models.py
```diff
@@ -617,6 +698,335 @@ def check_bet_disk_3d_coefficients_in_polars(self):
...
+        ...     n_loading_nodes=20,
+        ...     entities=bet_cylinder_imperial,
+        ...     length_unit=fl.u.m,
+        ...     angle_unit=fl.u.deg,
+        ...     number_of_blades=3,
+        )
+        """
+
+        params = generate_xfoil_bet_json(
+            geometry_file_name=file.file_path,
```
maciej-flexcompute, line 936:
```suggestion
            geometry_file_name=file.content,
```

flow360/component/simulation/models/volume_models.py
```diff
@@ -617,6 +698,335 @@ def check_bet_disk_3d_coefficients_in_polars(self):
...
+        ...     chord_ref=14 * fl.u.m,
+        ...     n_loading_nodes=20,
+        ...     entities=bet_cylinder,
+        ...     angle_unit=fl.u.deg,
+        ...     length_unit=fl.u.m,
+        ... )
+        """
+
+        params = generate_xrotor_bet_json(
+            xrotor_file_name=file.file_path,
```
maciej-flexcompute, line 1016:
```suggestion
            xrotor_file_name=file.content,
```

flow360/component/simulation/models/volume_models.py
benflexcompute: Good point.

flow360/component/simulation/models/volume_models.py
benflexcompute: Done

flow360/component/simulation/models/volume_models.py
benflexcompute: Done

flow360/component/simulation/models/bet/bet_translator_interface.py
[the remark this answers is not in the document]
benflexcompute: Applied

maciej-flexcompute (approved):

benflexcompute (approved):

PR #640, opened 2025-01-06
Support the mix of Point and PointArray in ProbeOutput and SurfaceProbeOutput

angranl-flex:


flow360/component/simulation/outputs/outputs.py
```diff
@@ -479,12 +479,6 @@ def load_point_location_from_file(cls, file_path: str):
         """Load probe point locations from a file. (Not implemented yet)"""
         raise NotImplementedError("Not implemented yet.")
 
-    @pd.field_validator("entities", mode="after")
-    @classmethod
-    def check_unique_probe_type(cls, value):
```
benflexcompute, line 484: Are all these functiona removed? I remember there was one for surface slice which does not make sense

benflexcompute (approved):
Nicely done 👍

flow360/component/simulation/outputs/outputs.py
angranl-flex: Just checked and removed all of them in this file.

PR #660, opened 2025-01-15
Add Updater to 24.11.7

angranl-flex:
Handle the missing private_attribute_id of PointArray in the version < 24.11.6

### DIFF
--- a/flow360/component/simulation/framework/updater.py
+++ b/flow360/component/simulation/framework/updater.py
@@ -7,10 +7,10 @@ TODO: remove duplication code with FLow360Params updater.
 # pylint: disable=R0801
 
 import re
-from typing import List
 
 from ....exceptions import Flow360NotImplementedError, Flow360RuntimeError
 from .entity_base import generate_uuid
+from .updater_utils import compare_dicts
 
 
 def _no_update(params_as_dict):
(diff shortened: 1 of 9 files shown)


flow360/component/simulation/framework/updater.py
```diff
@@ -7,16 +7,70 @@
...
 
 
+def _24_11_6_to_24_11_7_update(params_as_dict):
+    # Check if PointArray has private_attribute_id. If not, generate the uuid and assign the id
+    # to all occurance of the same PointArray
+    def compare_entity(
+        entity_dict_new: dict, entity_dict_to_match: dict, exclude_keys: List[str] = None
+    ):
+        for key in entity_dict_to_match.keys():
+            if key in exclude_keys:
```
benflexcompute, line 27: What if exclude_keys is None?

benflexcompute: Can you reuse Flow360/tests/utils.py->compare_dicts ?

benflexcompute: Move the function out if necessary.

flow360/component/simulation/framework/updater.py
```diff
@@ -7,16 +7,70 @@
...
+                    entity.get("private_attribute_entity_type_name") == "PointArray"
+                    and entity.get("private_attribute_id") is None
+                ):
+                    new_uuid = generate_uuid()
+                    entity["private_attribute_id"] = new_uuid
+                    point_array_list.append(entity)
+
+    if params_as_dict.get("private_attribute_asset_cache"):
+        for idx, draft_entity in enumerate(
+            params_as_dict["private_attribute_asset_cache"]["project_entity_info"]["draft_entities"]
```
benflexcompute, line 52: Are you sure this chain of keys is always available?

tests/data/simulation/simulation_24_11_6.json
benflexcompute, line 1: Can we delete the part that are not involved? This is a large file.

angranl-flex: OK, I used the simulation.json from a smaller case for this test.

flow360/component/simulation/framework/updater.py
angranl-flex: Since "private_attribute_asset_cache" and "draft_entities" are not None by default, I only checked the existence of "project_entity_info" key now.

flow360/component/simulation/framework/updater.py
angranl-flex: OK, the related compare_* functions are moved to updater_utils.py

benflexcompute (approved):

PR #666, opened 2025-01-17
[SCFD-4128] [SCFD-4126] feat(): Removed validation process for unit system conversion

benflexcompute:
- [x] feat(): Removed validation process for unit system conversion
- [x] Change temperature unit for imperial to be Fahrenheit


flow360/component/simulation/services.py
```diff
@@ -58,6 +60,7 @@
 )
 from flow360.component.utils import remove_properties_by_name
 from flow360.exceptions import Flow360TranslationError
+from flow360.log import log
```
angranl-flex, line 63: Should we remove this import since it is not used?

tests/simulation/service/test_services_v2.py
```diff
@@ -690,3 +690,45 @@ def get_validation_levels_to_use(root_item_type, requested_levels):
         "Case",
         "VolumeMesh",
     ]
+
+
+def test_unit_system_conversion():
+    with open("data/simulation_param.json", "r") as fp:
+        dict_to_convert = json.load(fp)
+    services.change_unit_system(data=dict_to_convert, new_unit_system="Imperial")
```
angranl-flex, line 698: Do we also need to test the conversion to "CGS" and "SI"?

angranl-flex (approved):

tests/simulation/service/test_services_v2.py
benflexcompute: Added to CGS conversion and SI conversion.

PR #672, opened 2025-01-21
Updated geometry to report example

maciej-flexcompute:
Updated example to use latest solver hotfix

### DIFF
--- a/examples/plugins/geometry_to_report.py
+++ b/examples/plugins/geometry_to_report.py
@@ -13,8 +13,10 @@ from flow360.plugins.report.report_items import (
 )
 from flow360.plugins.report.utils import Average, DataItem, Delta, Expression, Variable
 
-project_id = None  # if running for the first time, then replace it with project ID to avoid re-creation of projects
-# project_id = "prj-b5a0ae52-14c7-4f0c-813b-542763f993a2"
+project_id = None  # if running for the first time
+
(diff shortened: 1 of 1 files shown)


examples/plugins/geometry_to_report.py
```diff
@@ -13,8 +13,8 @@
 )
 from flow360.plugins.report.utils import Average, DataItem, Delta, Expression, Variable
 
-# do project_id = None if running for the first time, than replace it with project ID to avoid re-creation of projects
-project_id = "prj-7096e7c1-2d6e-4e0a-803d-da1d61a99d96"
+project_id = None  # if running for the first time, than replace it with project ID to avoid re-creation of projects
```
wiktor-flex, line 16:
```suggestion
project_id = None  # if running for the first time, then replace it with project ID to avoid re-creation of projects
```

examples/plugins/geometry_to_report.py
```diff
@@ -13,8 +13,8 @@
...
+# project_id = "prj-b5a0ae52-14c7-4f0c-813b-542763f993a2"
```
wiktor-flex, line 17: Is this an example of what a project_id is supposed to look like? If so then I think an appropriate comment would be nice for user's clarity

wiktor-flex (approved):

PR #673, opened 2025-01-21
added support for UAT env when setting API key

maciej-flexcompute:
How to use:

```
flow360 configure --apikey=... --profile=auto_test_1 --uat
```
then, when using:
```python
import flow360 as fl
from flow360.user_config import UserConfig

UserConfig.set_profile("auto_test_1")
fl.Env.uat.active()
```
it will parse correct API key.


wiktor-flex (approved):
Looks good, uploading project to uat from python works

PR #675, opened 2025-01-21
fix(): add check for the existence of _webapi attribute when building asset in the local mode.

angranl-flex:


flow360/component/simulation/web/asset_base.py
```diff
@@ -218,9 +218,11 @@ def _from_local_storage(
             params_dict = json.load(f)
 
         asset_obj = cls._from_supplied_entity_info(params_dict, cls(asset_id))
-        asset_obj._webapi._download_file = _local_download_file  # pylint: disable=protected-access
-        if meta_data is not None:
-            asset_obj._webapi._set_meta(meta_data)  # pylint: disable=protected-access
+        # pylint: disable=protected-access
+        if hasattr(asset_obj, "_webapi"):
```
benflexcompute, line 222: if not hasattr(asset_obj, "_webapi"):
  return ...

benflexcompute: Add a comment here indicating that this is supposed to handle local execution which has no valid ID.

benflexcompute (approved):

PR #676, opened 2025-01-21
Extend report generation feature for service tests

angranl-flex:
- [x] Add linear residual result
- [x] Add `include` option to the `Chart2D` and `DataItem` class
- [x] Improve get_variables function in the `Expression` class to also filter math functions (such as `abs`, `where`) in the expression.
- [x] Fix the data filter in the PerEntityResultsCSVModel
- [x] Add TableSettings Class for selecting certain cases to show in the table.
- [x] ~Add `output_bool` option in the Average operation to output the boolean result after averaging.~ Add "bool" formatter to convert the output integer to boolean type result

### DIFF
--- a/flow360/plugins/report/report_items.py
+++ b/flow360/plugins/report/report_items.py
@@ -307,6 +307,8 @@ class Table(ReportItem):
         def make_callable(fmt):
             if callable(fmt):
                 return fmt
+            if fmt == "bool":
+                return lambda x: f"{bool(x)}"
             return lambda x, ff=fmt: f"{x:{ff}}"
 
         formatters = [make_callable(f) for f in formatters]  # pylint: disable=not-an-iterable
--- a/flow360/plugins/report/report_items.py
+++ b/flow360/plugins/report/report_items.py
@@ -771,7 +771,7 @@ class Chart2D(Chart):
     type_name : Literal["Chart2D"], default="Chart2D"
         Specifies the type of report item as "Chart2D"; this field is immutable.
     include : Optional[List[str]]
-        List of boundaries to include from data. Applicable to:
+        List of boundaries to include in data. Applicable to:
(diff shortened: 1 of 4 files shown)


flow360/plugins/report/report_items.py
```diff
@@ -106,6 +106,23 @@ class Settings(Flow360BaseModel):
...
+        The indices of cases to be included in the table
+
+    case_excluded : List[pd.NonNegativeInt], optional
+        The indices of cases to be excluded from the table
+    """
+
+    case_include: Optional[List[pd.NonNegativeInt]] = None
+    case_exclude: Optional[List[pd.NonNegativeInt]] = None
+
+
```
maciej-flexcompute, line 125: in other charts we use "select_indices" and this here has very similar function. I would suggest to unify:
https://github.com/flexcompute/Flow360/blob/c721753c36ae0643e95014add61ae62bcad67b57/flow360/plugins/report/report_items.py#L413

I think `select_indices` is not the best name.

flow360/plugins/report/utils.py
```diff
@@ -378,6 +383,7 @@ class Average(GenericOperation):
     start_time: Optional[pd.NonNegativeFloat] = None
     end_time: Optional[pd.NonNegativeFloat] = None
     fraction: Optional[pd.PositiveFloat] = pd.Field(None, le=1)
+    output_bool: pd.StrictBool = pd.Field(False)
```
maciej-flexcompute, line 386: I don't understand intention here?

angranl-flex: In the service test, we want to compare the computed value with a reference value and output the comparison result in True/False. One example of this is shown in the figure below:
image

To recreate this column in the table, I define an Expression using `where` operation and use average to reduce the comparison result to a single value (1 or 0) and output as boolean type.

maciej-flexcompute (approved):

PR #683, opened 2025-01-23
change location and keys of X/Y_slicing_forceDistribution.csv

feilin-flexcompute:
Due to a change of algorithm of force distribution, some changes:
1. force distribution csv files were uploaded to results/ instead of results/postprocess/
2. an old key CD_per_length is removed
3. an new key stride is added to force distribution over Y


benflexcompute (approved):

flow360/component/results/case_results.py
```diff
@@ -156,8 +156,8 @@ class CaseDownloadable(Enum):
     BET_FORCES = "bet_forces_v2.csv"
     ACTUATOR_DISKS = "actuatorDisk_output_v2.csv"
     LEGACY_FORCE_DISTRIBUTION = "postprocess/forceDistribution.csv"
-    Y_SLICING_FORCE_DISTRIBUTION = "postprocess/Y_slicing_forceDistribution.csv"
-    X_SLICING_FORCE_DISTRIBUTION = "postprocess/X_slicing_forceDistribution.csv"
+    Y_SLICING_FORCE_DISTRIBUTION = "Y_slicing_forceDistribution.csv"
+    X_SLICING_FORCE_DISTRIBUTION = "X_slicing_forceDistribution.csv"
```
feilin-flexcompute, line 160: note: this change is only for >=release-25.2.

flow360/component/results/case_results.py
```diff
@@ -793,8 +793,7 @@ class XSlicingForceDistributionResultCSVModel(PerEntityResultCSVModel):
         CaseDownloadable.X_SLICING_FORCE_DISTRIBUTION.value, frozen=True
     )
 
-    _variables: List[str] = [_CUMULATIVE_CD_CURVE, _CD_PER_LENGTH]
-    _filter_when_zero = [_CD_PER_LENGTH]
+    _variables: List[str] = [_CUMULATIVE_CD_CURVE]
```
maciej-flexcompute, line 796: why do we remove "per_length". It was used in filter when 0. Essentially it was used to get extend of data in x when excluding some boundaries. If you exclude all the boundaries you can get rid of all rows where all values are 0 and you will end up with non-zero values of which you can get x extend and zoom in plot.

feilin-flexcompute: In the new algorithm, the per_length data can be calculated by the X value and cumulative_CD data, so no need to have duplication.

maciej-flexcompute: I see, make sense. But now, how do I zoom in to the relevant range of data given that I don't know BBOXes of boundaries?

flexcompute-readthedocs: updated that the CD_per_strip is calculated in the _preprocess() function of the XSlicingForceDistributionResultCSVModel(). It is added into the _filter_when_zero array too. 
It is tested in the unit tests.

maciej-flexcompute (approved):

benflexcompute:
Please also send hotfix to 25.2

flow360/component/results/case_results.py
```diff
@@ -793,10 +795,26 @@ class XSlicingForceDistributionResultCSVModel(PerEntityResultCSVModel):
...
 
+    def _preprocess(self, filter_physical_steps_only: bool = False, include_time: bool = False):
+        """
+        add _CD_PER_STRIP for filtering purpose and preprocess
+        """
+        for entity in self.entities:
+            header = f"{entity}_{_CUMULATIVE_CD_CURVE}"
+            cumulative_cd = np.array(self._values[header])
+            _cd_per_strip = np.diff(cumulative_cd)
+            cd_per_strip = np.insert(_cd_per_strip, 0, cumulative_cd[0])
```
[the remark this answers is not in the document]
feilin-flexcompute: fixed.

PR #692, opened 2025-01-27
fix(): Removed auto error message about mesh file format. And bump default solver version to 25.0

benflexcompute:


angranl-flex (commented):
Do we need to update the error message corresponding to the unknown format in SurfaceMeshFile/VolumeMeshFile/GeometryMeshFiles? For instantce, "Unsupported/Unknown file extensions: {parser.format.ext()}."

benflexcompute:
> Do we need to update the error message corresponding to the unknown format in SurfaceMeshFile/VolumeMeshFile/GeometryMeshFiles? For instantce, "Unsupported/Unknown file extensions: {parser.format.ext()}."

I think Unsupported is sufficient.

angranl-flex (approved):

PR #697, opened 2025-01-28
Add fluid property to simulation params class

angranl-flex:


flow360/component/simulation/simulation_params.py
```diff
@@ -490,6 +490,41 @@ def _update_entity_private_attrs(self, registry: EntityRegistry) -> EntityRegist
...
+    def fluid(self) -> Fluid:
+        """
+        Returns the fluid model of the simulation params
+
+        Returns
+        -------
+        Fluid
+            The fluid model in the params
+        """
+        return self._get_model_by_type(model_type=Fluid)
```
maciej-flexcompute, line 526: although I understand desire to make it accessible from top level, I think it will create more confusion, why all the sudden fluid is a top level and other models are not? On contrary if we put more things on top level, we will lose hierarchical structure and come back to flat structure. Note that user might want to do `params.fluid = fl.Fluid(...)`, such assignment is not allowed.

How about we keep it under models and do:
```python
params.models.fluid
```
We need to extend "List" type. This way we can also do: `fluid`, `solid` `rotations`, `bet_disks` etc...

PR #698, opened 2025-01-29
feat(): Flow360.json parser repack for BET translator

benflexcompute:
TODO:

- [ ] Add MultiConstructor test to see if the Simulation json can be desacralized to generate BETDisk instances.
- [ ] Add SimulationParam.time_unit, length_unit etc to make user's life easier.
- [ ] Add quantitative reference results for the Flow360 BETDisk translators.


flow360/component/simulation/models/volume_models.py
```diff
@@ -542,8 +565,15 @@ def _extract_content(cls, input_data):
         return {"file_name": input_data["file_name"], "content": content_to_store}
 
 
+class Flow360File(BETInputFileBase):
+    """Flow360 File containing setting for a single BET disk"""
+
+    type_name: Literal["Flow360File"] = pd.Field("Flow360File", frozen=True)
```
maciej-flexcompute, line 571: Is this for V1 BET config?

flow360/component/simulation/models/volume_models.py
```diff
@@ -1041,6 +1073,102 @@ def from_xrotor(
...
 
+    @MultiConstructorBaseModel.model_constructor
+    @pd.validate_call
+    def from_flow360(
+        cls,
+        file: Flow360File,
+        mesh_unit: LengthType.NonNegative,
+        time_unit: TimeType.Positive,
+    ):
+        """Constructs a :class: `BETDisk` instance from a given V1 (legacy) Flow360 input.
```
maciej-flexcompute, line 1084: This is for V1 migration. Do we want to keep it here? I would imagine interface like:
```python
from flow360 import migrations

bet = migrations.BETDisk.from_v1_one(...) # when migrating single disk
bet_list = migrations.BETDisk.from_v1_all("flow360.json") # when migrating all disks
```

flow360/component/simulation/models/volume_models.py
```diff
@@ -1041,6 +1073,102 @@ def from_xrotor(
 
         return cls(**params)
 
+    @MultiConstructorBaseModel.model_constructor
```
maciej-flexcompute, line 1076: why multiconstructor for migration? I think we don't need to store original input in this instance? That would double the storage and BET are already large

flow360/component/simulation/models/volume_models.py
```diff
@@ -486,6 +492,23 @@ class BETDiskSectionalPolar(Flow360BaseModel):
...
 
 
+class BETInputFileBase(Flow360BaseModel, metaclass=ABCMeta):
+    """Base class for BET Disk input files"""
+
+    file_name: str
+    type_name: Literal["BETInputFileBase"] = pd.Field("BETInputFileBase", frozen=True)
+    content: Optional[str] = pd.Field(None)
+
+    def read_file(self):
```
maciej-flexcompute, line 502: see my comment here: https://github.com/flexcompute/Flow360/pull/634#discussion_r1933646841

flow360/component/simulation/models/volume_models.py
benflexcompute: It is for Flow360.json so V0. Not the V1 Flow360Params.

flow360/component/simulation/models/volume_models.py
benflexcompute: Ok. Yes if we think this is migration instead of a construction method.

flow360/component/simulation/models/volume_models.py
benflexcompute: Sure.

maciej-flexcompute (approved):

PR #699, opened 2025-01-29
Add updater for beta version and remove treat_as_file_content/use_updater option

angranl-flex:


flow360/component/simulation/framework/updater_utils.py
```diff
@@ -63,21 +63,22 @@ class Flow360Version:
...
         # Match three groups of digits separated by dots
-        match = re.match(r"^(\d+)\.(\d+)\.(\d+)$", version.strip())
+        match = re.match(r"^(\d+)\.(\d+)\.(\d+)(?:b(\d+))?$", version.strip())
         if not match:
             raise ValueError(f"Invalid version string: {version}")
 
         self.major = int(match.group(1))
         self.minor = int(match.group(2))
         self.patch = int(match.group(3))
+        self.beta = int(match.group(4)) if match.group(4) else None
```
maciej-flexcompute, line 81: how about comparisons? I guess we should add them as well?

flow360/component/simulation/translator/utils.py
```diff
@@ -62,15 +62,14 @@ def get_simulation_param_dict(
...
             # If input is a file path
-            with open(input_params, "r", encoding="utf-8") as file:
-                param_dict = json.load(file)
-        if param_dict is None:
-            raise ValueError(f"Invalid input <{input_params}> for translator. ")
-        param = SimulationParams(use_updater=True, **param_dict)
+            param = SimulationParams(filename=input_params)
     elif isinstance(input_params, dict):
-        param = SimulationParams(use_updater=True, **input_params)
+        param = SimulationParams(**input_params)
```
maciej-flexcompute, line 72: why do we need to worry about updates in translator. Can we use translator *always* on SimulationParams? Then we know they are updated.

benflexcompute: Since preprocess (actually just non-dimensionalization for now) should not be called by user. We should hide it and restrict the possible inputs. The only place it is called now is in the service. 
Let's do this:
1. Make SimulationParam.preprocess() private (_preprocess()).
2. Delete this dict path. I do not think it is triggered anywhere. The str path is probably also useless except for some unit test but let's keep for now.
3. rename get_simulation_param_dict to preprocess_param. It is not doing what the old name is saying.

flow360/component/simulation/framework/updater_utils.py
benflexcompute: Most likely the beta release is on some version that is not released yet. So I think if we have the updater for that release then it is fine. E.g. 26.6.0b2 should be using 26.6.0's updater. I think in the version milestones there should not be any beta versions.

Also Angran told me that the previous behavior was to ignore the beta version number.

flow360/component/simulation/translator/utils.py
angranl-flex: Done

flow360/component/simulation/framework/updater_utils.py
maciej-flexcompute: I can agree on the usage, but I am just looking at code and I find it not logical or incomplete. Why do we store "beta" if it is not used?

angranl-flex: I can remove the store of "beta" and ensure Flow360Version can match the beta version like 26.6.0b2, will this work?

maciej-flexcompute (approved):

benflexcompute (approved):

PR #700, opened 2025-01-29
Add rotation_correction support for SST

angranl-flex:


benflexcompute (approved):
Please also hotfix 25.2

PR #703, opened 2025-01-30
fix(): Add validation that output fields should match turb model used

benflexcompute:


flow360/component/simulation/validation/validation_output.py
```diff
@@ -68,3 +70,42 @@ def extract_literal_values(annotation):
...
+            turbulence_model = model.turbulence_model_solver.type_name
+            break
+
+    for output_index, output in enumerate(params.outputs):
+        if output.output_type == "AeroAcousticOutput":
+            continue
+        for item in output.output_fields.items:
+            if item in invalid_output_fields[turbulence_model]:
+                raise ValueError(
+                    f"In `outputs`[{output_index}]:, {item} is not valid"
```
angranl-flex, line 99:
```suggestion
                    f"In `outputs`[{output_index}]:, {item} is not a valid"
```

angranl-flex: Other than the output_index, do we need to display the output type in the error message?

flow360/component/simulation/validation/validation_output.py
```diff
@@ -68,3 +70,42 @@ def extract_literal_values(annotation):
...
+                    f" output field when using turbulence model: {turbulence_model}."
+                )
+
+        if output.output_type == "IsosurfaceOutput":
+            # using the 1st item as all items have same field definition as the 1st one.
```
angranl-flex, line 104: What does this comment mean?

tests/simulation/params/test_validators_output.py
```diff
@@ -21,3 +34,53 @@ def test_aeroacoustic_observer_unit_validator():
                 fl.Observer(position=[0.0001, 0.02, 0.03] * u.mm, group_name="1"),
             ],
         )
+
+
+def test_aeroacoustic_observer_unit_validator():
```
angranl-flex, line 39: The function name needs to be changed.

angranl-flex (commented):
Do we also need to check if these invalid fields are used in the UDF expression?

benflexcompute:
> Do we also need to check if these invalid fields are used in the UDF expression?

Good point. UDF has its dedicated checking mechanism in casePipeline validaiton. It does not have this particular validation though. Let's not worry about that now.

flow360/component/simulation/validation/validation_output.py
benflexcompute: Added "a" and the output type name.

tests/simulation/params/test_validators_output.py
benflexcompute: done

flow360/component/simulation/validation/validation_output.py
benflexcompute: Comment updated. I am suing 1st item to get the allowed field since allowed field is the same whichever isosurface I select.

angranl-flex (approved):

PR #706, opened 2025-02-03
feat(): Added function to compute Reynolds number

benflexcompute:
Also changed the output for the mach property function so that it returns number instead of dimensionless unyt object.

### DIFF
--- a/flow360/component/simulation/operating_condition/operating_condition.py
+++ b/flow360/component/simulation/operating_condition/operating_condition.py
@@ -389,13 +389,44 @@ class AerospaceCondition(MultiConstructorBaseModel):
     @property
     def mach(self) -> pd.PositiveFloat:
         """Computes Mach number."""
-        return self.velocity_magnitude / self.thermal_state.speed_of_sound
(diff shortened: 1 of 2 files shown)


flow360/component/simulation/operating_condition/operating_condition.py
```diff
@@ -389,14 +389,41 @@ def check_valid_reference_velocity(self) -> Self:
...
     @pd.field_validator("alpha", "beta", "thermal_state", mode="after")
     @classmethod
     def _update_input_cache(cls, value, info: pd.ValidationInfo):
         setattr(info.data["private_attribute_input_cache"], info.field_name, value)
         return value
 
+    @pd.validate_call
+    def reynolds_number(self, length_unit: LengthType.Positive):
+        """
+        Computes the Reynolds number defined in Flow360 solver.
```
yifan-flex, line 403:
```suggestion
        Computes the Reynolds number defined in Flow360 solver.
```Computes length_unit based Reynolds number

yifan-flex (approved):

PR #707, opened 2025-02-03
Fix updater when checking draft_entities

angranl-flex:


benflexcompute (commented):
[simulation.json](https://github.com/user-attachments/files/18647979/simulation.json)
Can you add this as a test (it is a webUI json) and make sure it goes through?

benflexcompute (approved):

PR #713, opened 2025-02-04
[SCFD-4264] fix(): Add validation to prevent steady simulation with aeroacoustics

benflexcompute:


flow360/component/simulation/validation/validation_output.py
```diff
@@ -108,3 +110,20 @@ def _check_output_fields_valid_given_turbulence_model(params):
...
                     )
     return params
+
+
+def _check_unsteadiness_to_use_aero_acoustics(params):
+
+    if not params.outputs:
+        return params
+
+    if params.time_stepping is not None and isinstance(params.time_stepping, Steady):
```
angranl-flex, line 120: Do we need to check if `params.time_stepping` is None here? 
In SimulationParams, we have `time_stepping: Union[Steady, Unsteady]`

angranl-flex (approved):

flow360/component/simulation/validation/validation_output.py
benflexcompute: Interesting idea. I saw it here and copied it:
https://github.com/flexcompute/Flow360/blob/9355f91a37da3f2238efe846946c2ed33f33d83f/flow360/component/simulation/validation/validation_simulation_params.py#L238
Let me remove it and see if it breaks anything

PR #720, opened 2025-02-07
moved Slater from an Outflow to a wall BC

johnmoore4:


flow360/component/simulation/models/surface_models.py
```diff
@@ -161,6 +160,9 @@ class SlaterPorousBleed(Flow360BaseModel):
     # pylint: disable=no-member
     static_pressure: PressureType.Positive = pd.Field(description="The static pressure value.")
     porosity: float = pd.Field(gt=0, le=1, description="The porosity of the bleed region.")
+    initial_step: Optional[pd.PositiveInt] = pd.Field(
```
benflexcompute, line 163: can we have a more accurate name for this? Maybe activation_step?

flow360/component/simulation/models/surface_models.py
```diff
@@ -296,6 +307,11 @@ class Wall(BoundaryBase):
         0 * u.m,
         description="Equivalant sand grain roughness height. Available only to `Fluid` zone boundaries.",
     )
+    wall_velocity_model: Optional[SlaterPorousBleed] = pd.Field(
```
yifan-flex, line 311: Should we have a more specific name to distinguish this from the `velocity` field? Something like `bleed_velocity_model`?

johnmoore4: I wanted to keep it generic, in case we wanted to add some other wall velocity model that is not necessarily a bleed. It would be a union of the possible models.

yifan-flex: Or should we make `velocity` `Union[VelocityVectorType, SlaterPorousBleed]`?

yifan-flex: Will user be allowed to specify both? If they are conflicting fields we try to use Union, like for `heat_spec`.

benflexcompute: If these are conflicting settings we should have them as a union under same field.

benflexcompute: And we would need to write updater and update the translation function.

benflexcompute: @yifan-flex FYI we decided to keep these as separate options since it is also a stretch putting these two into the same option. I asked @johnmoore4 to add validation for that.

flow360/component/simulation/validation/validation_simulation_params.py
```diff
@@ -57,6 +57,20 @@ def _check_consistency_wall_function_and_surface_output(v):
...
 
+def _check_consistency_wall_velocity_and_model(v):
+    models = v.models
+    if models:
+        for model in models:
+            if isinstance(model, Wall):
+                if model.velocity is not None and model.wall_velocity_model is not None:
+                    raise ValueError(
+                        "Cannot specify both 'velocity' and 'wall_velocity_model' for the same patch. "
+                        "Please specify either one or the other."
```
benflexcompute, line 68: Or neither?

benflexcompute: we can just remove this line. The above message is clear enough.

tests/simulation/params/test_validators_params.py
```diff
@@ -181,6 +182,36 @@ def test_consistency_wall_function_validator(
...
+    )
+
+    # Invalid simulation params
+    with SI_unit_system, pytest.raises(ValueError, match=re.escape(message)):
+        _ = SimulationParams(
+            models=[
+                Wall(
+                    velocity=[0.1, 0.2, 0.3],
+                    surfaces=[Surface(name="noSlipWall")],
+                    wall_velocity_model=SlaterPorousBleed(porosity=0.2, static_pressure=0.1),
```
benflexcompute, line 209: I had discussion with @yifan-flex and let's rename `wall_velocity_model` to `velocity_model` or `modeled_velocity`
1. This is more evident that the velocity is in conflict with this option.
2. "wall" is redundant here.

benflexcompute:
image

yifan-flex (approved):

PR #726, opened 2025-02-10
fix(): Clean Params Based On Validation Levels

angranl-flex:
Fix the issue here: https://www.notion.so/flexcompute/Pending-Angran-velocity-required-when-requesting-meshing-only-131db96dc14e80dab030cd212d014ccf?pvs=4

### DIFF
--- a/flow360/component/simulation/services.py
+++ b/flow360/component/simulation/services.py
@@ -264,12 +264,12 @@ def validate_model(
     validation_warnings = None
     validated_param = None
 
-    params_as_dict = clean_params_dict(params_as_dict, root_item_type)
-
     # The final validation levels will be the intersection of the requested levels and the levels available
     # We always assume we want to run case so that we can expose as many errors as possible
     available_levels = _determine_validation_level(up_to="Case", root_item_type=root_item_type)
     validation_levels_to_use = _intersect_validation_levels(validation_level, available_levels)
+    print(validation_levels_to_use)
+    params_as_dict = clean_params_dict(params_as_dict, validation_levels_to_use)
     try:
         params_as_dict = parse_model_dict(params_as_dict, globals())
         with ValidationLevelContext(validation_levels_to_use):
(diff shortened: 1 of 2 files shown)


flow360/component/simulation/services.py
```diff
@@ -304,8 +304,31 @@ def clean_params_dict(params: dict, root_item_type: str) -> dict:
...
+
+    if params.get("meshing") is not None and "VolumeMesh" not in validation_levels_to_use:
+        params["meshing"].pop("gap_treatment_strength", None)
+        params["meshing"]["defaults"].pop("boundary_layer_growth_rate", None)
+        params["meshing"]["defaults"].pop("boundary_layer_first_layer_thickness", None)
+
+    if params.get("meshing") is not None and "SurfaceMesh" not in validation_levels_to_use:
+        params["meshing"]["defaults"].pop("surface_edge_growth_rate", None)
+        params["meshing"]["defaults"].pop("surface_max_edge_length", None)
+        params["meshing"]["defaults"].pop("curvature_resolution_angle", None)
```
maciej-flexcompute, line 331: this code is dangerous and error prone. Why do we need to clean up? Where in the workflow will we fail if we don't?

maciej-flexcompute: I believe pydantic is able to dump model partially, meaning we dump only parts that can pass validation. I think this is the intention here?

angranl-flex: > this code is dangerous and error prone. Why do we need to clean up? Where in the workflow will we fail if we don't?

In the issue reported by Jim [Link](https://www.notion.so/flexcompute/Pending-Angran-velocity-required-when-requesting-meshing-only-131db96dc14e80dab030cd212d014ccf?pvs=4), pydantic refused to deserialize the simulation.json with an incomplete PorousMedium model from WebUI and raised validation error. 

If we want to pass the validation, we need to make all CASE related fields to be optional ConditionalField or we pop the keys unrelated to the validation level. The first way involves a lot of work and it may also confuse the user since all the fields become optional.

cc: @benflexcompute

benflexcompute: Yes I think popping is the way to go for the deserialization. @maciej-flexcompute What do you think?

PR #727, opened 2025-02-11
allow configuring shutter execution in parallel or serial

feilin-flexcompute:
since sometimes shutter can not be executed in parallel.


benflexcompute (commented):
~~Any penalty having this to always run in async mode? Any scenario where we can only do it in seral mode? If not can we just do that?~~

Why it is not possible to run parallel?

benflexcompute (approved):

PR #730, opened 2025-02-11
[SCFD-4271] Cleanup redundant keys in the json file in the constructor of SimulationParams

angranl-flex:
- clean redundant keys from webUI json
- clean unnecessary keys based on the validation level

### DIFF
--- a/flow360/component/simulation/services.py
+++ b/flow360/component/simulation/services.py
@@ -59,7 +59,6 @@ from flow360.component.simulation.validation.validation_context import (
     ParamsValidationInfo,
     ValidationContext,
 )
-from flow360.component.utils import remove_properties_by_name
 from flow360.exceptions import Flow360TranslationError
 
 unit_system_map = {
@@ -265,7 +264,8 @@ def validate_model(
     validation_warnings = None
     validated_param = None
 
-    params_as_dict = clean_params_dict(params_as_dict, root_item_type)
+    params_as_dict = clean_params_dict_for_validation(params_as_dict, root_item_type)
+
     # The final validation levels will be the intersection of the requested levels and the levels available
     # We always assume we want to run case so that we can expose as many errors as possible
     available_levels = _determine_validation_level(up_to="Case", root_item_type=root_item_type)
@@ -288,7 +288,7 @@ def validate_model(
     return validated_param, validation_errors, validation_warnings
 
 
-def clean_params_dict(params: dict, root_item_type: str) -> dict:
+def clean_params_dict_for_validation(params: dict, root_item_type: str) -> dict:
     """
     Cleans the parameters dictionary by removing unwanted properties.
 
@@ -304,8 +304,6 @@ def clean_params_dict(params: dict, root_item_type: str) -> dict:
     dict
         The cleaned parameters dictionary.
     """
-    params = remove_properties_by_name(params, "_id")
-    params = remove_properties_by_name(params, "hash")  # From client
 
     if root_item_type == "VolumeMesh":
         params.pop("meshing", None)
--- a/flow360/component/simulation/services.py
+++ b/flow360/component/simulation/services.py
@@ -264,7 +264,7 @@ def validate_model(
     validation_warnings = None
     validated_param = None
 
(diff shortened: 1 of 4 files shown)


flow360/component/simulation/simulation_params.py
```diff
@@ -143,6 +144,15 @@ def _update_input(cls, model_dict):
             )
         return model_dict
 
+    @classmethod
+    def _clean_params_dict_from_webui(cls, model_dict):
+        """
+        Clean the redundant content in the params dict from WebUI
+        """
+        model_dict = remove_properties_by_name(model_dict, "_id")
+        model_dict = remove_properties_by_name(model_dict, "hash")  # From client
```
maciej-flexcompute, line 153: why do we remove "hash"? Or why do we put hash? This seems be weird now. We add here:
https://github.com/flexcompute/Flow360/blob/1f257d70886109f20a895384448df3e3d063dbd2/flow360/component/simulation/simulation_params.py#L110
and here:
https://github.com/flexcompute/Flow360/blob/1f257d70886109f20a895384448df3e3d063dbd2/flow360/component/simulation/framework/base_model.py#L466
https://github.com/flexcompute/Flow360/blob/1f257d70886109f20a895384448df3e3d063dbd2/flow360/component/simulation/framework/base_model.py#L535

The current implementation supports:
1. using model with hash
2. using model without hash

What is the intention here? (I understand it was already in services.py but I want to understand why it was there in first place)


Also, I believe SimulationParams should not have function with "webui" in its name. I think services is a better place for it.

angranl-flex: I just double checked this removal of hash with Ben, I think the original intention is to 
1. Avoid the trigger of the following warning when loading a simulation json from WebUI, https://github.com/flexcompute/Flow360/blob/bb99e4c1db8b294a17c5058d45079f77d28f9636/flow360/component/simulation/framework/base_model.py#L544
2. Avoid validation error due to extra "hash" key, which is already be addressed in https://github.com/flexcompute/Flow360/blob/1f257d70886109f20a895384448df3e3d063dbd2/flow360/component/simulation/framework/base_model.py#L536

But it seems that the hash key is only exported when `to_file()` is called, and we want to trigger the above warning when user manually edit the json file. So this removal of hash will be deleted.

flow360/component/simulation/services.py
```diff
@@ -288,7 +288,7 @@ def validate_model(
     return validated_param, validation_errors, validation_warnings
 
 
-def clean_params_dict(params: dict, root_item_type: str) -> dict:
+def clean_params_dict_for_validation(params: dict, root_item_type: str) -> dict:
```
benflexcompute, line 291:
```suggestion
def clean_unrelated_setting_from_params_dict(params: dict, root_item_type: str) -> dict:
```

flow360/component/simulation/services.py
```diff
@@ -288,7 +288,7 @@ def validate_model(
...
     """
     Cleans the parameters dictionary by removing unwanted properties.
```
benflexcompute, line 293:
```suggestion
    Cleans the parameters dictionary by removing properties if they do not affect the remaining workflow.
```

flow360/component/simulation/simulation_params.py
```diff
@@ -147,6 +148,14 @@ def _update_param_dict(cls, model_dict):
             )
         return model_dict
 
+    @classmethod
+    def _clean_params_dict_from_service(cls, model_dict):
```
benflexcompute, line 152:
```suggestion
    def _sanitize_params_dict(cls, model_dict):
```

benflexcompute (approved):

benflexcompute (approved):

flow360/component/simulation/simulation_params.py
angranl-flex: Done

flow360/component/simulation/services.py
angranl-flex: Done

flow360/component/simulation/services.py
angranl-flex: updated

PR #736, opened 2025-02-14
changed omega dimensionalization for bet converter

wiktor-flex:
instead of time_unit as an input, now freestream_temperature is needed

fixed the unittest for bet converter as it was not working properly

for the example conversion the calculated value of omega is now very close to reference value from the tutorial
Tutorial states 460 RPM, calculated = 460.5687478679744 RPM but it most likely comes from simplifications made in the calculation of non-dimensional omega in the documentation

### DIFF
--- a/flow360/component/simulation/migration/BETDisk.py
+++ b/flow360/component/simulation/migration/BETDisk.py
@@ -113,7 +113,7 @@ def _parse_flow360_bet_disk_dict(
     )
 
     log.info("Provided temperature was used to calculate the value of omega in rad/s.")
-    log.info("Omega can be manually changed to a desired value and unit.")
+    log.info("You can print and correct the value and unit of `Omega` afterwards if needed.")
 
     updated_bet_dict["chord_ref"] = updated_bet_dict["chord_ref"] * mesh_unit
     updated_bet_dict["sectional_radiuses"] = updated_bet_dict["sectional_radiuses"] * mesh_unit
--- a/flow360/component/simulation/migration/BETDisk.py
+++ b/flow360/component/simulation/migration/BETDisk.py
@@ -5,16 +5,18 @@
 import json
 import os
 
+from numpy import sqrt
 from pydantic import validate_call
 
 import flow360.component.simulation.units as u
 from flow360.component.simulation.models.volume_models import BETDisk
 from flow360.component.simulation.primitives import Cylinder
-from flow360.component.simulation.unit_system import LengthType, TimeType
+from flow360.component.simulation.unit_system import LengthType, AbsoluteTemperatureType
+from flow360.log import log
 
 
 def _parse_flow360_bet_disk_dict(
-    *, flow360_bet_disk_dict: dict, mesh_unit, time_unit, bet_disk_index: int = 0
+    *, flow360_bet_disk_dict: dict, mesh_unit, freestream_temperature, bet_disk_index: int = 0
 ):
     """
     Read in the provided Flow360 BETDisk config.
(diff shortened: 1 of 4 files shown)


flow360/component/simulation/migration/BETDisk.py
```diff
@@ -34,6 +36,11 @@ def _parse_flow360_bet_disk_dict(
             raise ValueError("Input file does not contain BETDisk setting.")
         flow360_bet_disk_dict = flow360_bet_disk_dict["BETDisks"][0]
 
+    specific_heat_ratio = 1.4
```
benflexcompute, line 39: Can you import Air and use https://github.com/flexcompute/Flow360/blob/59bd005bf649c060eeb8a07b1deb35d8db09ddb1/flow360/component/simulation/models/material.py#L178
So there is less duplication?

flow360/component/simulation/migration/BETDisk.py
```diff
@@ -104,6 +111,10 @@ def _parse_flow360_bet_disk_dict(
     updated_bet_dict["omega"] = (
         updated_bet_dict["omega"] * u.rad / time_unit  # pylint: disable = no-member
     )
+
+    log.info("Provided temperature was used to calculate the value of omega in rad/s.")
+    log.info("Omega can be manually changed to a desired value and unit.")
```
benflexcompute, line 116: It won't be possible to change since the V0 Flow360.json has
```python
{
...
"omega":0.2
...
}
```
so there is not option for changing unit and also we want to divert user from hand writing JSON directly.

tests/simulation/converter/ref/ref_single_bet_disk.json
```diff
@@ -463,7 +462,7 @@
     "number_of_blades": 3,
     "omega": {
         "units": "rad/s",
-        "value": 0.0023
```
benflexcompute, line 466: Why this change was not detected previously by unit test? Do we have deficiency in unit test?

tests/simulation/converter/test_bet_disk_flow360_converter.py
```diff
@@ -19,16 +19,23 @@ def change_test_dir(request, monkeypatch):
...
+def test_single_flow360_bet_convert(atol=1e-15, rtol=1e-10, debug=False):
     disk = read_single_v1_BETDisk(
         file_path="./data/single_flow360_bet_disk.json",
         mesh_unit=1 * u.cm,
-        time_unit=2 * u.s,
+        freestream_temperature=288.15 * u.K,
     )
     assert isinstance(disk, BETDisk)
+    disk = disk.model_dump_json()
+    disk = json.loads(disk)
```
benflexcompute, line 30: you can do `disk = disk.model_dump(mode="json")` instead

wiktor-flex: Thanks, haven't thought about changing the mode

tests/simulation/converter/ref/ref_single_bet_disk.json
wiktor-flex: Because previously the compare values was not working properly as it had no assert to it
`compare_values(disk.model_dump(), ref_dict)`

flow360/component/simulation/migration/BETDisk.py
wiktor-flex: Good idea

flow360/component/simulation/migration/BETDisk.py
wiktor-flex: The idea behind this info was to inform the user that after their BET was translated, they can manually set up the omega value and unit they want as showcased in example_bet_conversion.py
`"""
params.models[X].omega = 500 * fl.u.rpm
"""`

wiktor-flex: I can delete line 116 which states that omega can be changed if you want

benflexcompute: I see what you mean. I thought you are asking user to change the JSON. Let's change the 
```suggestion
    log.info("You can print and correct the value and unit of `Omega` afterwards if needed.")
```
for less confusion.

benflexcompute (approved):

PR #738, opened 2025-02-14
[FL-226] TimeAverageProbeOutput supports frequency=-1

angranl-flex:


flow360/component/simulation/outputs/outputs.py
```diff
@@ -613,8 +613,19 @@ class TimeAverageProbeOutput(ProbeOutput):
 
     """
 
-    frequency: int = pd.Field(default=1, ge=1)
-    frequency_offset: int = pd.Field(default=0, ge=0)
+    # pylint: disable=abstract-method
+    frequency: int = pd.Field(
+        default=-1,
```
yifan-flex, line 618: Default should be 1, we used to always output every time step before this becomes an adjustable parameter.

flow360/component/simulation/outputs/outputs.py
```diff
@@ -691,8 +702,18 @@ class TimeAverageSurfaceProbeOutput(SurfaceProbeOutput):
     """
 
     # pylint: disable=abstract-method
-    frequency: int = pd.Field(default=1, ge=1)
-    frequency_offset: int = pd.Field(default=0, ge=0)
+    frequency: int = pd.Field(
+        default=-1,
```
yifan-flex, line 706: I think default here should be 1 too to be consistent.

flow360/component/simulation/outputs/outputs.py
angranl-flex: I think we currently have default frequency =-1 here: https://github.com/flexcompute/Flow360/blob/4e0aa69c8d9fe5ea831748caebb92c972706db5a/flow360/component/simulation/outputs/outputs.py#L78

Should we change it to 1?

angranl-flex: Update to 1 now

flow360/component/simulation/outputs/outputs.py
angranl-flex: Update to 1 now

yifan-flex (approved):

PR #743, opened 2025-02-18
[Hotfix-24.11] Remove DDES settings in periodic BC tutorial

angranl-flex:
#733


angranl-flex:
> Have you checked if the DDES and grid_size_for_LES has default and the validation can pass?

Yes, both of them has default values and I can run the case in webUI.
image

benflexcompute (approved):

PR #750, opened 2025-02-20
Add length unit property to project

angranl-flex:


flow360/component/project.py
```diff
@@ -449,6 +449,28 @@ def id(self) -> str:
...
         return self.metadata.id
 
+    @property
+    def length_unit(self) -> pd.PositiveFloat:
+        """
+        Returns the length unit of the project.
+
+        Returns
+        -------
+        pd.PositiveFloatr
```
benflexcompute, line 459: Why is it float? Shouldn't it have dimension?

flow360/component/project.py
```diff
@@ -449,6 +449,28 @@ def id(self) -> str:
...
+            The length unit.
+        """
+
+        defaults = self._root_simulation_json
+
+        cache_key = "private_attribute_asset_cache"
+        length_key = "project_length_unit"
+
+        if cache_key not in defaults:
```
benflexcompute, line 468: The if logic is conflicting here

flow360/component/project.py
```diff
@@ -449,6 +449,28 @@ def id(self) -> str:
...
+            if length_key not in defaults[cache_key]:
+                raise Flow360ValueError("Simulation params do not contain default length unit info")
```
benflexcompute, line 470: Mark this internal error by prefix "[Internal]"

flow360/component/project.py
angranl-flex: Fixed

flow360/component/project.py
angranl-flex: Done

angranl-flex:
> Can you try quick_example.py and see if it works?

Just tried  quick_start.py and it works.

PR #755, opened 2025-02-20
Add pytest-xdist to improve CI speed

angranl-flex:


benflexcompute:
Does not improve speed.

PR #761, opened 2025-02-24
Add validation before draft submission.

angranl-flex:


flow360/component/project.py
```diff
@@ -27,20 +27,13 @@
     GeometryFiles,
     SurfaceMeshFile,
     VolumeMeshFile,
-    replace_ghost_surfaces,
+    set_up_params_for_draft,
```
benflexcompute, line 30:
```suggestion
    set_up_params_for_uploading,
```

flow360/component/project.py
```diff
@@ -1017,10 +965,29 @@ def _run(
...
         """
 
-        with model_attribute_unlock(params.private_attribute_asset_cache, "project_length_unit"):
-            params.private_attribute_asset_cache.project_length_unit = self.length_unit
+        params = set_up_params_for_draft(
+            params=params, root_asset=self._root_asset, length_unit=self.length_unit
+        )
 
-        root_asset = self._root_asset
+        params, errors = validate_params_for_draft_submission(
```
benflexcompute, line 972:
```suggestion
        params, errors = validate_params_with_context(
```

flow360/component/project.py
```diff
@@ -1017,10 +965,29 @@ def _run(
...
+            params=params,
+            root_item_type=self.metadata.root_item_type.value,
+            up_to=target._cloud_resource_type_name,
+        )
+
+        if errors is not None:
+            error_msg = ""
+            for idx, error in enumerate(errors):
```
benflexcompute, line 980: Can we warp this in a function?
We need to print the backend returned errors with the same formatting (once backend is ready).

flow360/component/project.py
```diff
@@ -1017,10 +965,29 @@ def _run(
...
+                error_msg += f"\n\t({idx+1}) Message: {error['msg']}"
+                if error.get("loc") != ():
```
benflexcompute, line 982: Would be best to have a Pydantic model for this or maybe this is the exact schema for pd.ValidationError? We can use the response as is but it is just not elegant as using a Pydantic model.

benflexcompute: Not required for now.

flow360/component/project_utils.py
```diff
@@ -255,3 +270,89 @@ def _find_ghost_surfaces(*, model, ghost_entities_from_metadata):
     _find_ghost_surfaces(model=params, ghost_entities_from_metadata=ghost_entities_from_metadata)
 
     return params
+
+
+def _set_up_param_entity_info(entity_info, params: SimulationParams):
```
benflexcompute, line 275: Let's not make this a single function. It does not make sense to mis non-persistent entity processing with the draft entities.

flow360/component/project.py
benflexcompute: Also in the new function docstring have an example of how the printed error message will look like.

flow360/component/project.py
benflexcompute: No need if it is just dict.

benflexcompute (approved):

PR #764, opened 2025-02-25
[Hotfix-24.11][SCFD-4474] Add validation before draft submission (#761)

angranl-flex:


benflexcompute (approved):
Are you able to submit a case?

angranl-flex:
> Are you able to submit a case?

Yes, I tested with the tutorial code for CHT solver. I will test the quick_start again before merging.

PR #774, opened 2025-02-27
Replaced the old updater mechanism on 24.11

benflexcompute:


angranl-flex (commented):
The unit test for the Flow360Version in updater_utils.py is missing: 

https://github.com/flexcompute/Flow360/blob/develop/tests/simulation/framework/test_version_parser.py

benflexcompute:
> The unit test for the Flow360Version in updater_utils.py is missing:
> 
> https://github.com/flexcompute/Flow360/blob/develop/tests/simulation/framework/test_version_parser.py

Good catch

angranl-flex (approved):

PR #775, opened 2025-02-27
[HOTFIX] add updater to 24.11.10 to 25.2

benflexcompute:


examples/tutorials/quick_start.py
```diff
@@ -7,6 +7,7 @@
 project = fl.Project.from_file(
     files=Airplane.geometry,
     name="Python Project (Geometry, from file)",
+    solver_version="release-24.11.25",
```
angranl-flex, line 10: Why adding this?

benflexcompute: good catch, will remove

angranl-flex (approved):

PR #785, opened 2025-03-05
[SCFD-4498] Formatting error msg from backend

angranl-flex:


flow360/component/simulation/web/draft.py
```diff
@@ -158,7 +160,9 @@ def run_up_to_target_asset(
             log.error(">>Submission failed.<<")
             try:
                 detailed_error = json.loads(err.auxiliary_json["detail"])["detail"]
-                log.error(f"Failure detail: {detailed_error}")
+                log.error(
+                    f"Failure detail: {formatting_validation_errors(ast.literal_eval(detailed_error))}"
```
benflexcompute, line 164: can you inside formatting_validation_errors add a blank line or some seperator like "-------- ERROR -------". Currently the error message in print out does not have enough separation from the other error stacks.

benflexcompute (approved):

benflexcompute:
Add `raise_on_error` option of these. Default behavior should be to silently return None.

benflexcompute (approved):

PR #799, opened 2025-03-10
Fix "_id" not removed from file_content init and added unit test

angranl-flex:


tests/simulation/service/test_services_v2.py
benflexcompute, line 1: Will this error out otherwise?

benflexcompute (approved):

tests/simulation/service/test_services_v2.py
angranl-flex: Yes, it will error out.

PR #817, opened 2025-03-13
[SCFD-4591] Add face/edge renaming function

angranl-flex:


benflexcompute:
Once https://github.com/flexcompute/Flow360/pull/829 is merged, can you add in example/ an example/test script to ensure the renaming workflow works?

benflexcompute:
Also we may need to consider moving the bodies not directly in the `geometry[""]` but instead geometry.bodyGroups[""] instead? This avoids ambiguity when user use `geometry["body0001"]` when using bodyId and faceId as grouping.

PR #824, opened 2025-03-14
[SCFD-4744] added access to mesh bounding box data

maciej-flexcompute:
https://flow360.atlassian.net/browse/SCFD-4744

Access to bounding box data for volume mesh:
- [x] access to filtering option to include / exclude boundaries
- [x] integrated with report API


flow360/plugins/report/utils.py
```diff
@@ -344,6 +345,44 @@ def calculate(
...
+        self, data, case, cases, variables, new_variable_name
+    ):  # pylint: disable=too-many-arguments
+        """
+        Getting attribute on the provided data.
+        """
+
+        try:
+            result = getattr(data, self.attr_name)
+        except AttributeError as err:
+            raise AttributeError(f"Attibute {self.attr_name} not found in {type(data)=}") from err
```
angranl-flex, line 381:
```suggestion
            raise AttributeError(f"Attribute {self.attr_name} not found in {type(data)=}") from err
```

PR #825, opened 2025-03-14
added content to TurbulenceQuantities docstring

wiktor-flex:
added content to docstring from https://docs.flexcompute.com/projects/flow360/en/beta-v2.0/knowledgeBase/PreProcessing/BoundaryConditions/BoundaryConditions.html so that this documentation page can be deleted

### DIFF
--- a/flow360/component/simulation/models/turbulence_quantities.py
+++ b/flow360/component/simulation/models/turbulence_quantities.py
@@ -203,7 +203,7 @@ def TurbulenceQuantities(
     turbulent_length_scale=None,
     turbulent_intensity=None,
 ) -> TurbulenceQuantitiesType:
-    """
+    r"""
 
     :func:`TurbulenceQuantities` function specifies turbulence conditions
     for the :class:`~flow360.Inflow` or :class:`~flow360.Freestream`
--- a/flow360/component/simulation/models/turbulence_quantities.py
+++ b/flow360/component/simulation/models/turbulence_quantities.py
@@ -209,7 +209,7 @@ def TurbulenceQuantities(
     at boundaries. The turbulence properties that can be
     specified are listed below. All values are dimensional.
     For valid specifications as well as the default values,
-    please refer to :ref:`knowledge base<knowledgeBaseTurbulenceQuantities>`.
+    please see the `Valid combinations` section below.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


flow360/component/simulation/models/turbulence_quantities.py
```diff
@@ -251,6 +251,43 @@ def TurbulenceQuantities(
     ValueError
         If the TurbulenceQuantities inputs do not represent a valid specification.
 
+
+    Valid combinations
+    ------------------
```
angranl-flex, line 256:
```suggestion
    Notes
    -----

    The valid combinations of multiple turbulence quantities is summarized as follows,

```

I would suggest using the existing numpy docstring section keyword like "Notes" as title. The current documentation format is not uniform with other sections: 
Current

flow360/component/simulation/models/turbulence_quantities.py
```diff
@@ -251,6 +251,43 @@ def TurbulenceQuantities(
...
+        For SST model only.
+    turbulent_kinetic_energy or turbulent_intensity with viscosity_ratio
+        For SST model only.
+    turbulent_kinetic_energy or turbulent_intensity with turbulent_length_scale
+        For SST model only.
+    specific_dissipation_rate with viscosity_ratio
+        For SST model only.
+    specific_dissipation_rate with turbulent_length_scale
+        For SST model only.
+    viscosity_ratio with with turbulent_length_scale
```
angranl-flex, line 288: Seems to be missing "For SST model only"?

flow360/component/simulation/models/turbulence_quantities.py
```diff
@@ -251,6 +251,43 @@ def TurbulenceQuantities(
...
+
+
     Example
```
angranl-flex, line 291: Can you also add some examples to demonstrate the combination of different quantities? I do remember Jim or Roberto mentioned that we should demonstrate more function parameters/class attributes in the examples. You can pick 1 or 2 combinations from the section you added. What do you think?

flow360/component/simulation/models/turbulence_quantities.py
```diff
@@ -251,6 +251,43 @@ def TurbulenceQuantities(
...
+    ------------------
+    default
+        The default turbulence depends on the turbulence model.
+        For SA model without transition model this is equivalent to set
+        modifiedTurbulentViscosityRatio = 3.0 (or effectively turbulentViscosityRatio = 0.210438).
+        For SA model with transition model, modifiedTurbulentViscosityRatio = 0.1
+        (or effectively turbulentViscosityRatio = 2.794e-7). For SST model the default turbulence is
+        turbulentViscosityRatio = 0.01 with default specificDissipationRate = reference_mach/l_box
+        where l_box is the bounding box dimension for wall boundaries.
+    viscosity_ratio alone
```
angranl-flex, line 265: Maybe use :code: format to distinguish the parameter with the normal text?

flow360/component/simulation/models/turbulence_quantities.py
```diff
@@ -251,6 +251,43 @@ def TurbulenceQuantities(
...
+    Valid combinations
+    ------------------
+    default
+        The default turbulence depends on the turbulence model.
+        For SA model without transition model this is equivalent to set
+        modifiedTurbulentViscosityRatio = 3.0 (or effectively turbulentViscosityRatio = 0.210438).
+        For SA model with transition model, modifiedTurbulentViscosityRatio = 0.1
+        (or effectively turbulentViscosityRatio = 2.794e-7). For SST model the default turbulence is
+        turbulentViscosityRatio = 0.01 with default specificDissipationRate = reference_mach/l_box
+        where l_box is the bounding box dimension for wall boundaries.
```
angranl-flex, line 264: Will you include the latex format equation from the original table here? I know you convert it to normal text now but it does not look good right now. Can you try copy the equation from the table's csv file?

angranl-flex (commented):
@wiktor-flex I guess you have not finished this PR yet but just left some comments. Thank you for taking care of this!

flow360/component/simulation/models/turbulence_quantities.py
wiktor-flex: I will try that

flow360/component/simulation/models/turbulence_quantities.py
wiktor-flex: Good idea

flow360/component/simulation/models/turbulence_quantities.py
wiktor-flex: Sure

flow360/component/simulation/models/turbulence_quantities.py
wiktor-flex: True

flow360/component/simulation/models/turbulence_quantities.py
```diff
@@ -251,16 +252,71 @@ def TurbulenceQuantities(
...
+        For SA model *with transition model*, :code:`modified_viscosity_ratio = 0.1`
+        (or effectively :code:`viscosity_ratio = 2.794e-7`). For SST model the default turbulence is
+        :code:`viscosity_ratio = 0.01` with default :code:`specific_dissipation_rate` = :math:`MachRef/L_{box}`
+        where :math:`L_{box} \triangleq exp\left(\displaystyle\sum_{i=1}^{3}log(x_{i,max}-x_{i,min}\right)`.
+        :math:`x_{i,max},x_{i,min}` is the bounding box dimension for wall boundaries.
+    :code:`viscosity_ratio` alone
+        This applies to both SST and SA model. For SST model this is effectively
+        an override of the above default :code:viscosity_ratio value while keeping
+        the default specificDissipationRate. For SA model the :code:viscosity_ratio
+        will be converted to the :code:modified_viscosity_ratio.
```
angranl-flex, line 273:
```suggestion
        This applies to both SST and SA model. For SST model this is effectively
        an override of the above default :code:`viscosity_ratio` value while keeping
        the default specificDissipationRate. For SA model the :code:`viscosity_ratio`
        will be converted to the :code:`modified_viscosity_ratio`.
```

PR #843, opened 2025-03-19
[SCFD-4765] Support mixture of geometry and surface mesh files for from_geometry()

angranl-flex:


flow360/component/geometry.py
```diff
@@ -166,9 +166,9 @@ def submit(self, description="", progress_callback=None) -> Geometry:
             files=[
                 GeometryFileMeta(
                     name=os.path.basename(file_path),
-                    type="main" if item_index == 0 else "dependency",
+                    type="main",
```
benflexcompute, line 169: Could you add comment here explaining the current main vs dependency meaning/usage?

flow360/component/utils.py
```diff
@@ -801,29 +809,43 @@ class GeometryFiles(InputFileModel):
     @pd.field_validator("file_names", mode="after")
     @classmethod
     def _validate_files(cls, value):
-        if isinstance(value, str):
+        supported_geometry_surfacemesh_file = SUPPORTED_GEOMETRY_FILE_PATTERNS.copy() + [
+            MeshFileFormat.UGRID.ext(),
```
benflexcompute, line 813: Are these 3 the only 3 supported? Does upload support all format defined in `SurfaceMeshFiles`?

flow360/component/utils.py
```diff
@@ -801,29 +809,43 @@ class GeometryFiles(InputFileModel):
     @pd.field_validator("file_names", mode="after")
     @classmethod
     def _validate_files(cls, value):
-        if isinstance(value, str):
+        supported_geometry_surfacemesh_file = SUPPORTED_GEOMETRY_FILE_PATTERNS.copy() + [
```
benflexcompute, line 812: why copy()?

flow360/component/utils.py
```diff
@@ -801,29 +809,43 @@ class GeometryFiles(InputFileModel):
...
+            MeshFileFormat.UGRID.ext(),
+            MeshFileFormat.CGNS.ext(),
+            MeshFileFormat.STL.ext(),
+        ]
+
+        def _validate_single_file(value):
             if not match_file_pattern(SUPPORTED_GEOMETRY_FILE_PATTERNS, value):
```
benflexcompute, line 819: Why the validation of geometry and surface mesh are nested? Can we check geometry format and then surface mesh format instead?

flow360/component/utils.py
```diff
@@ -801,29 +809,43 @@ class GeometryFiles(InputFileModel):
...
-                raise ValueError(
-                    f"The given file: {value} is not a supported geometry file. "
-                    f"Allowed file suffixes are: {SUPPORTED_GEOMETRY_FILE_PATTERNS}"
-                )
+                try:
+                    # pylint: disable=protected-access
+                    SurfaceMeshFile._validate_files(value=value)
+                except ValueError as err:
+                    raise ValueError(
+                        f"The given file: {value} is not a supported geometry file. "
```
benflexcompute, line 825:
```suggestion
                        f"The given file: {value} is not a supported surface mesh file. "
```

flow360/component/utils.py
```diff
@@ -801,29 +809,43 @@ class GeometryFiles(InputFileModel):
...
     @classmethod
     def _validate_files(cls, value):
-        if isinstance(value, str):
+        supported_geometry_surfacemesh_file = SUPPORTED_GEOMETRY_FILE_PATTERNS.copy() + [
+            MeshFileFormat.UGRID.ext(),
+            MeshFileFormat.CGNS.ext(),
+            MeshFileFormat.STL.ext(),
+        ]
+
+        def _validate_single_file(value):
```
benflexcompute, line 818: Add docstring reiterate that we are accepting/validating both geometry and surface mesh files.

flow360/component/utils.py
```diff
@@ -801,29 +809,43 @@ class GeometryFiles(InputFileModel):
...
         return value
 
+    def _check_files_existence(self) -> None:
+        """
+        Check if the file exists or not. If it is ugrid file then check existence of mapbc file.
+        """
+        super()._check_files_existence()
+        # pylint: disable=not-an-iterable
+        for file_name in self.file_names:
+            _check_mapbc_existence(value=file_name)
```
benflexcompute, line 843: Can we verify that the mapbc is uploaded and will change the boundary names accordingly?

angranl-flex: Yes, I have modified the code to upload mapbc and verified the boundary names are changed accordingly on WebUI

flow360/component/geometry.py
benflexcompute: So "dependency" is not needed anymore? I forgot what was the conclusion.

PR #857, opened 2025-03-21
[SCFD-4590] Add get_tranformation_matrix

yifan-flex:


flow360/component/simulation/primitives.py
```diff
@@ -96,19 +96,48 @@ class Transformation(Flow360BaseModel):
     """Transformation that will be applied to a body group."""
 
     type_name: Literal["BodyGroupTransformation"] = pd.Field("BodyGroupTransformation", frozen=True)
+
+    # pylint:disable=no-member
```
benflexcompute, line 100: Can we not put this at the root level?

Let's only add pylint disable when needed. Adding here will blind us of any real no-member error at this indent level.

flow360/component/simulation/primitives.py
```diff
@@ -96,19 +96,48 @@ class Transformation(Flow360BaseModel):
...
+    translation: LengthType.Point = pd.Field((0, 0, 0) * u.m)
 
     private_attribute_matrix: Optional[list[float]] = pd.Field(None)
 
-    def get_transformation_matrix(self):
-        """WIP, need to check with front end"""
-        return None
+    def get_transformation_matrix(
+        self,
+        project_length_unit: LengthType.Positive = pd.Field(
```
benflexcompute, line 118: Is this valid syntax? Would `project_length_unit: LengthType.Positive` be sufficient?

This to me looks like assigning a default value to project_length_unit with a pd.Field().

flow360/component/simulation/primitives.py
```diff
@@ -96,19 +96,48 @@ class Transformation(Flow360BaseModel):
...
+            description="The Length unit of the project."
+        ),
+    ) -> np.ndarray:
+        """Find 3x4 transformation matrix and store as row major. Last column applies to project length unit"""
+        origin_in_length_unit = np.asarray(self.origin / project_length_unit)
```
benflexcompute, line 123: How do we guard against accidental call after non-dimensionalization? What if self has been non-dimensionalized for example when used in translator?

I know we do not use it in translator for now but we should prevent future misuse.

tests/simulation/params/test_simulation_params.py
```diff
@@ -429,3 +430,14 @@ def test_geometry_entity_info_to_file_list_and_entity_to_file_map():
             "farfield_only_sphere_volume_mesh.lb8.ugrid": "farfield_only_sphere_volume_mesh.lb8.ugrid",
         }.items()
     )
+
+    transformation_matrix = geometry_entity_info.grouped_bodies[0][
```
benflexcompute, line 434: Have we verified that the matrix we computed is the same as what frontend team computes?

benflexcompute:
Also a general comment would be include JIRA ticket number in your PR title. This link with JIRA and help everyone track ticket progress easily. Also this provides context of why the PR is sent in the first place. Thank you!

PR #862, opened 2025-03-25
Add toggle for GAI

angranl-flex:


flow360/component/project.py
```diff
@@ -1197,11 +1198,15 @@ def _run(
             root asset (Geometry or VolumeMesh) is not initialized.
         """
 
+        if use_geometry_AI and not use_beta_mesher:
+            raise Flow360ValueError("Enabling Geometry AI requires beta mesher.")
```
benflexcompute, line 1202:
```suggestion
            raise Flow360ValueError("Enabling Geometry AI requires also enabling beta mesher.")
```

benflexcompute (approved):

flow360/component/project.py
```diff
@@ -1289,7 +1300,7 @@ def generate_surface_mesh(
         name: str = "SurfaceMesh",
         run_async: bool = True,
         solver_version: str = None,
-        use_beta_mesher: bool = False,
+        use_beta_mesher: bool = None,
```
benflexcompute, line 1303: Does this pass   @pd.validate_call ? Or should we do 
```suggestion
        use_beta_mesher: Optional[bool] = None,
```

benflexcompute (approved):

flow360/component/project.py
benflexcompute: Ok seems solver version is also None without optional and it worked...

PR #866, opened 2025-03-26
Update the quickstart tutorial with new eVTOL geometry

malhardp333:
Updating the quick start tutorial here so that a reference to this can be used in the documentation.


examples/tutorials/quick_start.py
```diff
@@ -1,63 +1,93 @@
...
+            fl.Wall(surfaces=geometry["v_tail"], name="v_tail"),
+            fl.Freestream(surfaces=[farfield_zone.farfield], name="Freestream"),
         ],
-        # Define output parameters for the simulation
+        # Output format could be 'paraview' or 'tecplot' or 'both'.
         outputs=[
             fl.SurfaceOutput(
-                surfaces=geo["*"],  # Select all surfaces for output
-                output_fields=["Cp", "Cf", "yPlus", "CfVec"],  # Output fields for post-processing
+                surfaces=geometry["*"],
```
angranl-flex, line 68: Add a comment to explain the usage of wildcard here.

examples/tutorials/quick_start.py
```diff
@@ -1,63 +1,93 @@
...
+                fl.SurfaceEdgeRefinement(
+                    name="leading_edges",
+                    edges=[geometry["leadingEdge"]],
+                    method=fl.AngleBasedRefinement(value=2 * fl.u.deg),
+                )
+            ],
         ),
-        # Reference geometry parameters for the simulation (e.g., center of pressure)
+        # Set the reference geometry.
         reference_geometry=fl.ReferenceGeometry(),
```
angranl-flex, line 46: I know that we are using the default value in the ReferenceGeometry class. Can we still explicitly set the default values for reference center, length and area to demonstrate how to use this class?

examples/tutorials/quick_start.py
```diff
@@ -1,63 +1,93 @@
...
-        # Operating conditions: setting speed and angle of attack for the simulation
+        # Set the operating condition.
         operating_condition=fl.AerospaceCondition(
```
angranl-flex, line 48: Same comment here, should we demonstrate more options in the AerospaceCondition class even we are using the default values here?

examples/tutorials/quick_start.py
malhardp333: Updated this

PR #877, opened 2025-04-03
feat(): Add support of t_seconds to AngleExpression

benflexcompute:


yifan-flex (approved):

flow360/component/simulation/framework/expressions.py
```diff
@@ -9,3 +11,97 @@
...
+    """
+    ALLOWED_NAMES = {  # pylint:disable=invalid-name
+        "t_seconds",
+        "t",
+        "pi",
+        "sin",
+        "cos",
+        "tan",
+        "atan",
+    }
```
angranl-flex, line 51: Are these all the names we allowed? What if the user want to use something like abs, pow, sqrt, etc.?

benflexcompute: added all the functions that are allowed on case pipeline side

PR #885, opened 2025-04-09
added drivaer notebook

wiktor-flex:
notebook based on the automotive_report.py script

it includes an example 5.7m mesh as well as all parameters to run the case and generate the report

### DIFF
--- a/examples/tutorials/notebooks/notebook_automotive.ipynb
+++ b/examples/tutorials/notebooks/notebook_automotive.ipynb
@@ -366,7 +366,7 @@
   },
   {
    "cell_type": "code",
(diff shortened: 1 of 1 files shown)


piotrkluba:
Reset the cell outputs before saving the file or run it from the top so the outputs seen in the file are those to be expected
Zrzut ekranu 2025-04-18 100801

piotrkluba:
`cases = []` from cell 4 can be moved to cell 7

piotrkluba (commented):
- Text before cell 4 was not modified when `cases = []` was moved

piotrkluba (approved):
OK

PR #887, opened 2025-04-10
Fix local report generation on Windows

piotrkluba:
Fixed the local report generation error on Windows due to the Windows style paths from os.path interfering with latex syntax.

### DIFF
--- a/tests/report/test_report.py
+++ b/tests/report/test_report.py
@@ -1,4 +1,5 @@
 import os
+import re
 
 import pytest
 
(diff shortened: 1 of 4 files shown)


wiktor-flex (commented):
Fix black and isort, the rest seems okay

tests/report/test_report.py
```diff
@@ -133,3 +135,19 @@ def test_reporttemplate_no_items():
...
+def test_filepaths_format():
+    report_doc = ReportDoc(title="test_doc")
+    tex = report_doc.doc.dumps()
+    lines = tex.split("\n")
+    for line in lines:
+        line = line.lstrip()
+        if line.startswith(r"\includegraphics"):
+            path_match = re.search(r"\{(.*?)\}", line)
+            if path_match:
+                path: str = path_match.group(1)
```
wiktor-flex, line 150: Does specifying type here do anything? It will always be a string and type hinting is unnecessary

piotrkluba: Fair, it made it easier for me to look through string methods. I will remove it.

maciej-flexcompute (approved):

PR #888, opened 2025-04-10
Added complete list of usable solver variables

andrzej-krupka:


andrzej-krupka:
Merging to the feature branch, unit test issues will be fixed in a separate PR.

PR #890, opened 2025-04-11
Fix a bug in generating captions for 2D charts

piotrkluba:
Fixed a bug where report generation crashed when adding a 2D chart with items_in_row parameter, corrected the main caption when specifying selected_indices to be ".... for selected cases" rather than "... for all cases"


wiktor-flex (approved):
Looks good

maciej-flexcompute (approved):

PR #893, opened 2025-04-11
[SCFD-5035] Add default names for BCs

angranl-flex:


maciej-flexcompute (approved):

flow360/component/simulation/models/surface_models.py
maciej-flexcompute, line 1: @wiktor-flex can you update sweep tutorial to remove BC names?

wiktor-flex: Will do

PR #897, opened 2025-04-11
[SCFD-4891] Add validation for isosurface name

angranl-flex:


flow360/component/simulation/simulation_params.py
```diff
@@ -386,6 +387,12 @@ def _disable_expression_for_liquid(cls, value, info: pd.ValidationInfo):
             )
         return value
 
+    @pd.field_validator("outputs", mode="after")
+    @classmethod
+    def check_duplicate_output_entites_names(cls, outputs):
+        """Cehck if we have isosurfaces, slices or monitors with a duplicate name"""
```
maciej-flexcompute, line 393: typo in Check
also, I think we check for duplicates only in isosurfaces, not in slices and monitors.

PR #899, opened 2025-04-14
[SCFD-4512] Enforce unit conversion output json to use front-end supported unit string.

benflexcompute:


tests/simulation/service/test_services_v2.py
```diff
@@ -741,14 +742,18 @@ def test_imperial_unit_system_conversion():
     # General comparison\
     with open("./ref/unit_system_converted_imperial.json", "r") as fp:
         ref_dict = json.load(fp)
+
+    with open("./DEBUG.json", "w") as fp:
+        json.dump(dict_to_convert, fp, indent=4)
```
angranl-flex, line 747: Remove the debug code.

tests/simulation/service/test_services_v2.py
```diff
@@ -822,3 +827,41 @@ def test_updater_service():
         errors[0]
         == "Input `SimulationParams` have higher version than the target version and thus cannot be handled."
     )
+
+
+def test_unit_conversion_front_end_compatibility():
+    from flow360.component.simulation.exposed_units import supported_units_by_front_end
+    from flow360.component.simulation.unit_system import _PredefinedUnitSystem
```
angranl-flex, line 834: Should we put them import at the beginning?

flow360/component/simulation/exposed_units.py
```diff
@@ -39,3 +39,61 @@
     "length": [u.m, u.cm, u.mm, u.ft, u.inch],
     "temperature": [u.K, "degC", "degF", u.R],  # Cannot use u.degC since it __str__() to "\u00b0C"
 }
+
+supported_units_by_front_end = {
+    "(mass)": {"SI": "kg", "CGS": "g", "Imperial": "lb"},  # Not used in SimulationParams
+    "(length)*(mass)/(time)**2": {
+        "SI": "N",
+        "CGS": "dyne",
```
angranl-flex, line 47: Is there any difference between "dyn" and "dyne"? Seems that we have both used in the supported units.
Or it is just the unit that used by the front end?

tests/simulation/service/test_services_v2.py
benflexcompute: Done

angranl-flex (approved):

PR #900, opened 2025-04-15
removed explicit name assignments for bcs and outputs in examples

wiktor-flex:
There are also some minor fixes like removing dev env and using Project.from_cloud instead of normal upload


piotrkluba (approved):
Very good - I don't see anything wrong

PR #901, opened 2025-04-15
added default name fields to all classes that appear on webUI

wiktor-flex:
some of the classes are python only for now, but I still added default names for them


piotrkluba (approved):
OK

PR #903, opened 2025-04-15
Plotting different variables in the plot tool made possible

piotrkluba:
Additionally: implemented a Residuals() class responsible for automaticly plotting the residuals from different cases, added the grid control to the user interface, ensured correct plotting of any nonlinear residual

### DIFF
--- a/flow360/plugins/report/report_items.py
+++ b/flow360/plugins/report/report_items.py
@@ -889,8 +889,8 @@ class Chart2D(Chart):
     ----------
     x : Union[str, Delta]
         The data source for the x-axis, which can be a string path or a `Delta` object.
-    y : Union[str, Delta]
-        The data source for the y-axis, which can be a string path or a `Delta` object.
+    y : Union[str, Delta, List[str]]
+        The data source for the y-axis, which can be a string path or their list or a `Delta` object.
     background : Union[Literal["geometry"], None], optional
         Background type for the chart; set to "geometry" or None.
     _requirements : List[str]
@@ -908,10 +908,12 @@ class Chart2D(Chart):
     ylim : Optional[Union[ManualLimit, SubsetLimit, FixedRangeLimit, Tuple[float, float]]]
         Defines the range of y values that will be displayed on the chart.
         This helps with highlighting a desired portion of the chart.
+    y_log : Optional[bool] 
+        Sets the y axis to logarithmic scale.
     """
 
     x: Union[str, Delta, DataItem]
-    y: Union[str, Delta, DataItem]
+    y: Union[str, Delta, DataItem, List[str]]
     background: Union[Literal["geometry"], None] = None
     _requirements: List[str] = [_requirements_mapping["total_forces"]]
     type_name: Literal["Chart2D"] = Field("Chart2D", frozen=True)
@@ -929,6 +931,7 @@ class Chart2D(Chart):
     ] = None
     xlim: Optional[Union[ManualLimit, Tuple[float, float]]] = None
     ylim: Optional[Union[ManualLimit, SubsetLimit, FixedRangeLimit, Tuple[float, float]]] = None
+    y_log:  Optional[bool] = False
 
     def get_requirements(self):
         """
@@ -938,18 +941,13 @@ class Chart2D(Chart):
 
     def is_log_plot(self):
         """
-        Determines if the plot is logarithmic based on the data path of the y-axis.
+        Determines if the plot is logarithmic.
 
         Returns
         -------
         bool
-            True if the root path of `self.y` corresponds to "nonlinear_residuals"
-            and "linear_residuals", indicating a logarithmic plot; False otherwise.
         """
-        root_path = get_root_path(self.y)
-        return root_path.startswith("nonlinear_residuals") or root_path.startswith(
-            "linear_residuals"
-        )
+        return self.y_log
 
     # pylint: disable=unpacking-non-sequence
     @pd.model_validator(mode="after")
@@ -999,7 +997,7 @@ class Chart2D(Chart):
         return False
 
     def _handle_data_with_units(self, x_data, y_data, x_label, y_label):
-
+        # TODO: manage when different variables are plotted
         if self._check_dimensions_consistency(x_data) is True:
             x_unit = x_data[0].units
             x_data = [data.value for data in x_data]
@@ -1019,10 +1017,23 @@ class Chart2D(Chart):
 
     def _load_data(self, cases):
         x_label = split_path(self.x)[-1]
-        y_label = split_path(self.y)[-1]
 
-        x_data = [data_from_path(case, self.x, cases) for case in cases]
-        y_data = [data_from_path(case, self.y, cases) for case in cases]
+        if not isinstance(self.y, list):
+            y_label = split_path(self.y)[-1]
+            ys = [self.y.copy()]
+            vars = 1
+        else:
+            y_label = "value"
+            vars = len(self.y)
+            ys = self.y.copy()
+
+        x_data = []
+        y_data = []
+
+        for case in cases:
+            for idx in range(vars):
+                x_data.append(data_from_path(case, self.x, cases))
+                y_data.append(data_from_path(case, ys[idx], cases))
 
         x_data, y_data, x_label, y_label = self._handle_data_with_units(
             x_data, y_data, x_label, y_label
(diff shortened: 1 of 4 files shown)


flow360/plugins/report/report_items.py
```diff
@@ -1034,11 +1058,11 @@ def _load_data(self, cases):
                 data.filter(include=self.include, exclude=self.exclude)
                 x_data[i] = data.values[component]
 
-        component = y_label
+        components = [split_path(y)[-1] for y in ys]
         for i, data in enumerate(y_data):
             if isinstance(data, case_results.PerEntityResultCSVModel):
                 data.filter(include=self.include, exclude=self.exclude)
-                y_data[i] = data.values[component]
+                y_data[i] = data.values[components[i % vars_quant]]
```
maciej-flexcompute, line 1065: zip x_data and x_component
zip y_data and y_component

flow360/plugins/report/report_items.py
```diff
@@ -1200,6 +1224,16 @@ def get_data(self, cases: List[Case], context: ReportContext) -> PlotModel:
...
+        elif (len(self.y) > 1) and isinstance(self.y, list):
+            legend = []
+            for case in cases:
+                for y in self.y:
+                    if len(cases) > 1:
+                        legend.append(f"{case.name} - {split_path(y)[-1]}")
+                    else:
+                        legend.append(f"{split_path(y)[-1]}")
+
+            style = "-"
```
maciej-flexcompute, line 1236: move to separate legend handling function. Use `component` list instead of `split_path`

flow360/plugins/report/report_items.py
```diff
@@ -1256,8 +1291,10 @@ def get_background_chart3d(self, cases) -> Tuple[Chart3D, Case]:
         """
         Returns Chart3D for background.
         """
-        x_data, _, _, _ = self._load_data(cases)
-        reference_case = cases[0]
+        # pylint: disable=unsubscriptable-object
+        reference_case_idx = self.select_indices[0] if self.select_indices is not None else 0
+        x_data, _, _, _ = self._load_data([cases[reference_case_idx]])
+        reference_case = cases[reference_case_idx]
```
maciej-flexcompute, line 1297:
```suggestion
        reference_case = cases[reference_case_idx]
        x_data, _, _, _ = self._load_data([reference_case])
```

flow360/plugins/report/report_items.py
piotrkluba: Defined a path_variable_name function in utils and used it instead

flow360/plugins/report/report_items.py
```diff
@@ -812,6 +818,32 @@ class ManualLimit(Flow360BaseModel):
...
+    the range of first values to plot.
+
+    Parameters
+    first : float
+        How much first values are supposed to be plotted.
+    """
+
+    first: float
+
+
```
maciej-flexcompute, line 846: It took me a while to understand how this is supposed to work. I think example is needed. Also, it is in a sense similar to "FixedRangeLimit", the "range" here is last or first. The difference is that our FixedRangeLimit is aligning to "center", we can align to "begin" or "end". @wiktor-flex can you take a look?

flow360/plugins/report/report_items.py
```diff
@@ -1339,6 +1289,229 @@ def get_doc_item(self, context: ReportContext, settings: Settings = None) -> Non
...
+class Chart2D(BaseChart2D):
+    """
+    Represents a 2D chart within a report, plotting x and y data.
+
+    Parameters
+    ----------
+    x : Union[str, Delta]
+        The data source for the x-axis, which can be a string path or a `Delta` object.
+    y : Union[str, Delta, List[str]]
+        The data source for the y-axis, which can be a string path or their list or a `Delta` object.
```
maciej-flexcompute, line 1301: the code also allows for DataItem which is not mentioned in docstring

flow360/plugins/report/report_items.py
```diff
@@ -1339,6 +1289,229 @@ def get_doc_item(self, context: ReportContext, settings: Settings = None) -> Non
...
+    background : Union[Literal["geometry"], None], optional
+        Background type for the chart; set to "geometry" or None.
+    _requirements : List[str]
+        Internal list of requirements associated with the chart.
```
maciej-flexcompute, line 1305: remove

flow360/plugins/report/report_items.py
```diff
@@ -1339,6 +1289,229 @@ def get_doc_item(self, context: ReportContext, settings: Settings = None) -> Non
...
+
+
+class NonlinearResiduals(BaseChart2D):
+    """
+    Residuals is an object for showing the solution history of nonlinear residuals.
+
+    """
+
+    section_title: Literal["Nonlinear residuals"] = Field("Nonlinear residuals", frozen=True)
+    fig_name: Literal["fig-residuals"] = Field("fig-residuals", frozen=True)
```
maciej-flexcompute, line 1451: I think fig_name should not be literal. If you want to create separate charts with nonlinear residual, the PNG files will overwrite themselves. I think it will break validation if you use two `NonlinearResiduals()` instances in one ReportTemplate: https://github.com/flexcompute/Flow360/blob/7ec8f3886aebba6b34bf90cd4d64f103a9d8d6ea/flow360/plugins/report/report.py#L158

I think you can just remove this line:
```suggestion
```

flow360/plugins/report/report_items.py
```diff
@@ -1339,6 +1289,229 @@ def get_doc_item(self, context: ReportContext, settings: Settings = None) -> Non
...
+    caption: Literal[None] = None
```
maciej-flexcompute, line 1452: why we don't allow for caption?

flow360/plugins/report/report_items.py
```diff
@@ -1339,6 +1289,229 @@ def get_doc_item(self, context: ReportContext, settings: Settings = None) -> Non
...
+    x: Literal["nonlinear_residuals/pseudo_step"] = Field(
+        "nonlinear_residuals/pseudo_step", frozen=True
+    )
+    y_log: Literal[True] = Field(True, frozen=True)
+    type_name: Literal["NonlinearResiduals"] = Field("NonlinearResiduals", frozen=True)
```
maciej-flexcompute, line 1457: put type_name as last item. In general Let's put thing user can set as first options and things user cannot touch as last. This is how the will be displayed in IDE hints

flow360/plugins/report/report_items.py
```diff
@@ -1339,6 +1289,229 @@ def get_doc_item(self, context: ReportContext, settings: Settings = None) -> Non
...
+    _requirements: List[str] = [_requirements_mapping["nonlinear_residuals"]]
+    show_grid: Optional[bool] = True
+    separate_plots: Optional[bool] = True
+    xlim: Optional[Union[ManualLimit, Tuple[float, float], LastLimit, FirstLimit]] = LastLimit(
+        last=4000
+    )
```
maciej-flexcompute, line 1463: I think we rather need to strip off first data point. Some simple cases can converge in 500 steps, this will look weird.

flow360/plugins/report/report_items.py
piotrkluba: It is not mentioned as it does not work yet, I can delete it from the definition and add it when it will be fixed

flow360/plugins/report/report_items.py
piotrkluba: The first point is stripped anyway, this is set so when transient cases are supported it will show the last 4000 pseudo steps by default

flow360/plugins/report/report_items.py
piotrkluba: I thing it could be done with FixedRangeLimit, but i would have to be rewritten a bit to become more generic to work for both x axis and y axis. For example the last percent argument does not make sense for x axis, the center_strategy would have to be renamed to alignment_strategy etc..

wiktor-flex: Isn't that much more like setting a ManualLimit or just a tuple for the xlim, except the fact that only one start/end of the limit can be specified? As in I can say that I want my X axis to start at 100 without having to know what is the upper limit I want to aim for.

wiktor-flex: I think that if you want to implement this, then we should remove one of the other types, because it seems like there would be quite a bit of overlap in the functionality. While it is convenient to have many ways to do the same thing, each tailored more to a particular application, I feel like we will start to overwhelm the users with the amount of choices for this. My suggestion would be to take a look at how generic plotting libraries handle the limits and decide on which approach we want to keep. [Matplotlib](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.xlim.html), for example, just has a tuple as both left and right limits and that seems to be enough for them.

piotrkluba: > Isn't that much more like setting a ManualLimit or just a tuple for the xlim, except the fact that only one start/end of the limit can be specified? As in I can say that I want my X axis to end at 100 without having to know what is the exact value of the upper limit I want to aim for.

It is done so that when there are plots with different values on the x axis (for example cases with different amount of pseudo steps), the user can choose to show only the last n pseudo steps within the whole object. With manuallimit user would have to create an object for each case and set the limits manually there

flow360/plugins/report/report_items.py
piotrkluba: we do now

maciej-flexcompute (approved):

PR #904, opened 2025-04-15
[FL-500] Disable DDES when steady

benflexcompute:
Note: No need for hotfix since `test_hybrid_model_for_unsteady_validator` is already in 25.2

### DIFF
--- a/flow360/component/simulation/simulation_params.py
+++ b/flow360/component/simulation/simulation_params.py
@@ -349,7 +349,7 @@ class SimulationParams(_ParamModelBase):
 
     @pd.model_validator(mode="after")
     def check_unsteadiness_to_use_hybrid_model(self):
-        """Only allow hybrid RANS-LES output field for unsteady simulations"""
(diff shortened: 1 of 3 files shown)


flow360/component/simulation/simulation_params.py
```diff
@@ -346,6 +347,11 @@ def check_consistency_wall_function_and_surface_output(self):
         """Only allow wallFunctionMetric output field when there is a Wall model with a wall function enabled"""
         return _check_consistency_wall_function_and_surface_output(self)
 
+    @pd.model_validator(mode="after")
+    def check_unsteadiness_to_use_hybrid_model(self):
+        """Only allow hybrid RANS-LES output field for unsteady simulations"""
```
angranl-flex, line 352: Maybe "Only allow hybrid model for unsteady simulations"? The output fields are not explicitly checked in the validation function.

angranl-flex (approved):

PR #910, opened 2025-04-16
Expression unit test fixes

andrzej-krupka:
@maciej-flexcompute this fixes most of the unit tests (apart from one where I don't want to fix it as the failing unit test is related to error message filtering that we want to implement)

Not sure whether this should be merged - I suppose that depends on whether the `ReferenceGeometry.area` field will get migrated to use expressions in the initial version (which was my understanding - i.e. all scalar/vector fields will be moved to the value/expression union type) or are we only using this for user defined output for now. If it's the second case then I think this PR can be ignored as after frontend finishes integration the `ReferenceGeometry.area` will move back to the value type again...

### DIFF
--- a/flow360/component/simulation/translator/solver_translator.py
+++ b/flow360/component/simulation/translator/solver_translator.py
@@ -975,7 +975,12 @@ def get_solver_json(
         geometry = remove_units_in_dict(dump_dict(input_params.reference_geometry))
         translated["geometry"] = {}
         if input_params.reference_geometry.area is not None:
-            translated["geometry"]["refArea"] = geometry["area"]
+            if geometry["area"]["typeName"] == "number":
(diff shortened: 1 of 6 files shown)


flow360/component/simulation/translator/solver_translator.py
```diff
@@ -975,7 +975,12 @@ def get_solver_json(
         geometry = remove_units_in_dict(dump_dict(input_params.reference_geometry))
         translated["geometry"] = {}
         if input_params.reference_geometry.area is not None:
-            translated["geometry"]["refArea"] = geometry["area"]
+            if geometry["area"]["typeName"] == "number":
```
andrzej-krupka, line 978: This is the only important part of the PR - if refArea will keep on being a value type then we can ignore this.

PR #919, opened 2025-04-18
Add option for inner layer wall model

yifan-flex:


flow360/component/simulation/models/surface_models.py
```diff
@@ -348,6 +348,7 @@ class Wall(BoundaryBase):
         description="Specify if use wall functions to estimate the velocity field "
         + "close to the solid boundaries.",
     )
+    wall_model_type: Literal["BoundaryLayer", "InnerLayer"] = pd.Field("BoundaryLayer")
```
benflexcompute, line 351: Please add description as argument.

benflexcompute: Also do you expect more tunning parameter associated with each? If so we can turn "BoundaryLayer", "InnerLayer" into class/models instead of string literals.

benflexcompute: What happens if use_wall_function is False? Do we need validation? If so it seems out defaults are conflicting with themselves.

yifan-flex: This is likely a temporary parameter. We are planning to introduce blending between the two and remove the need for this selection to be done by user.
If use_wall_function is false, this parameter won't do anything.

benflexcompute (approved):

PR #930, opened 2025-04-22
Fixed the generation of multi-variable Chart2D in cloud

piotrkluba:
Corrected the get_requirements function to account for list in self.y


wiktor-flex (approved):
Good, but fix tests

PR #938, opened 2025-04-23
changed auto meshing example to use user defined farfield

wiktor-flex:


angranl-flex (approved):
Please remember to hotfix 25.2 and 25.5.

PR #941, opened 2025-04-23
Support for transient cases in Chart2D in report plugit

piotrkluba:
- Implemented plotting the history of variables in transient meshes (time/physical_step/cumulative pseudo_step on x axis).
- Implemented the transient nonlinear residuals. 
- When using pseudo_step, the physical_step gets added on secondary axis for specific xlimits.
- Added three new test cases for unit testing (two transient (SA, SST), one steady SST).
Zrzut ekranu 2025-04-23 153948
Zrzut ekranu 2025-04-23 153927


flow360/plugins/report/report_items.py
```diff
@@ -1113,6 +1166,36 @@ def _handle_plot_style(self, x_data, y_data):
...
+                x_min = min(x_data[0])
+                x_max = max(x_data[0])
+                lower_idx = x_data[0].index(max(x_min, x_lim[0] if x_lim is not None else -np.inf))
+                upper_idx = x_data[0].index(min(x_max, x_lim[1] if x_lim is not None else np.inf))
+                physical_steps_to_show = sec_x_data[upper_idx] - sec_x_data[lower_idx]
+                if physical_steps_to_show <= 5:
+                    return [sec_x_data] * len(x_data), "physical_step"
+            else:
+                log.warning("Does not show physical step with multiple cases plotted.")
+        return None, None
```
wiktor-flex, line 1197: Why did we decide to have an opposite implementation from the webUI (pseudo primary, physical secondary)? Another thing is why not check the length of cases earlier here?

wiktor-flex (commented):
Looks really good in general, can we add some docstrings for example, that would clearly explain to the user when the secondary axis will appear and how?

flow360/component/results/base_results.py
```diff
@@ -445,7 +445,7 @@ def include_time(self):
                     "Cannot find time step size for this simulation. Check flow360.json file."
                 )
 
-        elif isinstance(params, SimulationParams):
+        if isinstance(params, SimulationParams):
```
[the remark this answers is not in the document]
piotrkluba: It is a relict of some testing, repaired

flow360/plugins/report/report_items.py
piotrkluba: Q1: Becuse it is complicated to implement, and it is not worth it
Q2: I divided the function into two to split their responsibility

wiktor-flex: Regarding Q2: Isn't the current implementation making the for loop run every time, where it could very well run only when len(cases) == 1?

flow360/plugins/report/report_items.py
```diff
@@ -1176,12 +1176,25 @@ def _cumulate_pseudo_step(self, pseudo_steps):
...
+
+    def _handle_secondary_x_axis(self, cases, x_data, x_lim, x_label):
+        """
+        Creates physical_step array to use on the secondary axis
+        when the primary axis is pseudo_step, and the number of physical_steps visible
+        within x axis limits is less or equal than 5.
+        """
+        if x_label == "pseudo_step" and any(
+            isinstance(case.params.time_stepping, Unsteady) for case in cases
+        ):
```
wiktor-flex, line 1197: Isn't having just one function call with less looping, better?
```suggestion
    def _handle_secondary_x_axis(self, cases, x_data, x_lim, x_label):
        if len(cases) == 1:
            if x_label == "pseudo_step" and any(
                isinstance(case.params.time_stepping, Unsteady) for case in cases
            ):
                for idx, x_series in enumerate(x_data):
                    x_data[idx] = self._cumulate_pseudo_step(x_series)
                path_to_physical_step = self.x.rstrip("pseudo_step") + "physical_step"
                sec_x_data = data_from_path(cases[0], path_to_physical_step, [])
                x_min = min(x_data[0])
                x_max = max(x_data[0])
                lower_idx = x_data[0].index(max(x_min, x_lim[0] if x_lim is not None else -np.inf))
                upper_idx = x_data[0].index(min(x_max, x_lim[1] if x_lim is not None else np.inf))
                physical_steps_to_show = sec_x_data[upper_idx] - sec_x_data[lower_idx]
                if physical_steps_to_show <= 5:
                    return [sec_x_data] * len(x_data), "physical_step"
        else:
            log.warning("Does not show physical step with multiple cases plotted.")

        return None, None
```

flow360/plugins/report/report_items.py
```diff
@@ -1245,6 +1258,8 @@ def get_data(self, cases: List[Case], context: ReportContext) -> PlotModel:
         xlim = self._handle_xlimits()
         ylim = self._calculate_ylimits(x_data, y_data)
 
+        self._handle_transient_pseudo_step(x_data=x_data, x_label=x_label, cases=cases)
+
```
wiktor-flex, line 1262: And then this would be unnecessary

flow360/plugins/report/report_items.py
wiktor-flex: For clarity, we will keep 2 separate functions.

wiktor-flex (approved):

PR #945, opened 2025-04-23
Improved error messages in expressions

maciej-flexcompute:


andrzej-krupka (approved):
LGTM, tested it with the validation service (to see if the wrap validator issue didn't appear again but it seems like it only happened when nesting them)

PR #946, opened 2025-04-23
Nested expression support + expression validation endpoints

andrzej-krupka:


flow360/component/simulation/blueprint/codegen/parser.py
```diff
@@ -73,6 +74,13 @@ def parse_expr(node: ast.AST, ctx: EvaluationContext) -> Any:
             right=parse_expr(node.comparators[0], ctx),
         )
 
+    elif isinstance(node, ast.Subscript):
```
andrzej-krupka, line 77: This adds support for subscripting in expressions, i.e. the user can now write

```
expr = x[0] + y[1]
```

The question remains on how to best handle bounds checking in terms of the error messages

flow360/component/simulation/blueprint/core/expressions.py
```diff
@@ -31,14 +32,20 @@ def used_names(self) -> set[str]:
         raise NotImplementedError
 
 
-class Name(Expression):
+class Name(Expression, Evaluable):
```
andrzej-krupka, line 35: A Name is evaluable, because a name can refer to another name object by reference.

flow360/component/simulation/blueprint/core/expressions.py
```diff
@@ -102,6 +109,28 @@ def used_names(self) -> set[str]:
         return left.union(right)
 
 
+class Subscript(Expression):
+
+    type: Literal["Subscript"] = "Subscript"
+    value: "ExpressionType"
+    slice: "ExpressionType" # No proper slicing for now
+    ctx: Literal["Load", "Store"] # Only load context
```
andrzej-krupka, line 117: For now we don't implement the `Store` context. This is not necessary for rvalue expressions but when we move to full function support we might want those.

flow360/component/simulation/blueprint/flow360/symbols.py
```diff
@@ -84,14 +90,29 @@ def _import_flow360(name: str) -> Any:
...
+    },
 }
 
 # Define allowed modules
-ALLOWED_MODULES = {"flow360", "fl"}
+ALLOWED_MODULES = {"flow360", "fl", "np"}
 
 ALLOWED_CALLABLES = {
     "fl": None,
+    "np": {
```
andrzej-krupka, line 101: This is a very basic set of numpy functions that I have decided to expose. Do we need more? Also, do we want different aliasing perhaps? @maciej-flexcompute

flow360/component/simulation/services.py
```diff
@@ -767,3 +772,54 @@ def update_simulation_json(*, params_as_dict: dict, target_python_api_version: s
...
+    Validate an expression using the specified variable space
+    """
+    errors = []
+    values = []
+    units = []
+
+    loc = ""
+
+    # Populate variable scope
+    for i in range(len(variables)):
```
andrzej-krupka, line 788: Variables are essentially named expressions that don't do internal type validation (their internal type is Any) and can be referred to from the global scope once defined, so we evaluate them all the same as expressions.

flow360/component/simulation/services.py
```diff
@@ -767,3 +772,54 @@ def update_simulation_json(*, params_as_dict: dict, target_python_api_version: s
...
+            variable = UserVariable(name=variable["name"], value=variable["value"])
+            if variable and isinstance(variable.value, Expression):
+                _ = variable.value.evaluate()
+        except (ValueError, KeyError, NameError, UnitParseError) as e:
+            errors.append({
+                "loc": loc,
+                "msg": str(e)
+            })
+
+    for i in range(len(expressions)):
```
andrzej-krupka, line 801: The reasoning behind having a separate expressions field is cases where we want to validate a single expression field but don't want to pass and validate the entire simulation.json file.

flow360/component/simulation/user_code.py
```diff
@@ -91,7 +83,7 @@ class SerializedValueOrExpression(Flow360BaseModel):
 
 class Variable(Flow360BaseModel):
     name: str = pd.Field()
-    value: Union[list[float], float, unyt_quantity, unyt_array] = pd.Field()
+    value: ValueOrExpression[Any] = pd.Field()
```
andrzej-krupka, line 86: This allows us to assign arbitrary expressions to the Variable

flow360/component/simulation/user_code.py
```diff
@@ -106,6 +98,9 @@ def __sub__(self, other):
         return Expression(expression=f"{self.name} - {str_arg}")
 
     def __mul__(self, other):
+        if isinstance(other, Number) and other == 0:
```
andrzej-krupka, line 101: Small optimization to avoid generating dumb `0 * (<very-long-expression-here>)` statements

flow360/component/simulation/user_code.py
```diff
@@ -174,26 +172,53 @@ def __rpow__(self, other):
         str_arg = arg if not parenthesize else f"({arg})"
         return Expression(expression=f"{str_arg} ** {self.name}")
 
+    def __getitem__(self, item):
```
andrzej-krupka, line 175: Allows subscripting

flow360/component/simulation/user_code.py
```diff
@@ -174,26 +172,53 @@ def __rpow__(self, other):
...
+        (arg, _) = _convert_argument(item)
+        return Expression(expression=f"{self.name}[{arg}]")
+
     def __str__(self):
         return self.name
 
     def __repr__(self):
         return f"Variable({self.name} = {self.value})"
 
+    def sqrt(self):
```
andrzej-krupka, line 185: Defining those is required for numpy functions to work. Actual calling of numpy function happens within blueprint when we invoke `evaluate()`

flow360/component/simulation/services.py
```diff
@@ -767,3 +772,54 @@ def update_simulation_json(*, params_as_dict: dict, target_python_api_version: s
         # Expected exceptions
         errors.append(str(e))
     return updated_params_as_dict, errors
+
+
+def validate_expression(variables: list[dict], expressions: list[str]):
+    """
+    Validate an expression using the specified variable space
```
benflexcompute, line 779:
```suggestion
    Validate all given expressions using the specified variable space
```
If I understand correctly?

flow360/component/simulation/services.py
```diff
@@ -767,3 +772,54 @@ def update_simulation_json(*, params_as_dict: dict, target_python_api_version: s
...
+            })
+
+    for i in range(len(expressions)):
+        expression = expressions[i]
+        loc = f"expressions/{i}"
+        value = None
+        unit = None
+        try:
+            expression_object = Expression(expression=expression)
+            result = expression_object.evaluate()
```
benflexcompute, line 808: How do we handle value dependency when seeing nested expression? Are we certain that the expressions are listed in proper order? 

a = "x+1"
b = "a-2"


validate_expression (["x"], [b, a])
?

andrzej-krupka: Yes, the order of declaration matters. We could create a dependency graph and resolve it that way, but I assume since the user declares project variables in a list the frontend can pass us the variable list in the order of declaration.

flow360/component/simulation/blueprint/core/expressions.py
```diff
@@ -3,6 +3,7 @@
 import pydantic as pd
 
 from .context import EvaluationContext
+from .types import Evaluable
```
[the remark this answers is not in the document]
andrzej-krupka: Both are correct it seems, but apparently Evaluatable hasn't been used for a while in academic writing: https://english.stackexchange.com/questions/304948/evaluable-vs-evaluatable

flow360/component/simulation/services.py
andrzej-krupka: Fixed in [6db9e09](https://github.com/flexcompute/Flow360/pull/946/commits/6db9e0930c2bd723625a8df6e07b602aa577a12a)

benflexcompute:
To trigger the CI

benflexcompute (approved):

PR #958, opened 2025-04-25
John/add low dissipation control factors

johnmoore4:


flow360/component/simulation/models/solver_numerics.py
```diff
@@ -119,6 +119,8 @@ class NavierStokesSolver(GenericSolverSettings):
         + "dissipation of the numerical flux. The recommended starting value for most "
         + "low-dissipation runs is 0.2.",
     )
+    low_dissipation_control_factors: Optional[List[float]] = pd.Field(None)
```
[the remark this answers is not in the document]
johnmoore4: @NasserFlexCompute it is a list, and could potentially support more parameters in the future. In the solver it is also treated as a list capable of expanding in the future if needed.

NasserFlexCompute (approved):

flow360/component/simulation/models/solver_numerics.py
benflexcompute, line 122: Why this is not moved into a dict as you originally planned?

tests/simulation/converter/test_monitor_flow360_converter.py
```diff
@@ -44,7 +44,11 @@ def test_flow360_monitor_convert():
     )
     with open("./ref/ref_monitor.json", mode="r") as fp:
         ref_dict = json.load(fp=fp)
-    assert compare_values(params_dict, ref_dict, ignore_keys=["private_attribute_id"])
+    assert compare_values(
+        params_dict,
+        ref_dict,
+        ignore_keys=["private_attribute_id", "low_dissipation_control_factors"],
```
benflexcompute, line 50: Instead of just skipping keys we should update the reference. Adding "low_dissipation_control_factors" here is not a test_monitor_flow360_converter.py logic.

PR #960, opened 2025-04-28
[SCFD-5264] Fix duplication check in entity registry

angranl-flex:


tests/simulation/framework/test_entities_v2.py
```diff
@@ -824,6 +835,33 @@ def test_entity_registry_find_by_id():
         )
 
 
+def test_used_entity_registry_in_simulation_param():
```
benflexcompute, line 838:
```suggestion
def test_same_name_and_type_entities_in_entity_registry():
```

benflexcompute (approved):

PR #961, opened 2025-04-28
[SCFD-3899] Added support for mach reynolds input in Aerospace condition

wiktor-flex:
Changed operating_condition_from_mach_reynolds function from a standalone one to a classmethod of AerospaceCondition. so that it utilizes MultiConstructorBaseModel.


tests/simulation/translator/utils/symmetryBC_param_generator.py
```diff
@@ -66,7 +66,10 @@ def create_symmetryBC_param():
                 SymmetryPlane(entities=[my_symmetry_plane]),
                 Freestream(entities=[my_freestream]),
             ],
-            time_stepping=Steady(CFL=AdaptiveCFL(), max_steps=5000),
+            time_stepping=Steady(
+                CFL=AdaptiveCFL(max=10000, max_relative_change=1, convergence_limiting_factor=0.25),
+                max_steps=5000,
+            ),
```
maciej-flexcompute, line 72: why do we provide default values here? Does `exclude_none` have some undesired behaviour?

flow360/component/simulation/framework/base_model.py
```diff
@@ -483,6 +483,20 @@ def _to_json(self, filename: str, **kwargs) -> None:
         with open(filename, "w+", encoding="utf-8") as file_handle:
             json.dump(model_dict, file_handle, indent=4, sort_keys=True)
 
+    def model_dump(self, **kwargs) -> dict:
```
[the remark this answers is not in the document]
benflexcompute: I would not merge this later as a hotfix to 25.2. This is actually a big change and we have not tested end to end yet. Things may break (key missing error etc) now that the we have no None in the json.

tests/simulation/framework/test_multi_constructor_model.py
```diff
@@ -42,27 +42,47 @@ def get_aerospace_condition_default_and_thermal_state_using_from():
...
+    get_aerospace_condition_default,
+    get_aerospace_condition_using_from_mach,
+    get_aerospace_condition_using_from_mach_reynolds,
+):
     full_data = get_aerospace_condition_default.model_dump(exclude_none=True)
     data_parsed = parse_model_dict(full_data, globals())
     assert sorted(data_parsed.items()) == sorted(full_data.items())
 
-    full_data = get_aerospace_condition_using_from.model_dump(exclude_none=True)
+    full_data = get_aerospace_condition_using_from_mach.model_dump(exclude_none=True)
```
benflexcompute, line 73: can we now remove all the `exlude_none =True` in this test? This actually hide the problem from being tested in the first place.

benflexcompute:
This is not a complete fix since for existing simulation.json on S3 cloud, we still store the key-None pairs in `private_attribute_input_cache` and this PR does not fix that. We need to add None popper in the `parse_model_dict()` too.

Also like I mentioned above. Let's not merge this to 25.2 for now since it is a public version. model_dump fix may introduce bugs.

I asked @angranl-flex to add the None-popper in an dedicated PR. We can merge this into develop + 25.5 branch

tests/simulation/framework/test_multi_constructor_model.py
benflexcompute: Nevermind. Above comment can be addressed in Angran's PR. You do not need to change this.

tests/simulation/translator/utils/symmetryBC_param_generator.py
wiktor-flex: The default values for those field in RampCFL and AdaptiveCFL that were added, are currently set to None as default
https://github.com/flexcompute/Flow360/blob/9d787093a04accd91abe563a8ab7301885c34868/flow360/component/simulation/time_stepping/time_stepping.py#L18
https://github.com/flexcompute/Flow360/blob/9d787093a04accd91abe563a8ab7301885c34868/flow360/component/simulation/time_stepping/time_stepping.py#L57
That is why I provided these values explicitly.

flow360/component/simulation/operating_condition/operating_condition.py
```diff
@@ -370,6 +373,102 @@ def from_mach(
             reference_velocity_magnitude=reference_velocity_magnitude,
         )
 
+    # pylint: disable=too-many-arguments
+    @MultiConstructorBaseModel.model_constructor
+    @pd.validate_call
+    def from_mach_reynolds(
```
benflexcompute, line 379: Do we have a unit test to ensure this is what we expected? Like hand-calculated version vs function results?

wiktor-flex: We have unittests that compare the function output against expected values but I don't think we have reference hand calculations saved anywhere for those. Do you want me to prepare some?

flow360/component/simulation/operating_condition/operating_condition.py
```diff
@@ -370,6 +373,102 @@ def from_mach(
...
+
+        """
+
+        if temperature.units is u.K and temperature.value == 288.15:
+            log.info("Default value of 288.15 K will be used as temperature.")
+
+        material = Air()
+
+        velocity = (
+            mach if reference_mach is None else reference_mach
```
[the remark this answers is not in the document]
wiktor-flex: But this is only velocity used to calculate density which should be correct in this case. Below there is velocity_magnitude which is actually passed as velocity_magnitude to the AerospaceCondition class. I can change the name of this variable if it is confusing.

flow360/component/simulation/operating_condition/operating_condition.py
```diff
@@ -244,8 +244,11 @@ class AerospaceConditionCache(Flow360BaseModel):
     """[INTERNAL] Cache for AerospaceCondition inputs"""
 
     mach: Optional[pd.NonNegativeFloat] = None
+    reynolds: Optional[pd.PositiveFloat] = None
```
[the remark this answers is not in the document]
wiktor-flex: Doesn't PositiveFloat allow for infinity? I just tried using math.inf and it works.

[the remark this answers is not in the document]
maciej-flexcompute: can we check if is will serialise correctly? I remember JSON had some problems with that so in v1 (Flow360Params) we used Literal["inf"] as string.

flow360/component/simulation/operating_condition/operating_condition.py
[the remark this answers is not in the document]
maciej-flexcompute: this is incorrect. Reference values does not affect run but post-processing. Thus should not be used to specify any initial or boundary conditions.

flow360/component/simulation/operating_condition/operating_condition.py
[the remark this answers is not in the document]
wiktor-flex: It serialises to null but if that value is later extracted, it behaves properly when checking for example if it is bigger than some number.

flow360/component/simulation/operating_condition/operating_condition.py
```diff
@@ -370,6 +373,100 @@ def from_mach(
             reference_velocity_magnitude=reference_velocity_magnitude,
         )
 
+    # pylint: disable=too-many-arguments
+    @MultiConstructorBaseModel.model_constructor
+    @pd.validate_call
+    def from_mach_reynolds(
+        cls,
+        mach: pd.NonNegativeFloat,
```
benflexcompute, line 381: mach == 0 will result in div by zero exception.

benflexcompute:
Let's ensure the Reynolds is consistent with Solver definition. @wiktor-flex Could you also update the definition here?
https://github.com/flexcompute/Flow360/blob/fc764bbd5b631612d82d1c2585d0deb4aeb31ba3/flow360/component/simulation/operating_condition/operating_condition.py#L399

flow360/component/simulation/operating_condition/operating_condition.py
wiktor-flex: Right, thanks!

flow360/component/simulation/operating_condition/operating_condition.py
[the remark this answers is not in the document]
wiktor-flex: Now it uses mach only and not reference mach.

flow360/component/simulation/operating_condition/operating_condition.py
benflexcompute: Yes, having some hand calculated values are safer. You can hand calculated against these "expected" values and see if they work. If so you can comment your hand calculated expression along with it.

wiktor-flex:
I did hand calculation for 
`AerospaceCondition.from_mach_reynolds(
        mach=0.2,
        reynolds=5e6,
        temperature=288.15 * u.K,
        alpha=2.0 * u.deg,
        beta=0.0 * u.deg,
        project_length_unit=u.m
)`
and got matching results. I followed the equations piece by piece, starting from calculating dynamic viscosity from Sutherland, then freestream speed of sound, freestream velocity and finally density.
My results:
dynamic viscosity = 1.789297e-5 Pa*s
speed of sound = 340.294 m/s
velocity = 68.0588 m/s
density = 1.314523 kg/m^3

PR #964, opened 2025-04-28
replaced low_dissipation_control_factors with private_attribute_dict

johnmoore4:


flow360/component/simulation/models/solver_numerics.py
```diff
@@ -119,7 +119,7 @@ class NavierStokesSolver(GenericSolverSettings):
         + "dissipation of the numerical flux. The recommended starting value for most "
         + "low-dissipation runs is 0.2.",
     )
-    low_dissipation_control_factors: Optional[List[float]] = pd.Field(None)
+    private_attribute_dict: Optional[Dict] = pd.Field(None)
```
benflexcompute, line 122: There is already one in GenericSolverSettings. No need for re definition.

PR #968, opened 2025-04-29
Pop keys with None value in the private_attribute_input_cache for multi-constructor

angranl-flex:


flow360/component/simulation/framework/multi_constructor_model_base.py
```diff
@@ -182,6 +182,7 @@ def model_custom_constructor_parser(model_as_dict, global_vars):
             id_kwarg["private_attribute_id"] = model_as_dict["private_attribute_id"]
         if constructor_name != "default":
             constructor = get_class_method(model_cls, constructor_name)
+            input_kwargs = {key: val for key, val in input_kwargs.items() if val is not None}
```
benflexcompute, line 185: Can we add a comment here saying why we add this? Also write a note that in the future this will not be required once the model_dump always exclude none.

tests/simulation/framework/test_multi_constructor_model.py
```diff
@@ -50,41 +50,45 @@ def get_aerospace_condition_using_from():
     )
 
 
+def compare_objects_from_dict(dict1: dict, dict2: dict, object_class: Flow360BaseModel):
```
benflexcompute, line 53:
```suggestion
def compare_objects_from_dict(dict1: dict, dict2: dict, object_class: type[Flow360BaseModel]):
```

benflexcompute: It is a class not instance/object

flow360/component/simulation/framework/multi_constructor_model_base.py
angranl-flex: OK, comment addressed

tests/simulation/framework/test_multi_constructor_model.py
angranl-flex: Fixed!

flow360/component/simulation/framework/multi_constructor_model_base.py
```diff
@@ -182,6 +182,12 @@ def model_custom_constructor_parser(model_as_dict, global_vars):
             id_kwarg["private_attribute_id"] = model_as_dict["private_attribute_id"]
         if constructor_name != "default":
             constructor = get_class_method(model_cls, constructor_name)
+            # pylint: disable=fixme
+            # The current simulation.json file from model_dump includes keys with None value. This fix is to
```
benflexcompute, line 186:
```suggestion
            # The current & previous simulation.json file from model_dump() includes keys with None value. This fix is to
```

flow360/component/simulation/framework/multi_constructor_model_base.py
```diff
@@ -182,6 +182,12 @@ def model_custom_constructor_parser(model_as_dict, global_vars):
...
+            # pop such keys in the private_attribute_input_cache to ensure these files are parsed correctly.
```
benflexcompute, line 187:
```suggestion
            # pop such keys in the private_attribute_input_cache to ensure only required arguments are passed to constructor.
```

benflexcompute (approved):

flow360/component/simulation/framework/multi_constructor_model_base.py
angranl-flex: Updated

PR #983, opened 2025-05-02
made dcldastall always positive

jbungenerflex:
HEllo @maciej-flexcompute , after our discussion about the customer Xrotor input file I thought about it some more. 

IF we simply guarantee that dcldastall is positive then we get the expected behavior. 
We don't care about the values coming from Xrotor outside of Clim to Cl max region b/c we replace with the flat plate assumption anyway.
This is the cleanest, I feel.


jbungenerflex:
@maciej-flexcompute , any inputs here?

Thanks

maciej-flexcompute (approved):

PR #986, opened 2025-05-05
Reorganized solver variables into target namespaces

andrzej-krupka:


flow360/__init__.py
```diff
@@ -148,51 +150,8 @@
...
-    momentZ,
-    nodeNormals,
-    theta,
-    omega,
-    omegaDot,
-    wallFunctionMetric,
-    wallShearStress,
-    yPlus,
-)
+from flow360.component.simulation.variables.control_variables import *
```
benflexcompute, line 153: I recommend not import *. This makes the fl namespace very crowded.

flow360/component/simulation/blueprint/flow360/symbols.py
```diff
@@ -29,15 +29,74 @@ def _import_flow360(name: str) -> Any:
...
+            "kOmega",
+            "nuHat",
+            "solutionTransition",
+            "residualTransition",
+            "solutionHeatSolver",
+            "residualHeatSolver",
+            "coordinate",
+            "physicalStep",
+            "pseudoStep",
+            "timeStepSize",
```
benflexcompute, line 67: We had a discussion quiet long time ago on the complaint of output field names. They do not abide the snake case convention we enforce throughout Python client.

I think we should rename all these before release. No need to do it for now but I would add a TODO note here.

See
https://www.notion.so/flexcompute/Improve-naming-for-Flow360-outputs-1a1db96dc14e805d8c0cfd1beb43b88c

benflexcompute (approved):

flow360/__init__.py
andrzej-krupka: Good point. Will fix.

flow360/component/simulation/blueprint/flow360/symbols.py
andrzej-krupka: I added a `TODO`. If we change those we need to remember to update the documentation as well, as I took those names directly from our docs.

andrzej-krupka: Fixed in [1012c72](https://github.com/flexcompute/Flow360/pull/986/commits/1012c7249a5782bc0e9aacad35fd5f1c4edf0a04)

flow360/__init__.py
andrzej-krupka: Fixed in [1012c72](https://github.com/flexcompute/Flow360/pull/986/commits/1012c7249a5782bc0e9aacad35fd5f1c4edf0a04)

PR #992, opened 2025-05-05
[FL-729] [POC] Dimensioned output

benflexcompute:
⚠️ **Main issue worth thinking:**

1. Currently, the output unit selection is done by checking the unit system used in the simulation.json. Is this straightforward to user? It is almost impossible for user to change this without resetting the unit system. Should we store the unit system per output field? 
For e.g.

``` python
class SolverVariable(Variable):
    ...
    unit_system_name: str = pd.Field("SI")
    
    def in_SI(self) -> SolverVariable:
        return SolverVariable(unit_system_name = "SI", **(self.model_dump()))
```
and then allow per-instance overload?

2. There will be overlapping with the base branch. For example the varaible name in the UDF source code. (SolverVaraible -> solver_side_name) I think it is fine we do not use solver_side_name but instead just the user facing name but I was thinking to retire user-facing variable string completly so the solver does not need to know the mapping from "solverNavierStokes" --> "NavierStokes_solver".

👀 Following script demonstrate usage.

```python
import json

import flow360 as fl
from flow360.component.simulation.framework.param_utils import AssetCache
from flow360.component.simulation.solver_builtins import *

# from flow360.component.simulation.primitives import Surface
from flow360.component.simulation.unit_system import (
    MassFluxType,
)


print(">> ", fl.SI_unit_system[MassFluxType.dim_name])
print(">> ", MassFluxType.dim)

with imperial_unit_system:
    params = fl.SimulationParams(
        operating_condition=fl.AerospaceCondition(velocity_magnitude=10),
        outputs=[fl.VolumeOutput(output_fields=[solutionNavierStokes, solutionNavierStokes])],
        private_attribute_asset_cache=AssetCache(project_length_unit="m"),
    )
with open("ttt.json", "w") as f:
    f.write(params.model_dump_json())

with open("ttt.json", "r") as f:
    param_as_dict = json.load(f)

params, _validation_errors, _ = fl.services.validate_model(
    params_as_dict=param_as_dict,
    validated_by=fl.services.ValidationCalledBy.PIPELINE,
    root_item_type="VolumeMesh",
    validation_level="CASE",  # Only need to validate CASE related fields.
)
print(">> ", _validation_errors)

translated, _ = fl.services.simulation_to_case_json(params, "cm")
with open("translated.json", "w") as f:
    json.dump(translated, fp=f, indent=4)
```

```json
"userDefinedFields": [
        {
            "expression": "solutionNavierStokes_SI[0] = NavierStokes_solution[0] / 0.8163265306122448; solutionNavierStokes_SI[1] = NavierStokes_solution[1] / 0.0023988860123275884; solutionNavierStokes_SI[2] = NavierStokes_solution[2] / 0.0023988860123275884; solutionNavierStokes_SI[3] = NavierStokes_solution[3] / 0.0023988860123275884; solutionNavierStokes_SI[4] = NavierStokes_solution[4] / 7.049451272672674e-06;",
            "name": "solutionNavierStokes_SI"
        }
    ],
```

❓ **TODO:**

- [ ] Proper unit test.
- [ ] Figure out a proper way to associate the source entity with the variable.


flow360/component/simulation/translator/solver_translator.py
```diff
@@ -1283,13 +1320,14 @@ def get_solver_json(
     )
 
     ##:: Step 4: Get outputs (has to be run after the boundaries are translated)
-
-    translated = translate_output(input_params, translated)
+    with input_params.unit_system:
```
benflexcompute, line 1323: See the PR discussion first item. We use param unit system to determine the units used.

flow360/component/simulation/translator/utils.py
```diff
@@ -117,9 +117,10 @@ def convert_tuples_to_lists(input_dict):
 def remove_units_in_dict(input_dict):
     """Remove units from a dimensioned value."""
     unit_keys = {"value", "units"}
+    const_value_expression_keys = {"value", "units", "typeName"}
```
benflexcompute, line 120: Is this valid translation for :class:`ValueOrExpression`? I just changed this so om6Wing runs.

flow360/component/simulation/user_code.py
```diff
@@ -189,19 +190,113 @@ def update_context(cls, value):
...
+        return conversion_factors
 
 
-def _handle_syntax_error(se: SyntaxError, source: str):
-    caret = (
-        " " * (se.offset - 1) + "^"
-        if se.text and se.offset
-        else None
+class SolverVariable(Variable):
+    value: Union[list[float], float, unyt_quantity, unyt_array, None] = pd.Field(None)
```
benflexcompute, line 231: Previous float("Nan") cannot be serialized / will be serialized as None

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -218,6 +221,20 @@ def rotation_translator(model: Rotation):
     return volume_zone
 
 
+def _convert_output_field(output_field: Union[SolverVariable, str]):
+    """Convert output field to string"""
+    if isinstance(output_field, str):
+        return output_field
+    elif isinstance(output_field, SolverVariable):
```
andrzej-krupka, line 228: I don't fully understand this. The output field will usually not be a raw SolverVariable, but rather an expression.

If the user writes:

```
fl.timeStepSize * 2
```

the resulting object type is an `Expression`. If he assigns a raw variable to a `ValueOrExpression[T]` field it will also get converted to an expression in the validator.

I understand the `str` output field type is the legacy form of defining the expression?

flow360/component/simulation/user_code.py
```diff
@@ -189,19 +190,113 @@ def update_context(cls, value):
...
+        Get the name of the user defined function for this variable.
+        """
+        cleaned_field_name = self.name.removeprefix("fl.")
+        # TODO: Incorporate the entity info into the name.
+        if self.is_dimensioned:
+            return f"{cleaned_field_name}_{unit_system.name}"
+        else:
+            return cleaned_field_name
+
+    def _get_udf_source_code(self, unit_system: UnitSystem, params) -> str:
```
andrzej-krupka, line 270: I don't like generating source code like this... The main reason for using expressions is that we can serialize arbitrary python code to strings and deserialize strings into source code. This solution assumes that every user defined field can only contain the solver variable, right?  Let's have a call and discuss this today maybe?

flow360/component/simulation/utils.py
```diff
@@ -61,7 +61,7 @@ def is_instance_of_type_in_union(obj, typ) -> bool:
 
 class UnknownFloat(float):
```
andrzej-krupka, line 62: Oh - this can probably be removed. It was being used by solver variables before but has been deprecated since then.

flow360/component/simulation/translator/solver_translator.py
benflexcompute: Yes str is legacy support

flow360/component/simulation/user_code.py
benflexcompute: Sure we can use :class:`expression` to convert expression.

flow360/component/simulation/utils.py
benflexcompute: removed

flow360/component/simulation/translator/solver_translator.py
benflexcompute: Sure we can use expression but how do we tell expression what is the target unit? How about we do:

```Python
class SolverVariable(Variable):
    ...
    unit_system_name: str = pd.Field("SI")
    
    # def in_SI(self) -> SolverVariable:
    def in_SI(self) -> Expression
    #    return SolverVariable(unit_system_name = "SI", **(self.model_dump()))
        retuen Expression(self*conversion_constant)
```

benflexcompute: But the conversion rate depends on the entire SimulationParam.... Maybe we can defer the value substitution of `conversion_constant`?

PR #1009, opened 2025-05-07
Added ability to convert expressions to C++ syntax

andrzej-krupka:
This is a PR that modifies the blueprint code generator module to allow converting python expressions to C++-like syntax. 

We can now call `expression.to_solver_code()` which returns a valid C++ rvalue expression. Some python features like range calls, list comprehensions are not supported for now, but wherever possible operators are being converted, so for example for a python expression initialized as `expr=(x // 3) ** 2` calling `expr.to_solver_code()` will yield `pow(floor(x / 3), 2)`

Solver variables can be remapped to their C++ solver equivalent at the time of declaration like so:

```
mut = SolverVariable(name="solution.mut", solver_name="mut", value=float("NaN"))
```

which will be reflected in the generated C++ code.

Some things to consider which are still not implemented:

1. Unit stripping and unit system conversion - before sending the code to the solver we need to get rid of all units and swap them for conversion factors instead - this can probably be done by splitting the code by whitespaces and operator delimiters to obtain all operands, but I haven't implemented anything yet.
2. Vector operations - how do we handle vector operations? In #946 we introduce numpy interop which works great in python, but if we want to do the same in C++ we should probably define a vector type supporting vector algebra on the solver side, e.g. `np.dot(a, np.array([0, 0, 1]))` could convert to `Vec3::dot(a, Vec3{0, 0, 1})`. I guess we need to ask somebody in the solver team for opinion on this.

### DIFF
--- a/flow360/component/simulation/blueprint/codegen/generator.py
+++ b/flow360/component/simulation/blueprint/codegen/generator.py
@@ -1,3 +1,4 @@
+import functools
 from typing import Any
 
 from ..core.expressions import (
@@ -9,10 +10,12 @@ from ..core.expressions import (
     Name,
     RangeCall,
     Tuple,
+    UnaryOp,
 )
 from ..core.function import Function
 from ..core.statements import Assign, AugAssign, ForLoop, IfElse, Return, TupleUnpack
-from ..utils.operators import BINARY_OPERATORS
+from ..utils.operators import BINARY_OPERATORS, UNARY_OPERATORS
+from ..utils.types import TargetSyntax
 
 
 def _indent(code: str, level: int = 1) -> str:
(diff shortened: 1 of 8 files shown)


flow360/component/simulation/blueprint/codegen/generator.py
```diff
@@ -28,130 +31,246 @@ def _indent(code: str, level: int = 1) -> str:
...
+        if result is None:
+            raise ValueError(f"Unsupported syntax type, available {[syntax.name for syntax in TargetSyntax]}")
+        return result
+    return wrapper
+
+
+@check_syntax_type
+def _empty(syntax):
+    if syntax == TargetSyntax.PYTHON:
         return "None"
```
benflexcompute, line 47:
```suggestion
        return None
```
Seems check_syntax_type expects None instead.

Also why CPP and Python returns differently?

flow360/component/simulation/user_code.py
```diff
@@ -190,10 +193,13 @@ def update_context(cls, value):
 
 
 class SolverVariable(Variable):
+    solver_name: Optional[str] = pd.Field(None, alias="solverName")
```
benflexcompute, line 196: There is no need for alias for serialization purpose. As long as it is inherited from Flow360BaseModel.

https://github.com/flexcompute/Flow360/blob/d2f44b91bc7bbf9586a6be8f6e56f65066f55c74/flow360/component/simulation/framework/base_model.py#L166

What is the reason we add the alias here?

andrzej-krupka: Oh, I didn't realize we have this. I should remove a lot of aliasing in other places then. Thanks for the tip.

flow360/component/simulation/blueprint/codegen/generator.py
andrzej-krupka: Because the string representation of the null object is different. Now that I look I realize I made a mistake. It should either be `NULL` or `nullptr`

andrzej-krupka: `check_syntax_type` is just a decorator that will raise an error if the specified target syntax is invalid (its a bit hacky, as we always expect the functions to return a string representation of the expression so if we return `None` it probably means something has gone wrong.

andrzej-krupka: Changed to "nullptr" for CPP in [b7fbe78](https://github.com/flexcompute/Flow360/pull/1009/commits/b7fbe78d70b4085172cb2bb9cde85adf17cd1b68)

flow360/component/simulation/user_code.py
andrzej-krupka: Removed all aliases in user_code.py in [b7fbe78](https://github.com/flexcompute/Flow360/pull/1009/commits/b7fbe78d70b4085172cb2bb9cde85adf17cd1b68)

PR #1012, opened 2025-05-08
[FL-729] [FLPY-7] Dimensioned Volume Output

benflexcompute:
Previous implementation: https://github.com/flexcompute/Flow360/pull/992

### DIFF
--- a/flow360/component/simulation/outputs/outputs.py
+++ b/flow360/component/simulation/outputs/outputs.py
@@ -261,7 +261,6 @@ class VolumeOutput(_AnimationAndFileFormatSettings):
     """
 
     name: Optional[str] = pd.Field("Volume output", description="Name of the `VolumeOutput`.")
-    # TODO: Deprecate the string inputs and raise deprecation warning
     output_fields: UniqueItemList[Union[VolumeFieldNames, str, UserVariable]] = pd.Field(
         description="List of output variables. Including :ref:`universal output variables<UniversalVariablesV2>`,"
         " :ref:`variables specific to VolumeOutput<VolumeAndSliceSpecificVariablesV2>`"
--- a/flow360/component/simulation/outputs/outputs.py
+++ b/flow360/component/simulation/outputs/outputs.py
@@ -37,6 +37,7 @@ from flow360.component.simulation.primitives import (
     Surface,
 )
 from flow360.component.simulation.unit_system import LengthType
+from flow360.component.simulation.user_code.core.types import UserVariable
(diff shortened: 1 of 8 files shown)


flow360/component/simulation/outputs/outputs.py
```diff
@@ -260,7 +261,9 @@ class VolumeOutput(_AnimationAndFileFormatSettings):
     """
 
     name: Optional[str] = pd.Field("Volume output", description="Name of the `VolumeOutput`.")
-    output_fields: UniqueItemList[Union[VolumeFieldNames, str]] = pd.Field(
+    # TODO: Not all SolverVariables can be used here.
+    # TODO: `Expression` and `SolverVariable` still need business logic for validation (Surface Field/Expression?)
+    output_fields: UniqueItemList[Union[VolumeFieldNames, str, ValueOrExpression]] = pd.Field(
```
andrzej-krupka, line 266: Since ValueOrExpression is meant to be used as a generic type (e.g. `ValueOrExpression[LengthType.Vector]` etc..) it might be better to just use `Expression` here (I think its functionally identical?)... As I understand that pure values here (like `4 * u.m / u.s`) make no sense here, so the field type might be slightly misleading.

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -514,28 +635,39 @@ def process_output_fields_for_udf(input_params: SimulationParams):
...
     generated_udfs = []
-    for field_name in all_field_names:
-        udf_expression = generate_predefined_udf(field_name, input_params)
-        if udf_expression:
-            generated_udfs.append(UserDefinedField(name=field_name, expression=udf_expression))
+    for field in all_fields:
+        if isinstance(field, str):
+            udf_expression = generate_predefined_udf(field, input_params)
+            if udf_expression:
+                generated_udfs.append(UserDefinedField(name=field, expression=udf_expression))
```
andrzej-krupka, line 660: This `UserDefinedField` type looks very similar to what a `UserVariable` is - do you think we could leverage this or is there some extra business logic that we need to take care of? Maybe we could inherit?

flow360/component/simulation/utils.py
```diff
@@ -57,34 +57,3 @@ def is_instance_of_type_in_union(obj, typ) -> bool:
 
     # Otherwise, do a normal isinstance check.
     return isinstance(obj, typ)
-
```
andrzej-krupka, line 60: Thanks for removing that. I forgot about it.

PR #1013, opened 2025-05-08
Add dependency cycle checking and add non-dimensioned array handling to expression validation service

andrzej-krupka:
Additional variable validator and some small improvements to the expression validator service. Details in comments.

### DIFF
--- a/flow360/component/simulation/services.py
+++ b/flow360/component/simulation/services.py
@@ -7,6 +7,7 @@ from enum import Enum
 from numbers import Number
 from typing import Any, Collection, Dict, Literal, Optional, Tuple, Union
 
+import numpy as np
 import pydantic as pd
 from unyt import unyt_array, unyt_quantity
 from unyt.exceptions import UnitParseError
@@ -787,8 +788,6 @@ def validate_expression(variables: list[dict], expressions: list[str]):
     values = []
     units = []
 
-    loc = ""
-
     # Populate variable scope
     for i in range(len(variables)):
         variable = variables[i]
@@ -796,7 +795,7 @@ def validate_expression(variables: list[dict], expressions: list[str]):
         try:
             variable = UserVariable(name=variable["name"], value=variable["value"])
             if variable and isinstance(variable.value, Expression):
-                _ = variable.value.evaluate()
+                _ = variable.value.evaluate(strict=False)
         except (ValueError, KeyError, NameError, UnitParseError) as e:
             errors.append({"loc": loc, "msg": str(e)})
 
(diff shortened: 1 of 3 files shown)


flow360/component/simulation/services.py
```diff
@@ -787,16 +788,14 @@ def validate_expression(variables: list[dict], expressions: list[str]):
...
-
     # Populate variable scope
     for i in range(len(variables)):
         variable = variables[i]
         loc = f"variables/{i}"
         try:
             variable = UserVariable(name=variable["name"], value=variable["value"])
             if variable and isinstance(variable.value, Expression):
-                _ = variable.value.evaluate()
+                _ = variable.value.evaluate(strict=False)
```
andrzej-krupka, line 798: This means that using solver variables in expressions will not throw an error. We are fine with solver variables as long as their value is not being used.

flow360/component/simulation/services.py
```diff
@@ -807,15 +806,22 @@ def validate_expression(variables: list[dict], expressions: list[str]):
         unit = None
         try:
             expression_object = Expression(expression=expression)
-            result = expression_object.evaluate()
-            if isinstance(result, Number):
+            result = expression_object.evaluate(strict=False)
```
andrzej-krupka, line 809: If we pass validation and evaluate to NaN (perhaps because of a solver variable) we should not throw an error, just not write an evaluated value.

flow360/component/simulation/user_code.py
```diff
@@ -214,6 +212,28 @@ def update_context(cls, value):
         _user_variables.add(value.name)
         return value
 
+    @pd.model_validator(mode="after")
```
andrzej-krupka, line 215: This runs a regular DFS tree search but by keeping track of the path we traverse we check for cycles. If a cycle is found we throw an error and print the cycle in a readable format, e.g.

```
x -> y -> z -> x
```

flow360/component/simulation/user_code.py
```diff
@@ -286,7 +306,7 @@ def _validate_expression(cls, value) -> Self:
 
     def evaluate(
         self, context: EvaluationContext = None, strict: bool = True
-    ) -> Union[float, list[float], unyt_array]:
+    ) -> Union[float, np.ndarray, unyt_array]:
```
andrzej-krupka, line 309: After numpy interop using regular list types will not be supported to avoid confusion

flow360/component/simulation/user_code.py
```diff
@@ -214,6 +212,28 @@ def update_context(cls, value):
         _user_variables.add(value.name)
         return value
 
+    @pd.model_validator(mode="after")
+    @classmethod
+    def check_dependencies(cls, value):
```
benflexcompute, line 217:
```suggestion
    def check_cyclic_dependencies(cls, value):
```

benflexcompute (approved):

benflexcompute (approved):

PR #1017, opened 2025-05-09
[FL-874] Add support for adding tags from python API

piotrkluba:


flow360/component/resource_base.py
```diff
@@ -97,6 +97,15 @@ def handle_none_str(cls, value):
...
 
+    @property
+    def category_tag(self):
+        """
+        Returns the first tag.
+        """
+        if self.tags:
+            return self.tags[0]
+        return None
+
```
flexcompute-readthedocs, line 108: I suggest to remove concept of `category_tag` and in report allow user to chose by which tag to group, 1st being the default.

maciej-flexcompute, line 108: I suggest to remove concept of `category_tag` and in report allow user to chose by which tag to group, 1st being the default.

flow360/component/resource_base.py
```diff
@@ -301,6 +301,21 @@ def get_info(self, force=False) -> AssetMetaBaseModel:
...
+        ----------
+        tag: str
+            Tag to add, if tag is already present it will be moved to the end of the tags list
+        category_tag: Optional[bool]
+            Whether to set the tag as the category tag (insert as the first tag)
+        """
+        if tag in self.get_info().tags:
+            self.get_info().tags.remove(tag)
+        self.get_info().tags.insert(0, tag)
+
```
maciej-flexcompute, line 318: this will have no effect on cloud resource, we either need to remove this, all call `patch` webAPI endpoint to update in cloud

piotrkluba: removed for now

maciej-flexcompute (approved):

PR #1030, opened 2025-05-12
Validation service fixes, better error messages

andrzej-krupka:


flow360/component/simulation/framework/base_model.py
```diff
@@ -169,6 +196,7 @@ def __pydantic_init_subclass__(cls, **kwargs) -> None:
         # TODO: Remove alias_generator since it is only for translator
         alias_generator=pd.AliasGenerator(
             serialization_alias=snake_to_camel,
+            validation_alias=camel_to_snake,
```
andrzej-krupka, line 199: I do not understand why this doesn't seem to be getting applied when I try to load a JSON dict that is using a camelCase alias... Only explicitly specifying the alias works for me.

flow360/component/simulation/services.py
```diff
@@ -346,6 +345,9 @@ def validate_model(
     validation_warnings = None
     validated_param = None
 
+    if validated_by in [ValidationCalledBy.SERVICE, ValidationCalledBy.PIPELINE]:
+        Expression.force_serialized_format(True)
```
andrzej-krupka, line 349: Since the frontend is always using the serialized format we do not need the advanced type inference that the python client implementation of the expressions interface offers (In particular without this the internal validator of the variable type - Any - was causing issues when invalid data was passed from the frontend to the service)

flow360/component/simulation/services.py
```diff
@@ -409,7 +414,9 @@ def clean_unrelated_setting_from_params_dict(params: dict, root_item_type: str)
     return params
 
 
-def handle_generic_exception(err: Exception, validation_errors: Optional[list]) -> list:
+def handle_generic_exception(
+    err: Exception, validation_errors: Optional[list], loc_prefix: Optional[list[str]] = None
```
andrzej-krupka, line 418: Generic exceptions can have some meaningful loc too, so we allow passing it as an optional parameter (used in expression endpoint)

flow360/component/simulation/user_code.py
```diff
@@ -444,11 +461,15 @@ def _internal_validator(value: Expression):
...
         def _deserialize(value) -> Self:
             is_serialized = False
-            try:
+            if Expression._serialized_format_only:
                 value = SerializedValueOrExpression.model_validate(value)
                 is_serialized = True
-            except Exception as err:
-                pass
+            else:
+                try:
```
benflexcompute, line 468: What is the difference setting Expression._serialized_format_only True V.S. False?

andrzej-krupka: It's subtle. When we are using `Expression._serialized_format_only = False` the validator will fail if it fails to parse the simulation.json contents as a `SerializedValueOrExpression` object. With `Expression._serialized_format_only = True` it will try to fall back to other validators in the python client. 

The issue with that shows up when we declare variables. Since the variable value is stored as `ValueOrExpression[Any]` the internal type validator will accept anything. This is OK on the python client side (as the user can assign just about any value he wants to the variable he declares) but when validating data from the frontend (which is agreed to always be stored as `SerializedValueOrExpression`) it might obscure validation errors that we should return.

flow360/component/simulation/user_code.py
```diff
@@ -256,6 +262,12 @@ class Expression(Flow360BaseModel, Evaluable):
 
     model_config = pd.ConfigDict(validate_assignment=True)
 
+    _serialized_format_only: ClassVar[bool] = False
+
+    @classmethod
+    def force_serialized_format(cls, value: bool):
```
benflexcompute, line 268:
```suggestion
    def expects_serialized_format(cls, value: bool):
```

flow360/component/simulation/user_code.py
```diff
@@ -444,11 +461,15 @@ def _internal_validator(value: Expression):
...
                 value = SerializedValueOrExpression.model_validate(value)
                 is_serialized = True
-            except Exception as err:
-                pass
+            else:
+                try:
+                    value = SerializedValueOrExpression.model_validate(value)
+                    is_serialized = True
+                except Exception as err:
+                    pass
```
benflexcompute, line 472:
```suggestion
            try:
                value = SerializedValueOrExpression.model_validate(value)
                is_serialized = True
            except Exception as err:
                if Expression._serialized_format_only:
                    raise err
                pass
```

flow360/component/simulation/framework/base_model.py
```diff
@@ -168,7 +168,7 @@ def __pydantic_init_subclass__(cls, **kwargs) -> None:
         # pylint: disable=fixme
         # TODO: Remove alias_generator since it is only for translator
         alias_generator=pd.AliasGenerator(
-            serialization_alias=snake_to_camel,
+            alias=snake_to_camel,
```
benflexcompute, line 171: What is this change for?

flow360/component/simulation/user_code.py
benflexcompute: Ok. I still do not quiet understand the scenario where this is useful. Could you provide an example to demonstrate why this is needed?

andrzej-krupka: Sure. It's easiest to see the issues when we compare how the user initializes variables in Python and frontend:

In python its usually something like:

```
x = UserVariable(name="x", value=1)
y = UserVariable(name="y", value=x ** 2 + x + 1)
```

Since the type of the variable value is `ValueOrExpression[Any]` (so practically `Union[Expression, Any]`) we effectively always pass validation, even if the input is incorrect. Since in the frontend the input coming into the validation service is always `SerializedValueOrExpression` if we narrow the type checking we will produce meaningful validation errors... This is a workaround, but perhaps we'd be better off using a `Union[float, DimensionedTypes, np.ndarray]` type. The issue is that numpy ndarray is incompatible with pydantic... I'll try to write a wrapper.

flow360/component/simulation/user_code.py
```diff
@@ -82,10 +85,41 @@ class SerializedValueOrExpression(Flow360BaseModel):
     evaluated_value: Optional[Union[Number, Iterable[Number]]] = pd.Field(None)
     evaluated_units: Optional[str] = pd.Field(None)
 
+# This is a wrapper to allow using ndarrays with pydantic models
```
andrzej-krupka, line 88: After narrowing the internal type of variables we don't need the `force_serialized_format` method.

flow360/component/simulation/user_code.py
andrzej-krupka: Removed in [8f160b4](https://github.com/flexcompute/Flow360/commit/8f160b45158d24a514be9c32a3582eea6e04b4a8)

benflexcompute (approved):

flow360/component/simulation/framework/base_model.py
benflexcompute: But I believe front end already adopted snake case throughout their form. Why we are now allowing them to write camel case again?

PR #1031, opened 2025-05-12
[FLPY-12] Fixed the duplicate entity check logic and allow entity reuse between RotationCylinder and UniformRefinement

benflexcompute:
…

### DIFF
--- a/flow360/component/simulation/meshing_param/params.py
+++ b/flow360/component/simulation/meshing_param/params.py
@@ -20,7 +20,6 @@ from flow360.component.simulation.meshing_param.volume_params import (
     UniformRefinement,
     UserDefinedFarfield,
 )
-from flow360.component.simulation.primitives import Cylinder
 from flow360.component.simulation.unit_system import AngleType, LengthType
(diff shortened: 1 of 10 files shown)


flow360/component/simulation/meshing_param/params.py
```diff
@@ -221,48 +221,64 @@ def _check_volume_zones_has_farfied(cls, v):
...
                 # pylint: disable=protected-access
-                for cylinder in [
-                    item
+                _ = [
+                    usage.add_entity_usage(item, refinement.refinement_type)
                     for item in refinement.entities._get_expanded_entities(create_hard_copy=False)
-                    if isinstance(item, Cylinder)
-                ]:
-                    cylinder_name_to_usage_map[cylinder.name] = refinement.refinement_type
+                ]
```
angranl-flex, line 255: Why using this way?

angranl-flex (approved):

flow360/component/simulation/meshing_param/params.py
benflexcompute: Skips a for loop.

PR #1032, opened 2025-05-12
[FLPY-14] Change the liquid Mach to 0.05 and add private_attribute_dict in simulation params

angranl-flex:


flow360/component/simulation/translator/solver_translator.py
```diff
@@ -1331,5 +1331,7 @@ def get_solver_json(
             if "MachRef" in translated["freestream"].keys()
             else 1.0 / translated["freestream"]["Mach"]
         )
+    if input_params.private_attribute_dict is not None:
+        translated["privateAttributeDict"] = input_params.private_attribute_dict
```
benflexcompute, line 1335: Let's keep the setting in root level, in the same level as `use liquid as material`.

angranl-flex: Done

PR #1049, opened 2025-05-14
Added unit handling to solver code converter

andrzej-krupka:


tests/simulation/test_expressions.py
```diff
@@ -585,16 +585,18 @@ class TestModel(Flow360BaseModel):
...
-    assert str(model.field) == "(x // 3) ** 2"
+    assert model.field.evaluate() == 17 * u.m**2
+    assert str(model.field) == "((x * u.m) * u.m) // 3 + ((x ** 2) * u.m**2)"
 
     solver_code = model.field.to_solver_code()
 
-    assert solver_code == "pow(floor(x / 3), 2)"
+    # Unit symbols are stripped and unsupported
+    # operators are changed into their C++ equivalents
+    assert solver_code == "(floor(x / 3) + pow(x, 2))"
```
benflexcompute, line 602: Why solver code treat x as a variable? Why not evaluate before giving to solver?

flow360/component/simulation/user_code.py
```diff
@@ -303,8 +303,7 @@ def user_variables(self):
 
     def to_solver_code(self):
         expr = expr_to_model(self.expression, _global_ctx)
-        source = expr_to_code(expr, TargetSyntax.CPP, _solver_variables)
-        # TODO: What do we do with dimensioned expressions? We need to replace all units by their conversion factors.
+        source = expr_to_code(expr, TargetSyntax.CPP, _solver_variables, "u.*")
```
benflexcompute, line 306: Stripping off the units does not seem to be a solution to handling unit.

tests/simulation/test_expressions.py
andrzej-krupka: We could inline them into non-evaluable types. My idea is that we could have user variable declarations generated in solver code so that its more DRY

flow360/component/simulation/user_code.py
andrzej-krupka: I mean, eventually we have to do it, right? The question is how we ensure that the proper conversion factors are introduced (?)

andrzej-krupka:
@benflexcompute I analyzed your feedback and decided change two things:

1. Generalize the client->solver symbol translator by allowing us to pass an arbitrary `str->str` function to the generator
2. Instead of stripping units we convert them to scalar factors. The exact value of the factors will depend on the target unit system.

Now we have:

`x * u.m + x * u.cm` in Python code will translate to `x * 1.0 + x * 0.01`

tests/simulation/test_expressions.py
andrzej-krupka: Added user variable inlining in [9b3c9ae](https://github.com/flexcompute/Flow360/pull/1049/commits/9b3c9aee27ddc6353c0c79c7825575b2b4d43626)

flow360/component/simulation/user_code.py
andrzej-krupka: Added unit conversion in [b6d17d9](https://github.com/flexcompute/Flow360/pull/1049/commits/b6d17d9d3ca5b7a10779945d2b89c1088fa8e5e7)

benflexcompute (approved):

PR #1063, opened 2025-05-15
Updates examples for latest release

maciej-flexcompute:


flow360/component/simulation/migration/BETDisk.py
```diff
@@ -171,9 +171,9 @@ def read_single_v1_BETDisk(
     --------
     Create a BET disk from flow360.json.
 
-    >>> param = fl.BETDisk.from_flow360(
-    ...     file_path="flow360.json",
-    ...     mesh_unit=1 * fl.u.m,
+    >>> param = BETDisk.read_single_v1_BETDisk(
```
angranl-flex, line 174: Are we going to remove all `fl.` prefix? We currently have it in all class examples.

angranl-flex (approved):

flow360/component/simulation/migration/BETDisk.py
maciej-flexcompute: actually, it is not under `fl.` prefix but in:
```
from flow360.component.simulation.migration import BETDisk
```
how this should be reflected in the example?

angranl-flex: Oh right, make sense

PR #1072, opened 2025-05-16
[FLPY-6] [FL-852] Moved velocity direction out into `Inflow` model

benflexcompute:


benflexcompute:
Will serve as the updater source.

PR #1074, opened 2025-05-19
[FL-728] Add Default Report Config for Report Summary Table

angranl-flex:


flow360/component/project.py
```diff
@@ -1356,6 +1358,14 @@ def _run(
 
         log.info(f"Successfully submitted: {destination_obj.short_description()}")
 
+        if isinstance(destination_obj, Case):
```
benflexcompute, line 1361: Let's move this to run_case() function. This part is not generic enough to be put here.

flow360/plugins/report/report.py
```diff
@@ -319,3 +321,41 @@ def create_pdf(
                             item.get_doc_item(case_context, self.settings)
 
         report_doc.generate_pdf(os.path.join(data_storage, filename))
+
+
+def get_default_report_template() -> ReportTemplate:
```
benflexcompute, line 326:
```suggestion
def get_default_report_summary_template() -> ReportTemplate:
```

flow360/plugins/report/report.py
```diff
@@ -319,3 +321,41 @@ def create_pdf(
...
+        "Reference Length",
+        "Reference Area",
+        "CL",
+        "CD",
+        "CFy",
+        "CMx",
+        "CMy",
+        "CMz",
+    ]
+    formatter = [".4f" for _ in data]
```
benflexcompute, line 359: why .4f? Can we have more precision and use scientific notation?

flow360/plugins/report/report_items.py
```diff
@@ -109,6 +110,7 @@ class Settings(Flow360BaseModel):
     """
 
     dpi: Optional[pd.PositiveInt] = 300
+    dump_table_csv: Optional[pd.StrictBool] = False
```
benflexcompute, line 113: What is the reason having a single model handling output option for all types of report item? I think we should write a TODO here to separate the setting for each type in the future.

tests/simulation/service/test_services_v2.py
```diff
@@ -1258,3 +1258,12 @@ def _get_all_units(value):
                 f"Unit {unit_system_dimension_string} (A.K.A {field_name}) is not supported by the front-end.",
                 "Please ensure front end team is aware of this new unit and add its support.",
             )
+
+
+def test_get_default_report_config_json():
+    report_config_dict = services.get_default_report_config()
+    with open("ref/default_report_config.json", "r") as fp:
+        ref_dict = json.load(fp)
+    print(report_config_dict)
```
benflexcompute, line 1267: remove print

angranl-flex: Removed

flow360/plugins/report/report.py
angranl-flex: Updated

flow360/plugins/report/report_items.py
angranl-flex: We can add more classes if we have more output options that need to be exposed?
We only have two options for now and each type of report item can just grab what it needs from this global settings anyway.

flow360/plugins/report/report.py
angranl-flex: I just used Maciej's example for this part. This formatter only affects the format in the report pdf file, the csv file will have the accurate value. Do we still want to change this?

benflexcompute (approved):

PR #1083, opened 2025-05-21
Pylint Fix for `expression` branch

benflexcompute:


flow360/component/simulation/blueprint/flow360/symbols.py
```diff
@@ -18,27 +18,13 @@ def _unit_list():
 
 
 def _import_flow360(name: str) -> Any:
-    import flow360 as fl
```
benflexcompute, line 21: These are no longer needed? They cause cyclic import.

flow360/component/simulation/blueprint/flow360/symbols.py
```diff
@@ -18,27 +18,13 @@ def _unit_list():
...
-    if name == "fl":
-        return fl
 
     if name == "u":
-        from flow360 import u
+        from flow360.component.simulation import units as u
 
         return u
 
-    if name == "control":
```
benflexcompute, line 32: Same here. Do we rely on the `context`/`resolver` to parse "solution."?

andrzej-krupka: Since it is always used from the `flow360` client context those symbols will be already imported, so I think evaluation should work fine even with those removed... Overall we want to rely on context/resolver for security reasons - so that we only expose the whitelisted symbols and nothing else.

PR #1123, opened 2025-06-04
changed betdisk example to use xrotor instead of json file

wiktor-flex:
- removed unnecessary information from the file
- added xrotor file to the examples assets
- removed disk0 and cylinder0 from examples assets
- changed the BETDisk input method to use from_xrotor

There are some discrepancies in the results between original version and this one.
- the resulting BETDisk using from_xrotor has more alphas defined between -10 and 10
- slight differences in twists and chords (might be due to translation)
- original CL = 0.125, new CL = 1
- original CD = 0.0045, new CD = 0.0031


piotrkluba (approved):
If it runs, everything is great

PR #1138, opened 2025-06-06
changed mach reynolds to use charcteristic length

wiktor-flex:
I think characteristic_length is a better name than reference_length since it won't get confused with the one for moments and is also more common definition used in defining Reynolds number globally I believe.

- project_length_unit was renamed to characteristic_length
- fixed unittests
- fixed previously ommited operating_condition_from_mach_reynolds() calls

For reference, here is a Wikipedia page where it is also called characteristic length.
image

### DIFF
--- a/flow360/component/simulation/operating_condition/operating_condition.py
+++ b/flow360/component/simulation/operating_condition/operating_condition.py
@@ -400,7 +400,8 @@ class AerospaceCondition(MultiConstructorBaseModel):
         reynolds : PositiveFloat
             Freestream Reynolds number defined with mesh unit (must be positive).
         characteristic_length: LengthType.Positive
-            Length dimension that is used to define the scale of the system, for example, chord length.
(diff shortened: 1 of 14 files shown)


flow360/component/simulation/operating_condition/operating_condition.py
```diff
@@ -399,8 +399,8 @@ def from_mach_reynolds(
             Freestream Mach number (must be non-negative).
         reynolds : PositiveFloat
             Freestream Reynolds number defined with mesh unit (must be positive).
-        project_length_unit: LengthType.Positive
-            Project length unit.
+        characteristic_length: LengthType.Positive
+            Length dimension that is used to define the scale of the system, for example, chord length.
```
maciej-flexcompute, line 403:
```suggestion
            Characteristic length scale of the system used to compute the Reynolds number
            (e.g., the chord length of an airfoil). This value must be positive.
```

PR #1142, opened 2025-06-06
Add BET Validator to check initial_blade_direction in an unsteady simulation

angranl-flex:


flow360/component/simulation/validation/validation_context.py
```diff
@@ -95,6 +96,13 @@ def _get_use_geometry_AI_(cls, param_as_dict: dict):  # pylint:disable=invalid-n
         except KeyError:
             return False
 
+    @classmethod
+    def _get_is_unsteady_simulation_(cls, param_as_dict: dict):
+        try:
+            return param_as_dict["time_stepping"]["type_name"] == "Unsteady"
```
benflexcompute, line 102: It has 3 state actually, steady, unsteady, unset.
Can we name this to _get_time_stepping and return the above enum instead?

benflexcompute: I want avoid misuse where _get_is_unsteady_simulation_ being false being used as is_steady

flow360/component/simulation/models/validation/validation_bet_disk.py
```diff
@@ -51,6 +55,18 @@ def _check_bet_disk_duplicate_twists(value, info: ValidationInfo):
     return value
 
 
+def _check_bet_disk_initial_blade_direction(value, info: ValidationInfo):
```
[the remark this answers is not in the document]
angranl-flex: The unused-argument check is disabled in line 27

flow360/component/simulation/validation/validation_context.py
angranl-flex: OK, updated the code

PR #1165, opened 2025-06-16
Separate prepending code to declaration and computation parts

angranl-flex:


flow360/component/simulation/translator/solver_translator.py
```diff
@@ -678,7 +592,11 @@ def _compute_coefficient_and_offset(source_unit: u.Unit, target_unit: u.Unit):
 
     expression_length = expression.length
     prepending_code = [
-        udf_prepending_code[name]
+        (
```
benflexcompute, line 595: Logic is a bit complicated. Can we move this to a local function and have some commenting why we do this?

flow360/component/simulation/user_code/core/context.py
```diff
@@ -103,19 +102,21 @@ def _import_solution(_) -> Any:
             "turbulence_intermittency",
             "density",
             "velocity",
+            "velocity_magnitude",
```
benflexcompute, line 105: Why do we need this? We should provide minimum set of variables here.

benflexcompute (approved):

flow360/component/simulation/user_code/core/context.py
angranl-flex: Removed velocity and vorticity's magnitude

PR #1201, opened 2025-06-26
Add expression support for Isosurface

angranl-flex:


tests/simulation/translator/test_solver_translator.py
```diff
@@ -751,16 +772,42 @@ def test_param_with_user_variables():
...
             ],
             time_stepping=Unsteady(step_size=my_time_stepping_var + 0.5 * u.s, steps=123),
         )
     # Mimicking real workflow where the Param is serialized and then deserialized
-    params_validated, _, _ = validate_model(
+    params_validated, errors, _ = validate_model(
         params_as_dict=param.model_dump(mode="json"),
         validated_by=ValidationCalledBy.LOCAL,
         root_item_type=None,
     )
```
benflexcompute, line 809: For one of the isoSurface, can you add assertion that the iso value is translated into the output units not flow360 unit?

PR #1214, opened 2025-06-30
Update SurfaceIntegralOutput for UserVariable

angranl-flex:


flow360/component/simulation/translator/solver_translator.py
```diff
@@ -260,7 +262,11 @@ def translate_output_fields(
     for output_field in output_model.output_fields.items:
         if isinstance(output_field, UserVariable):
             # Remove the UserVariable object and add its name
-            output_fields.append(output_field.name)
+            output_fields.append(
+                output_field.name
```
benflexcompute, line 266: `translate_output_fields` is a generic utility function and should not be affected by the upstream process. Can you replace all user variables at around line 846 ("if has_instance_in_list(outputs, SurfaceIntegralOutput):") instead? Then all the downstream process should work without any modification.

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -719,7 +725,16 @@ def process_output_fields_for_udf(input_params: SimulationParams):
                 for output_field in output.output_fields.items:
                     if not isinstance(output_field, UserVariable):
                         continue
-                    udf_from_user_variable = user_variable_to_udf(output_field, input_params)
+                    if isinstance(output, SurfaceIntegralOutput):
```
benflexcompute, line 728: Same here. This should be handled at the very beginning.

tests/simulation/translator/test_solver_translator.py
```diff
@@ -798,6 +801,16 @@ def test_param_with_user_variables():
                         ),
                     ],
                 ),
+                SurfaceOutput(
+                    name="surface_output",
+                    output_fields=[surface_integral_variable],
+                    entities=Surface(name="VOLUME/LEFT"),
+                ),
+                SurfaceIntegralOutput(
```
benflexcompute, line 809: swap these two to ensure SurfaceIntegralOutput won't affect latter outputs.

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -829,6 +831,33 @@ def translate_output(input_params: SimulationParams, translated: dict):
             outputs, TimeAverageProbeOutput, inject_probe_info
         )
     if has_instance_in_list(outputs, SurfaceIntegralOutput):
+        for output in outputs:
+            if not isinstance(output, SurfaceIntegralOutput):
+                continue
+            output_fields_processed = []
+            for output_field in output.output_fields.items:
+                if isinstance(output_field, UserVariable):
+                    expression = output_field.value
```
benflexcompute, line 840: `.value` does not always guarantee return `class Expression`. You can try to call Expression.model_validate to convert it to Expression though. Similar to what we did in math.py.

Also we probably should also disable constant variable being integrated here like we did for all other outputs.

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -829,6 +831,33 @@ def translate_output(input_params: SimulationParams, translated: dict):
...
+                            for i in range(expression.length)
+                        ]
+
+                    output_fields_processed.append(
+                        UserVariable(
+                            name=output_field.name + "_integral",
+                            value=expression_processed,
+                        )
+                    )
+                    continue
```
benflexcompute, line 857: use else

benflexcompute (approved):

benflexcompute (approved):

benflexcompute (approved):

PR #1220, opened 2025-07-01
More IsosurfaceOutput changes for Expression

angranl-flex:
- Support using unit system name in iso_value's unit
- Add validation to disable dimensioned iso_value when legacy string field is used.


flow360/component/simulation/outputs/output_entities.py
```diff
@@ -112,6 +113,22 @@ def _preprocess_expression_and_solver_variable(cls, value):
...
 
+    @pd.field_validator("iso_value", mode="before")
+    @classmethod
+    def _preprocess_field_with_unit_system(cls, value, info: pd.ValidationInfo):
+        if not isinstance(value, dict) or info.data.get("field") is None or "units" not in value:
+            return value
+        units = value["units"]
+        field = info.data["field"]
+        field_dimensions = get_input_value_dimensions(value=field)
+        if units in ("SI", "SI_unit_system"):
```
benflexcompute, line 124:
```suggestion
        if units in  "SI_unit_system":
```
This is a UI only feature and we should limit the allowed input.

benflexcompute (approved):

PR #1253, opened 2025-07-10
[25.5] Add prefix option to read_all_v1_BETDisks function

angranl-flex:


flow360/component/simulation/migration/BETDisk.py
```diff
@@ -15,13 +15,16 @@
...
 def _parse_flow360_bet_disk_dict(
     *,
     flow360_bet_disk_dict: dict,
     mesh_unit,
     freestream_temperature,
+    bet_disk_name: str,
     bet_disk_index: int = 0,
-    bet_disk_name: str = None,
+    add_index: bool = False,
+    index_offset: int = 0,
```
benflexcompute, line 27: Can we remove add_index and index_offset = None ==> add_index = False?

benflexcompute (approved):

flow360/component/simulation/migration/BETDisk.py
angranl-flex: Updated

benflexcompute (approved):

PR #1282, opened 2025-07-22
fix(): misuse of .sort() which return None

benflexcompute:


angranl-flex (approved):

tests/simulation/translator/test_solver_translator.py
```diff
@@ -610,7 +608,9 @@ def test_boundaries():
                 ),
             ],
         )
-    translate_and_compare(param, mesh_unit=1 * u.m, ref_json_file="Flow360_boundaries.json")
+    translate_and_compare(
+        param, mesh_unit=1 * u.m, ref_json_file="Flow360_boundaries.json", debug=True
```
angranl-flex, line 612: Should we remove the debug option or it is fine leaving it as True?

PR #1284, opened 2025-07-23
adopted the proposed flex credit example

wiktor-flex:


examples/workflow_management/project_management/total_flex_credits.py
```diff
@@ -0,0 +1,13 @@
+import flow360 as fl
+
+project_id = "prj-842fe363-fb66-4d33-85f2-384a88a448e7"  # Replace with your project ID
+
+project = fl.Project.from_cloud(project_id=project_id)
+case_ids = project.get_case_ids()
+total_fc_cost = 0
+for case in case_ids:
+    case = fl.Case.from_cloud(case_id=case)
+    print(f"    {case.name}: cost = {case.info.computeCost}")
```
piotrkluba, line 10: Why are we printing with indentation if there is no introduction to what we are printing (these are the first lines the script prints)

piotrkluba: image
looks weird

wiktor-flex: So that your case names are distinguished from the rest, but it's not necessary

wiktor-flex: Fixed

piotrkluba (approved):
ok

PR #1286, opened 2025-07-24
Fix Deserialize function when handling None description

angranl-flex:


angranl-flex (commented):
Waiting for localTest run

benflexcompute (approved):

PR #1293, opened 2025-07-25
Added updater for entities being in multiple outputs

benflexcompute:


flow360/component/simulation/framework/updater.py
```diff
@@ -208,6 +198,83 @@ def _to_25_6_2(params_as_dict):
...
+                    entity_map[entity["name"]]["entity"] = entity
+
+        for entity_info in entity_map.values():
+            if entity_info["creates_new"]:
+                for index, output in enumerate(params_as_dict["outputs"]):
+                    if output.get("output_type") != output_type:
+                        continue
+                    for entity in output["entities"]["stored_entities"]:
+                        if entity["name"] == entity_info["entity"]["name"]:
+                            params_as_dict["outputs"][index]["entities"]["stored_entities"].remove(
```
angranl-flex: remove function will only remove the first occurrence in the list. Supposing the simulation.json contains duplicate surfaces, only one of them will be removed and the validation error will be triggered again.

benflexcompute: right, we should deduplicate within a SurfaceOutput first...

angranl-flex: OK, I will implement the fix and add test

benflexcompute: Regardless yes we should deduplicate

maciej-flexcompute (approved):

PR #1333, opened 2025-08-06
[FXC-476] feat: add rename method to v2 assets

dawoodhq:
- [x] Implement renaming feature to assets (all v2)
    - [x] Folder
    - [x] Project
    - [x] SurfaceMesh
    - [x] VolumeMesh
    - [x] Case

### DIFF
--- a/flow360/component/case.py
+++ b/flow360/component/case.py
@@ -14,7 +14,11 @@ import pydantic as pd
 import pydantic.v1 as pd_v1
 
 from .. import error_messages
-from ..cloud.flow360_requests import MoveCaseItem, MoveToFolderRequest
+from ..cloud.flow360_requests import (
+    MoveCaseItem,
+    MoveToFolderRequest,
+    RenameAssetRequestV2,
+)
 from ..cloud.rest_api import RestApi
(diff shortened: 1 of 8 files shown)


flow360/component/case.py
```diff
@@ -668,6 +672,25 @@ def move_to_folder(self, folder: Folder):
...
+
+        Returns
+        -------
+        self
+            Returns the modified case after it has been renamed.
+        """
+        RestApi(CaseInterfaceV2.endpoint).patch(
+            RenameAssetRequestV2(name=new_name).dict(), method=self.id
+        )
+        return self
```
benflexcompute, line 692: Any reason we return self at the end?

flow360/component/project.py
```diff
@@ -1197,6 +1197,25 @@ def refresh_project_tree(self):
...
+        ----------
+        new_name : str
+            The new name for the project.
+
+        Returns
+        -------
+        self
+            Returns the modified project after it has been renamed.
+        """
+        RestApi(ProjectInterface.endpoint).patch(
```
benflexcompute, line 1214: Is it possible to make this a base class' method?

benflexcompute (approved):

PR #1352, opened 2025-08-11
Add output classes for imported surfaces

yifan-flex:


flow360/__init__.py
```diff
@@ -132,6 +135,7 @@
 from flow360.component.simulation.primitives import (
     Box,
     Cylinder,
+    ImportedSurface,
```
benflexcompute, line 138: Why do we expose this class?

benflexcompute (commented):
What is the plan for back end integration? (Submit imported surfaces to cloud)?

PR #1371, opened 2025-08-18
allow n_edge=0 

feilin-flexcompute:
tested using:
- [x] single ugrid 
- [x] single STL


benflexcompute:
Can you test by loading geometry from the cloud with 0 edge?

PR #1394, opened 2025-08-28
Modular meshing and Snappy

piotrkluba:
Merging the snappy python interface as well as the newly discussed ModularMeshingWorkflow to the main python branch.


flow360/component/simulation/validation/validation_context.py
```diff
@@ -170,10 +175,20 @@ def _get_using_liquid_as_material_(cls, param_as_dict: dict):
...
-        except KeyError:
-            return False
+        beta_mesher_triggers = []
+        beta_mesher_triggers.append(
+            param_as_dict.get("private_attribute_asset_cache", {}).get("use_inhouse_mesher", False)
+        )
+        if param_as_dict.get("meshing", None) is not None:
+            beta_mesher_triggers.append(
+                param_as_dict.get("meshing", {}).get("surface_meshing", {}).get("type", "")
+                == "SnappySurfaceMeshingParams"
```
benflexcompute, line 185: So the Snappy is also considered "beta mesher"?

tests/test_current_flow360_version.py
```diff
@@ -2,4 +2,4 @@
 
 
 def test_version():
-    assert __version__ == "25.7.4b0"
+    assert __version__ == "25.8.0b3"
```
benflexcompute, line 5: With the introduction of these new schema we need to bump "real" version. The beta version is not used anywhere (for example updated) in Python client. 
```suggestion
    assert __version__ == "25.8.1b3"
```

flow360/component/simulation/validation/validation_context.py
```diff
@@ -144,17 +144,22 @@ def _get_farfield_method_(cls, param_as_dict: dict):
...
+
+        farfield_method = None
         if volume_zones:
             for zone in volume_zones:
                 if zone["type"] == "AutomatedFarfield":
                     return zone["method"]
-                if zone["type"] == "UserDefinedFarfield":
-                    return "user-defined"
-        return None
+                if zone["type"] in ["UserDefinedFarfield", "CustomVolume", "SeedpointZone"]:
```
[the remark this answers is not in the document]
piotrkluba: It is written that way based on the assumption that we do not accept `CustomVolume`/`SeedpointZone` with automated farfield yet so if any zone is of `CustomVolume` or `SeedpointZone` class that means that the farfield will be a user defined one. I can change that so it searches through all the zones for the automated farfield to induce a type so that when we do support automated farfield with multizone it does not cause problems.

flow360/component/simulation/framework/entity_registry.py
```diff
@@ -7,6 +7,37 @@
 from flow360.component.simulation.framework.base_model import Flow360BaseModel
 from flow360.component.simulation.framework.entity_base import EntityBase
 from flow360.component.utils import _naming_pattern_handler
+from flow360.exceptions import Flow360ValueError
+
+
+class DoubleIndexableList(list):
```
benflexcompute: This class alone does not implement the 2 layer of `__getitem__` so I would not call it `**Double**IndexableList` which can be misleading.

Can you rename it to `StringIndexableList` instead?

flow360/component/simulation/framework/entity_registry.py
```diff
@@ -224,3 +255,48 @@ def find_by_asset_id(self, *, entity_id: str, entity_class: type[EntityBase]):
     def is_empty(self):
         """Return True if the registry is empty, False otherwise."""
         return not self.internal_registry
+
+
+class SnappyBodyRegistry(EntityRegistry):
+    """
+    Extension of Entityregistry to be used with SnappyBody, allows double indexing.
```
benflexcompute:
```suggestion
    Extension of Entityregistry to be used with SnappyBody, allows double indexing for accessing the boundaries under certain `SnappyBody`.
```

flow360/component/simulation/framework/entity_registry.py
```diff
@@ -224,3 +255,48 @@ def find_by_asset_id(self, *, entity_id: str, entity_class: type[EntityBase]):
...
+    """
+
+    def find_by_naming_pattern(
+        self, pattern: str, enforce_output_as_list: bool = True, error_when_no_match: bool = False
+    ) -> list[EntityBase]:
```
benflexcompute: Wrong type hint. This returns a `DoubleIndexableList`

flow360/component/simulation/meshing_param/snappy/snappy_mesh_refinements.py
```diff
@@ -0,0 +1,176 @@
...
+    Refinement for snappyHexMesh body (searchableSurfaceWithGaps).
+
+    Parameters
+    ----------
+    gap_resolution: Optional[LengthType.NonNegative], default: None
+    """
+
+    # pylint: disable=no-member
+    refinement_type: Literal["SnappyBodyRefinement"] = pd.Field("SnappyBodyRefinement", frozen=True)
+    gap_resolution: Optional[LengthType.NonNegative] = pd.Field(None)
```
benflexcompute: What is the point if everything other than the entity is None? Is this expected?

flow360/component/simulation/meshing_param/snappy/snappy_mesh_refinements.py
```diff
@@ -0,0 +1,176 @@
...
+        "SnappySurfaceRefinement", frozen=True
+    )
+    entities: EntityList[Surface] = pd.Field(alias="regions")
+
+
+class SurfaceEdgeRefinement(Flow360BaseModel):
+    """
+    Edge refinement for bodies and regions in snappyHexMesh.
+
+    Parameters
```
benflexcompute: Again this is the wrong place to write documentation.

flow360/component/simulation/meshing_param/snappy/snappy_mesh_refinements.py
```diff
@@ -0,0 +1,176 @@
...
+            spacing_state = (True, len(self.spacing))
+        except TypeError:
+            # spacing is a scalar
+            spacing_state = (False, 1)
+
+        if (
+            distances_state[0] and spacing_state[0] and (spacing_state[1] != distances_state[1])
+        ) or (distances_state[0] is not spacing_state[0]):
+            raise ValueError(
+                f"When using a distance spacing specification both spacing ({self.spacing}) and distances"
```
benflexcompute: Add a unit test for this.

flow360/component/simulation/meshing_param/snappy/snappy_params.py
```diff
@@ -0,0 +1,144 @@
...
+    def _check_body_refinements_w_defaults(self):
+        # set body refinements
+        # pylint: disable=no-member
+        if self.refinements is None:
+            return self
+        for refinement in self.refinements:
+            if isinstance(refinement, BodyRefinement):
+                if refinement.min_spacing is None and refinement.max_spacing is None:
+                    continue
+                if refinement.min_spacing is None and self.defaults.min_spacing.to(
```
benflexcompute: I recommend putting self.defaults.min_spacing into `ParamsValidationInfo` and make these field validator instead to give better error location but this is up to you.

flow360/component/simulation/meshing_param/snappy/snappy_params.py
```diff
@@ -0,0 +1,144 @@
...
+                    "m"
+                ) < refinement.min_spacing.to("m"):
+                    raise ValueError(
+                        "Default maximum spacing is lower that refinement minimum spacing"
+                        + "and maximum spacing is not provided."
+                    )
+        return self
+
+    @pd.model_validator(mode="after")
+    def _check_uniform_refinement_entities(self):
```
benflexcompute: Missing unit test.

flow360/component/simulation/meshing_param/snappy/snappy_params.py
```diff
@@ -0,0 +1,144 @@
...
+
+        def check_spacing(spacing, location):
+            # pylint: disable=no-member
+            lvl, close = self.base_spacing.to_level(spacing)
+            spacing_unit = spacing.units
+            if not close:
+                closest_spacing = self.base_spacing[lvl]
+                msg = f"The spacing of {spacing:.4g} spcified in {location} will be cast to the first lower refinement"
+                msg += f" in the octree series which is {closest_spacing.to(spacing_unit):.4g}."
+                log.warning(msg)
```
benflexcompute: again just FYI these will not appear on webUI.

flow360/component/simulation/meshing_param/snappy/snappy_params.py
```diff
@@ -0,0 +1,144 @@
...
+        # pylint: disable=no-member
+        check_spacing(self.defaults.min_spacing, "defaults")
+        check_spacing(self.defaults.max_spacing, "defaults")
+
+        if self.refinements is not None:
+            # pylint: disable=not-an-iterable
+            for refinement in self.refinements:
+                if isinstance(refinement, SnappyEntityRefinement):
+                    if refinement.min_spacing is not None:
+                        check_spacing(refinement.min_spacing, type(refinement).__name__)
```
benflexcompute: This can also be adapted as field validator. But it is up to you on the decision. On the other hand introducing too many members in the context may not be easy to manage in the future.

flow360/component/simulation/meshing_param/snappy/snappy_specs.py
```diff
@@ -0,0 +1,258 @@
...
+
+    # pylint: disable=no-member
+    min_spacing: LengthType.Positive = pd.Field()
+    max_spacing: LengthType.Positive = pd.Field()
+    gap_resolution: LengthType.Positive = pd.Field()
+
+    @pd.model_validator(mode="after")
+    def _check_spacing_order(self) -> Self:
+        if self.min_spacing and self.max_spacing:
+            if self.min_spacing > self.max_spacing:
```
benflexcompute:
```suggestion
            if self.min_spacing >= self.max_spacing:
```
?

flow360/component/simulation/meshing_param/snappy/snappy_specs.py
```diff
@@ -0,0 +1,258 @@
...
+
+    min_vol_collapse_ratio : float, default: 0
+        Minimum volume collapse ratio. If > 0: preserves single cells with all points
+        on the surface if the resulting volume after snapping is larger than
+        min_vol_collapse_ratio times the old volume (i.e., not collapsed to flat cell).
+        If < 0: always deletes such cells.
+    """
+
+    # pylint: disable=no-member
+    max_non_ortho: Optional[AngleType.Positive] = pd.Field(default=85 * u.deg)
```
benflexcompute: What is the decision or criterion on using abbreviation? We do not have many abbreviation right now. It feels inconsistent using snappy with abbreviations. Is this exactly how Snappy defines the parameter?

flow360/component/simulation/meshing_param/snappy/snappy_specs.py
```diff
@@ -0,0 +1,258 @@
...
+    max_concave: Optional[AngleType.Positive] = pd.Field(default=50 * u.deg)
+    min_vol: Optional[float] = pd.Field(default=None)
+    min_tet_quality: Optional[float] = pd.Field(default=None)
+    min_area: Optional[AreaType.Positive] = pd.Field(default=None)
+    min_twist: Optional[float] = pd.Field(default=None)
+    min_determinant: Optional[float] = pd.Field(default=None)
+    min_vol_ratio: Optional[float] = pd.Field(default=0)
+    min_face_weight: Optional[float] = pd.Field(default=0)
+    min_triangle_twist: Optional[float] = pd.Field(default=None)
+    n_smooth_scale: Optional[pd.NonNegativeInt] = pd.Field(default=4, ge=0)
```
benflexcompute: duplicate range specification.

flow360/component/simulation/meshing_param/snappy/snappy_specs.py
```diff
@@ -0,0 +1,258 @@
...
+    error_reduction: Optional[float] = pd.Field(default=0.75, ge=0, le=1)
+    min_vol_collapse_ratio: Optional[float] = pd.Field(0)
+
+    @pd.field_validator("max_non_ortho", "max_concave", mode="after")
+    @classmethod
+    def disable_angle_metrics_w_defaults(cls, value):
+        """Disable a quality metric in OpenFOAM by setting a specific value."""
+        if value is None:
+            return 180 * u.deg
```
benflexcompute: Why not adding this during translator? It seems the customer does not need to know this value anyway.

Let's avoid changing value in validator as much as possible. When customer submits the params and found the value has been modified by us from None to 180 degree wouldn't he be confused?

flow360/component/simulation/meshing_param/snappy/snappy_specs.py
```diff
@@ -0,0 +1,258 @@
...
+        if value > 180 * u.deg:
+            raise ValueError("Value must be less that 180 degrees.")
```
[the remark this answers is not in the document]
benflexcompute: Also we should round the value within 0-360 before comparision?

flow360/component/simulation/meshing_param/snappy/snappy_specs.py
```diff
@@ -0,0 +1,258 @@
...
+    n_cells_between_levels: Optional[pd.NonNegativeInt] = pd.Field(1)
+    min_refinement_cells: Optional[pd.NonNegativeInt] = pd.Field(10)
+
+    @pd.field_validator("resolve_feature_angle", mode="after")
+    @classmethod
+    def angle_limits(cls, value):
+        """Limit angular values."""
+        if value is None:
+            return value
+        if value > 180 * u.deg:
```
benflexcompute: Same here. Let's round the value between 0-360 degrees first.

flow360/component/simulation/meshing_param/snappy/snappy_specs.py
```diff
@@ -0,0 +1,258 @@
...
+
+    strict_region_snap: bool, default: False
+        Attract points only to the surface they originate from. This can improve snapping of
+        intersecting surfaces.
+    """
+
+    # pylint: disable=no-member
+    n_smooth_patch: pd.NonNegativeInt = pd.Field(3)
+    tolerance: pd.PositiveFloat = pd.Field(2)
+    n_solve_iter: pd.NonNegativeInt = pd.Field(30)
```
benflexcompute: The abbreviation here is not consistent to for example here: https://github.com/flexcompute/Flow360/blob/f35c63a44710f1a73c4402611e063293c74f4d47/flow360/component/simulation/models/solver_numerics.py#L49

flow360/component/simulation/meshing_param/meshing_specs.py
```diff
@@ -0,0 +1,242 @@
...
+    @pd.model_validator(mode="before")
+    @classmethod
+    def _project_spacing_to_object(cls, input_data):
+        if isinstance(input_data, u.unyt.unyt_quantity):
+            return {"base_spacing": input_data}
+        return input_data
+
+    @pd.validate_call
+    def __getitem__(self, idx: int):
+        return self.base_spacing * (2 ** (-idx))
```
benflexcompute: can we limit they type and range for idx?

flow360/component/simulation/meshing_param/volume_params.py
```diff
@@ -472,6 +473,7 @@ class UserDefinedFarfield(_FarfieldBase):
     ====
     """
 
+    # pylint: disable=no-member
```
benflexcompute: remove.

flow360/component/simulation/meshing_param/params.py
```diff
@@ -55,175 +57,30 @@
         RotationCylinder,
         AutomatedFarfield,
         UserDefinedFarfield,
+        SeedpointZone,
```
benflexcompute: Why add it here? Non-snappy does not support this.

flow360/component/simulation/meshing_param/params.py
```diff
@@ -412,6 +271,239 @@ def farfield_method(self):
...
+class ModularMeshingWorkflow(Flow360BaseModel):
+    """
+    Structure consolidating surface and volume meshing parameters.
+    """
+
+    type_name: Literal["ModularMeshingWorkflow"] = pd.Field("ModularMeshingWorkflow", frozen=True)
+    surface_meshing: Optional[SurfaceMeshingParams] = ContextField(
+        default=None, context=SURFACE_MESH
+    )
+    volume_meshing: Optional[VolumeMeshingParams] = ContextField(default=None, context=VOLUME_MESH)
```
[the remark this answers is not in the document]
benflexcompute: If so I think it is worth rolling out the new interface and deprecate the old one as soon as possible.

flow360/component/simulation/models/volume_models.py
```diff
@@ -1217,12 +1218,12 @@ class Rotation(Flow360BaseModel):
...
-    parent_volume: Optional[Union[GenericVolume, Cylinder, AxisymmetricBody, CustomVolume]] = (
-        pd.Field(
-            None,
-            description="The parent rotating entity in a nested rotation case."
-            + "The entity should be :class:`Cylinder` or :class:`AxisymmetricBody` or :class:`GenericVolume` type.",
-        )
+    parent_volume: Optional[
+        Union[GenericVolume, Cylinder, CustomVolume, SeedpointZone, AxisymmetricBody]
+    ] = pd.Field(
+        None,
```
benflexcompute: please add discriminator here.

benflexcompute (commented):
Done with part 1 review

flow360/component/simulation/translator/volume_meshing_translator.py
```diff
@@ -227,59 +234,95 @@ def _get_custom_volumes(volume_zones: list):
...
             ["meshing", "refinement_factor"],
         )
-    translated["refinementFactor"] = meshing_params.refinement_factor
+    translated["refinementFactor"] = refinement_factor
 
     ##::  Step 2:  Get farfield
-    for zone in input_params.meshing.volume_zones:
-        if isinstance(zone, UserDefinedFarfield):
+    for zone in volume_zones:
+        if isinstance(zone, (UserDefinedFarfield, CustomZones, SeedpointZone)):
```
benflexcompute: Would this break for non-modular meshing workflow? CustomZones will overshadow automated farfield?

flow360/component/simulation/translator/volume_meshing_translator.py
```diff
@@ -405,8 +448,16 @@ def get_volume_meshing_json(input_params: SimulationParams, mesh_units):
...
     ##::  Step 6: Get custom volumes
-    custom_volumes = _get_custom_volumes(meshing_params.volume_zones)
+    custom_volumes = _get_custom_volumes(volume_zones)
     if custom_volumes:
         translated["zones"] = custom_volumes
 
+    ##::  Step 7: Get custom seedpoint zones
+    if isinstance(input_params.meshing, ModularMeshingWorkflow) and isinstance(
+        input_params.meshing.surface_meshing, snappy.SurfaceMeshingParams
+    ):
```
benflexcompute: Is this equivalent to checking if the user is trying to use snappy?

If so can we have this as a function and reuse the same `use_snappy()` check everywhere?

flow360/component/simulation/validation/validation_context.py
```diff
@@ -146,17 +146,22 @@ def _get_farfield_method_(cls, param_as_dict: dict):
         volume_zones = None
         try:
             if param_as_dict["meshing"]:
-                volume_zones = param_as_dict["meshing"]["volume_zones"]
+                if param_as_dict["meshing"]["type_name"] == "MeshingParams":
+                    volume_zones = param_as_dict["meshing"]["volume_zones"]
+                else:
+                    volume_zones = param_as_dict["meshing"]["zones"]
```
benflexcompute: Just curious why do we want to call it "zones" over "volume_zones"?

flow360/component/simulation/validation/validation_context.py
```diff
@@ -146,17 +146,22 @@ def _get_farfield_method_(cls, param_as_dict: dict):
...
+        farfield_method = None
         if volume_zones:
             for zone in volume_zones:
                 if zone["type"] == "AutomatedFarfield":
                     return zone["method"]
-                if zone["type"] == "UserDefinedFarfield":
-                    return "user-defined"
-        return None
+                if zone["type"] in ["UserDefinedFarfield", "CustomZones", "SeedpointZone"]:
+                    farfield_method = "user-defined"
```
benflexcompute: Again this will impact current workflow. Can we try to make sure that Snappy won't change behavior of current features?

flow360/component/simulation/unit_system.py
```diff
@@ -121,6 +120,14 @@ def _dimensioned_type_serializer(x):
     return {"value": _encode_ndarray(x.value), "units": str(x.units.expr)}
 
 
+def _dimensioned_vector_type_serializer(x):
+    """
+    encoder for dimensioned type (unyt_quantity, unyt_array, DimensionedType)
```
benflexcompute: Can you update the docstring explaining why the addition of this is necessary when we already have `_dimensioned_type_serializer`?

flow360/component/geometry.py
```diff
@@ -412,10 +427,25 @@ def group_bodies_by_tag(self, tag_name: str) -> None:
...
+        )
+        # pylint: disable=protected-access
+        self.snappy_body_registry = self._entity_info._group_faces_by_snappy_format()
+
     def reset_face_grouping(self) -> None:
         """Reset the face grouping"""
         # pylint: disable=protected-access,no-member
         self.internal_registry = self._entity_info._reset_grouping("face", self.internal_registry)
+        if self.snappy_body_registry is not None:
+            self.snappy_body_registry = self.snappy_body._reset_grouping(
```
benflexcompute: Is it possible for snappy to use non-faceID grouping? I remember last time you told me that when using snappy it is equivalent to use faceID for grouping? We just provide a SnappyBody interface for managing the surfaces?

flow360/component/simulation/primitives.py
```diff
@@ -735,6 +755,65 @@ def __str__(self):
         return ",".join(sorted([self.pair[0].name, self.pair[1].name]))
 
 
+class SnappyBody(EntityBase):
```
benflexcompute: Now that I review this class it is very similar to the CustomVolume. Is the only difference:
1. Patterned surface names vs arbitrary surface names.
2. Predefined enclosing surfaces v.s. User selected surfaces
?

flow360/component/geometry.py
benflexcompute: Actually where is this face grouping used?

flow360/component/simulation/meshing_param/snappy/snappy_params.py
piotrkluba: I do not want to overload `ParamsValidationInfo` with mesher-specific parameters, I will improve the error message

flow360/component/simulation/meshing_param/snappy/snappy_params.py
piotrkluba: When we discussed with @maciej-flexcompute we came to a conclusion that the base_spacing will heve to be a parameter for both surface meshing params and volume meshing params separately as not every algorithm will use them so introducing base spacing into context info is not a very good option

flow360/component/simulation/meshing_param/params.py
```diff
@@ -412,6 +271,239 @@ def farfield_method(self):
...
+            isinstance(volume_zone, AutomatedFarfield) for volume_zone in v
+        )
+        total_user_defined_farfield = sum(
+            isinstance(volume_zone, UserDefinedFarfield) for volume_zone in v
+        )
+        total_custom_zones = sum(isinstance(volume_zone, CustomZones) for volume_zone in v)
+        total_seedpoint_zone = sum(isinstance(volume_zone, SeedpointZone) for volume_zone in v)
+
+        if (total_custom_zones or total_seedpoint_zone) and total_user_defined_farfield:
+            raise ValueError(
```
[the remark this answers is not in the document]
piotrkluba: added `test_modular_workflow_zones_validation`

flow360/component/simulation/meshing_param/params.py
```diff
@@ -412,6 +271,239 @@ def farfield_method(self):
...
+    def _check_snappy_zones(self) -> Self:
+        if isinstance(self.surface_meshing, snappy.SurfaceMeshingParams):
+            if self.farfield_method != "auto" and not sum(
+                isinstance(volume_zone, SeedpointZone) for volume_zone in self.zones
+            ):
+                raise ValueError(
+                    "snappyHexMeshing requires at least one SeedpointZone when not using AutomatedFarfield."
+                )
+            for zone in self.zones:  # pylint: disable=not-an-iterable
+                if isinstance(zone, CustomZones):
```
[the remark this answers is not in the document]
piotrkluba: No, there is only snappy as an example for now, I think GAI should be next

flow360/component/simulation/translator/surface_meshing_translator.py
```diff
@@ -59,20 +78,416 @@ def SurfaceRefinement_to_faces(obj: SurfaceRefinement, global_max_edge_length):
...
+            "max": remove_numerical_noise_from_spacing(
+                surface_meshing_params.defaults.max_spacing, spacing_system
+            ).value.item(),
+        },
+    }
+    translated["geometry"] = {
+        "bodies": [
+            {
+                "bodyName": name,
+                **deepcopy(common_defaults),
```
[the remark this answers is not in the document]
piotrkluba: because normal copy does not work for some reason

flow360/component/simulation/translator/volume_meshing_translator.py
```diff
@@ -227,59 +234,95 @@ def _get_custom_volumes(volume_zones: list):
...
+    To be later filled with data from snappyHexMesh.
+    """
+    seedpoint_zones = []
+    for zone in volume_zones:
+        if isinstance(zone, SeedpointZone) or (
+            isinstance(zone, UserDefinedFarfield) and hasattr(zone, "point_in_mesh")
+        ):
+            seedpoint_zones.append(
+                {
+                    "name": zone.name,
```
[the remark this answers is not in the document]
piotrkluba: this is a relict, deleted

flow360/component/simulation/validation/validation_context.py
piotrkluba: Modified so its non-breaking. But still I would not require `UserDefinedFarfield` if multizone is used

flow360/component/simulation/validation/validation_simulation_params.py
```diff
@@ -412,7 +422,14 @@ def _check_complete_boundary_condition_and_unknown_surface(
...
 
-    if missing_boundaries:
+    ## disable missing boundaries with snappy multizone
+    snappy_multizone = False
+    for zone in volume_zones:
+        if isinstance(zone, SeedpointZone):
+            snappy_multizone = True
+            break
+
+    if missing_boundaries and not snappy_multizone:
```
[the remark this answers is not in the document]
piotrkluba: Yes

piotrkluba:
Three general comments to the review: 
1. Abbreviations - if some parameter names are shortened it is so they resemble the names of parameters from OpenFOAM directly.
2. Some validators in snappy_specs are changing user input - This is to allow the users to disable a QualityMetric by inputing None into it. The problem is that we want to have different defaults for some of the metrics than None and If we left it as none and dealt with the change in the translator the params would delete all nones and reconstructed them as defaults when sending to cloud. We also do not want to make the users disable the metrics as in OpenFOAM as there are different ways to disable different metrics.
3. The beta_mesher_triggers are there to automatically detect if the beta mesher should be used without the user having to input that into run_case. I have already a feature branch prepared to submit a PR for that functionality.

flow360/component/simulation/meshing_param/snappy/snappy_mesh_refinements.py
benflexcompute: What I am saying is that all attributes in parent class are `Optional`.
So the user technically can do
```Python
BodyRefinement(entities=geometry["body"])
```
And it looks strange to me. Are you sure we do not want to add validation for this?

flow360/component/simulation/meshing_param/snappy/snappy_params.py
benflexcompute: We will have to add a warning collector ourselves and apply to validate_model().

flow360/component/simulation/unit_system.py
benflexcompute: If so can you delete the `_dimensioned_vector_type_serializer()`?

flow360/component/simulation/validation/validation_context.py
benflexcompute: That is fine. I just want to make sure the non Snappy mesher does not break.

flow360/component/simulation/meshing_param/snappy/snappy_mesh_refinements.py
piotrkluba: In theory I could, but I wouldnt raise nothing more than a warning as such empty refinement is not breaking anything

flow360/component/simulation/meshing_param/params.py
```diff
@@ -412,6 +271,239 @@ def farfield_method(self):
             for zone in self.volume_zones:  # pylint: disable=not-an-iterable
                 if isinstance(zone, AutomatedFarfield):
                     return zone.method
-                if isinstance(zone, UserDefinedFarfield):
-                    return "user-defined"
+            return "user-defined"
```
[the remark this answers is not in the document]
piotrkluba: I see, forgot to roll it back in `MeshingParams`, corrected

flow360/component/simulation/translator/volume_meshing_translator.py
piotrkluba: It is the `ModularMeshingWorkflow` logic, but for the `MeshingParams` the logic is still valid. If it encounters `AutomatedFarfield` it sets the parameter and breaks the loop so it will be set always correctly, If it does not encounter the `AutomatedFarfield` It will set user-defined as it is the only other option (the loop does not break after setting user-defined so the `AutomatedFarfield` can still overwrite it).

flow360/component/simulation/unit_system.py
piotrkluba: Hmm, that must have gotten handled in some weird way because I was getting arrays of size 0 that were turned to floats later on. But it works now so I will delete the other serializer.

PR #1397, opened 2025-08-28
Streamline output fields and time averaging

johnmoore4:


flow360/component/simulation/outputs/outputs.py
```diff
@@ -1099,9 +1099,78 @@ class StreamlineOutput(Flow360BaseModel):
         + ":class:`~flow360.PointArray2D` "
         + "is used to define streamline originating from a parallelogram.",
     )
+    output_fields: Optional[UniqueItemList[Union[VolumeFieldNames, str]]] = pd.Field(
```
benflexcompute, line 1102: For new output types can we start to support ONLY the UserVariables?

flow360/component/simulation/translator/solver_translator.py
```diff
@@ -752,11 +753,27 @@ def process_output_fields_for_udf(input_params: SimulationParams):
...
+        "startAverageIntegrationStep": -1,
+        "animationFrequencyTimeAverage": -1,
+        "animationFrequencyTimeAverageOffset": 0,
+    }
     for output in output_params:
-        if isinstance(output, StreamlineOutput):
+        if isinstance(output, streamline_class):
+            streamline_output["outputFields"].extend(output.output_fields.items)
+            if isinstance(output, TimeAverageStreamlineOutput):
+                streamline_output["computeTimeAverages"] = True
```
benflexcompute, line 774: Do solver still need this key? It should be obvious that this is a time average output as indicated by the name?

flow360/component/simulation/translator/solver_translator.py
benflexcompute, line 1: Do we have local test for this? It would be nice to add the resulting Flow360.json into the reference data base so we make sure that the translator is tested with streamline outputs.

flow360/component/simulation/translator/solver_translator.py
johnmoore4: The solver does need the key, but it could be added on the solver side.

flow360/component/simulation/outputs/output_fields.py
```diff
@@ -111,6 +112,7 @@
     "velocity_z_m_per_s",
     "velocity_magnitude_m_per_s",
     "pressure_pa",
+    "helicity",
```
benflexcompute, line 115: Is this recently added from solver side?

benflexcompute: It won't be usable now that the output_fields changed to Variable only.

johnmoore4: I will remove it. It is not needed.

PR #1416, opened 2025-09-09
Jim/update dynamic derivatives

jbungenerflex:
I have updated the dynamic derivatives example to use Adaptive CFL instead of ramp CFL. This allows us to do away with the steady state initialization step. 
It makes the whole process quite simpler.

See matching PR in documentation

### DIFF
--- a/examples/advanced_simulations/aerodynamics/dynamic_derivatives/dynamic_derivatives.py
+++ b/examples/advanced_simulations/aerodynamics/dynamic_derivatives/dynamic_derivatives.py
@@ -1,3 +1,9 @@
+'''
+Example script for calculating dynamic derivatives using sliding interfaces in Flow360
+See documentation for more information:
+ https://docs.flexcompute.com/projects/flow360/en/latest/tutorials/DynamicDerivatives/DynamicDerivatives.html
+'''
+
 import math
 
 import flow360 as fl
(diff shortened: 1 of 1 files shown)


examples/advanced_simulations/aerodynamics/dynamic_derivatives/dynamic_derivatives.py
```diff
@@ -1,3 +1,9 @@
+'''
+Example script for calculating dynamic derivatives using sliding interfaces in Flow360
+See documentation for more information:
+ https://docs.flexcompute.com/projects/flow360/en/latest/tutorials/DynamicDerivatives/DynamicDerivatives.html
+'''
+
```
wiktor-flex, line 6: I think pretty much all other examples do not have a docstring like that, is there a reason for adding this particular one?
I don't dislike the idea but I would prefer consistency.

wiktor-flex (commented):
Remember to make sure that the code passes unit tests and style checks like black

examples/advanced_simulations/aerodynamics/dynamic_derivatives/dynamic_derivatives.py
jbungenerflex: I would like to keep it b/c I think it adds value. Especially for examples that have a matching section in the documentation.

wiktor-flex: Okay, I respect that

wiktor-flex (approved):

PR #1426, opened 2025-09-17
Fix surface output translation with boundaries eaten by mesher

piotrkluba:
Fixed the translation of surface output to solver.json in the case of some patches that were present in the geometry, disappearing in the meshing process.


flow360/component/simulation/translator/solver_translator.py
```diff
@@ -508,6 +508,9 @@ def translate_surface_output(
     surface_output["surfaces"] = translate_setting_and_apply_to_all_entities(
         output_params,
         surface_output_class,
+        entity_injection_func=lambda entity: (
```
benflexcompute, line 511: Why is this needed? I tried comment it out and the test still passes?

PR #1437, opened 2025-09-23
pythonUI of imported surface output

feilin-flexcompute:
https://flow360.atlassian.net/browse/FXC-2987


flow360/component/project.py
```diff
@@ -1439,6 +1443,12 @@ def _run(
         params.pre_submit_summary()
 
         draft.update_simulation_params(params)
+        imported_surface_file_paths = []
```
benflexcompute, line 1446: Can we warp this in a function to reduce size of _run()?

Also we need to check the file permission and file existence here and raise if file does not exist or not readable.

feilin-flexcompute: done

flow360/component/simulation/web/draft.py
```diff
@@ -152,7 +152,12 @@ def upload_imported_surfaces(self, file_paths):
             method="imported-surfaces",
         )
         for index, local_file_path in enumerate(file_paths):
-            self._upload_file(resp[index]["filename"], local_file_path)
+            if os.path.exists(local_file_path) and os.access(local_file_path, os.R_OK):
+                self._upload_file(resp[index]["filename"], local_file_path)
+            else:
+                raise Flow360RuntimeError(
```
benflexcompute, line 158: Can we split the two scenario into different messages?

benflexcompute (approved):

flow360/component/simulation/web/draft.py
feilin-flexcompute: addressed.

PR #1445, opened 2025-09-26
fix(): fix Chart2D's include/exclude is always None

JunchengXue:


piotrkluba:
This change was made to deprecate the include/exclude parameters in Chart2D and transfer that responsibility to the DataItem class. What this validator does is it converts the x and y attributes to DataItems with the include/exclude specified to the Chart2D. So the arguments of the chart should be None. The include/exclude should be transferred to x and y

JunchengXue:
> This change was made to deprecate the include/exclude parameters in Chart2D and transfer that responsibility to the DataItem class. What this validator does is it converts the x and y attributes to DataItems with the include/exclude specified to the Chart2D. So the arguments of the chart should be None. The include/exclude should be transferred to x and y

Ok. I will do the changes in the Chart2D `_get_background_chart` method to fix the issue.

PR #1465, opened 2025-10-07
Two level structure for geometry referencing snappy

piotrkluba:
- face grouping for snappy
- two level geometry grouping (snappyBody/region)
- wildcard support for both levels
- messages with geometry missing or not found

### DIFF
--- a/flow360/component/geometry.py
+++ b/flow360/component/geometry.py
@@ -412,6 +412,9 @@ class Geometry(AssetBase):
             "body", tag_name, self.internal_registry
         )
 
+    def group_faces_for_snappy(self) -> None:
+        self.internal_registry = self._entity_info._group_faces_by_snappy_format(self.internal_registry)
+
     def reset_face_grouping(self) -> None:
         """Reset the face grouping"""
         # pylint: disable=protected-access,no-member
--- a/flow360/component/simulation/framework/entity_registry.py
+++ b/flow360/component/simulation/framework/entity_registry.py
@@ -8,6 +8,28 @@ from flow360.component.simulation.framework.base_model import Flow360BaseModel
 from flow360.component.simulation.framework.entity_base import EntityBase
 from flow360.component.utils import _naming_pattern_handler
 
+class DoubleIndexableList(list):
+    def __getitem__(self, key: Union[str, slice, int]):
+        if isinstance(key, str):
+            returned_items = []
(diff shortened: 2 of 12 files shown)


flow360/component/geometry.py
```diff
@@ -263,6 +268,16 @@ def body_group_tag(self):
     def body_group_tag(self, new_value: str):
         raise SyntaxError("Cannot set body_group_tag, use group_bodies_by_tag() instead.")
 
+    @property
+    def snappy_bodies(self):
+        """Getter for the snappy registry."""
+        if hasattr(self, "snappy_body_registry") is False or self.snappy_body_registry is None:
```
benflexcompute: Since you changed the constructor to always populate self.snappy_body_registry as None, the first statement will overshadowed?

flow360/component/simulation/framework/entity_registry.py
```diff
@@ -7,6 +7,37 @@
 from flow360.component.simulation.framework.base_model import Flow360BaseModel
 from flow360.component.simulation.framework.entity_base import EntityBase
 from flow360.component.utils import _naming_pattern_handler
+from flow360.exceptions import Flow360ValueError
+
+
+class DoubleIndexableList(list):
```
benflexcompute: The addition of this DoubleIndexableList seems redundant? See my other comment.

flow360/component/simulation/primitives.py
```diff
@@ -716,14 +717,33 @@ def __str__(self):
...
 
-    body_name: str = pd.Field()
+    private_attribute_registry_bucket_name: Literal["SurfaceGroupedEntityType"] = pd.Field(
+        "SurfaceGroupedEntityType", frozen=True
+    )
+    private_attribute_entity_type_name: Literal["SnappyBody"] = pd.Field("SnappyBody", frozen=True)
+
+    surfaces: List[Surface] = pd.Field()
+
+    def __getitem__(self, key: str):
```
benflexcompute: Why do we still need to get `Surface` entities from SnappyBody?
To select a surface with pattern `patch*` under SnappyBodies with pattern `tunne*` we can just do:
```python
my_geo["tunne*::patch*"]
```
Why do user want to instead do:
```python
my_geo_snappy_body["tunne*"]["patch*"]
```
which is longer?

PR #1998, opened 2026-04-17
Relay migrated simulation schema services from flow360-schema

benflexcompute:
## What changed
- converted more of `flow360/component/simulation/` into thin relays over schema-owned implementations
- removed duplicate client implementations for migrated schema-owned logic, including validation service entry points, schema service helpers, updater helpers, conversion helpers, utils, and `services_utils`
- removed duplicate client tests that now belong to schema ownership in the paired schema PR

## Why
The migration target is for the client to keep runtime-only responsibilities such as translator orchestration, web API code, BET wrappers, and unit-switching, while schema owns Pydantic behavior, validation, and deserialization.

## Impact
- client-side duplication is reduced substantially in this batch
- schema-owned tests now live with the code they validate
- the remaining client work is more clearly limited to runtime orchestration and a smaller set of residual case-by-case migrations

## Validation
- `python -m py_compile flow360/component/simulation/services.py flow360/component/simulation/services_utils.py flow360/component/simulation/conversion.py flow360/component/simulation/utils.py flow360/component/simulation/framework/updater.py flow360/component/simulation/framework/updater_functions.py flow360/component/simulation/framework/updater_utils.py`
- targeted schema-side test suites were run in the paired schema PR for the migrated behavior
- targeted client smoke tests were run during the migration for relay boundaries, but a full client pytest run was not used for this batch because the local client environment still had an older installed `flow360-schema` package during development

## Paired Schema PR
- flexcompute/flex#11404

## Remaining Work From Plan
- audit and shrink legacy `unit_system.py` / `exposed_units.py` only after confirming real remaining consumers
- keep translator, BET request wrappers, web API code, process-json generation, and unit-switching orchestration on the client side
- continue the remaining case-by-case cleanup in client tests and any residual helper boundaries, especially the still-mixed portions of `tests/simulation/params/test_validators_params.py` and `tests/simulation/services/test_selector_expansion_validators.py`
- leave draft/project/web/runtime workflow tests on the client side unless a specific case proves to be pure schema behavior

## Follow-up
This PR is intentionally scoped as the first reviewable client-side relay cleanup batch. The remaining items should be finished in a follow-up PR rather than expanding this one further.


benflexcompute:
Superseded by #1999.

This batch was re-rooted onto `main` because changing the merge base from `BenY/CentralizedSchema` to `main` pulled in the historical stacked-branch ancestry and produced a noisy/conflicted diff. Please review #1999 instead.

PR #2023, opened 2026-04-29
[FXC-6350] feat: expose surface mesh statistics on SurfaceMeshV2

mikeparkflex:
## Summary

Adds a `stats` property on `SurfaceMeshV2` returning a `SurfaceMeshStats` pydantic model with `n_nodes`, `n_triangles`, `n_quadrilaterals`, and `version` — mirroring the existing `VolumeMeshV2.stats` / `VolumeMeshStats` pattern.

The webui already exposes these values (via `projectApi.getSurfaceMeshStatistics` in `flex/frontend/flow360-ui-next/src/apis/endpoints/project.ts`), reading the same `metadata/surfaceMeshMetaData.json` artifact this PR consumes.

Replaces the TODO at `flow360/component/surface_mesh_v2.py:369` (now removed) noting this was missing.

Jira: https://flow360.atlassian.net/browse/FXC-6350

## Use case

> One use case is to report the number of nodes in the report of Flow360's service test. — FXC-6350

```python
sm = SurfaceMeshV2.from_cloud(surface_mesh_id)
print(sm.stats.n_nodes)
```

## Test plan

- [x] `black --check` clean
- [x] `isort --check` clean
- [x] `pylint -e C0301` (line length) clean, 10.00/10
- [x] Smoke test: `SurfaceMeshStats` parses a real-shaped JSON dict (nNodes, nTriangles, nQuadrilaterals, version, plus `boundaries` extras silently ignored by pydantic v2 default)
- [ ] End-to-end: fetch a real `SurfaceMeshV2` from cloud and call `.stats` (tester to confirm)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Low risk: adds a read-only `stats` accessor for surface meshes and slightly relaxes `VolumeMeshStats` field declarations without changing runtime behavior for valid responses.
> 
> **Overview**
> Adds a cached `SurfaceMeshV2.stats` property that fetches and parses `metadata/surfaceMeshMetaData.json` into a new `SurfaceMeshStats` model (`n_nodes`, `n_triangles`, `n_quadrilaterals`, optional `version`), raising `Flow360RuntimeError` when the stats artifact can’t be retrieved (e.g., mesh not ready).
> 
> Aligns `VolumeMeshStats` field definitions by removing explicit required markers (`...`) while keeping the same JSON aliases for all count fields.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 79a2750d13295f003a160081690404d7a66cf22e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/flow360/component/surface_mesh_v2.py
+++ b/flow360/component/surface_mesh_v2.py
@@ -33,7 +33,11 @@ from flow360.component.utils import (
     SurfaceMeshFile,
     shared_account_confirm_proceed,
 )
-from flow360.exceptions import Flow360FileError, Flow360ValueError
+from flow360.exceptions import (
+    Flow360FileError,
+    Flow360RuntimeError,
+    Flow360ValueError,
+)
 from flow360.log import log
 
 from .simulation.primitives import Surface
@@ -388,8 +392,14 @@ class SurfaceMeshV2(AssetBase):
         SurfaceMeshStats
             return SurfaceMeshStats object
         """
-        # pylint: disable=protected-access
-        data = self._webapi._parse_json_from_cloud(self._mesh_stats_file)
+        try:
+            # pylint: disable=protected-access
+            data = self._webapi._parse_json_from_cloud(self._mesh_stats_file)
+        except Exception as exc:
+            raise Flow360RuntimeError(
+                "Could not retrieve surface mesh stats. The surface mesh may not be ready yet "
+                f"(status={self._webapi.status.value})."
+            ) from exc
         return SurfaceMeshStats(**data)
 
     @classmethod
--- a/flow360/component/surface_mesh_v2.py
+++ b/flow360/component/surface_mesh_v2.py
@@ -392,9 +392,13 @@ class SurfaceMeshV2(AssetBase):
             # pylint: disable=protected-access
             data = self._webapi._parse_json_from_cloud(self._mesh_stats_file)
         except Exception as exc:
+            try:
+                status_str = self._webapi.status.value
+            except Exception:  # pylint: disable=broad-exception-caught
+                status_str = "unknown"
             raise Flow360RuntimeError(
                 "Could not retrieve surface mesh stats. The surface mesh may not be ready yet "
-                f"(status={self._webapi.status.value})."
+                f"(status={status_str})."
             ) from exc
         return SurfaceMeshStats(**data)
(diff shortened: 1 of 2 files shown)


flow360/component/surface_mesh_v2.py
```diff
@@ -83,6 +84,17 @@ class SurfaceMeshMetaV2(AssetMetaBaseModelV2):
     dependency: bool = pd.Field(False)
 
 
+class SurfaceMeshStats(pd.BaseModel):
+    """
+    Surface mesh stats
+    """
+
+    n_nodes: int = pd.Field(..., alias="nNodes")
```
benflexcompute: Why do we write ... here?

flow360/component/surface_mesh_v2.py
```diff
@@ -363,10 +375,22 @@ class SurfaceMeshV2(AssetBase):
...
+        """
+        Get surface mesh stats
+
+        Returns
+        -------
+        SurfaceMeshStats
+            return SurfaceMeshStats object
+        """
+        # pylint: disable=protected-access
+        data = self._webapi._parse_json_from_cloud(self._mesh_stats_file)
```
benflexcompute: surface mesh ID gets generated even when the surface mesh file has not been generated yet. In this case there will be no "metadata/surfaceMeshMetaData.json" on S3. What is the behavior of this function in this case?

benflexcompute: We might need guard on this to only try downloading when the surface mesh status is complete or something

flow360/component/surface_mesh_v2.py
mikeparkflex: The `...` first argument follows the convention in the volume stats. The majority of the code uses a different and equivalent Pydantic convention. I have modified both surface and volume stats to match the majority of the code

flow360/component/surface_mesh_v2.py
mikeparkflex: implemented try-catch blocks to guard. small scope, only surface. No change to is_final() ENUM.

benflexcompute (approved):

